PR #4, opened 2020-06-16
Update Jenkinsfile to build docker container image

qixiang:
After this change, the Jenkins pipeline job defined by Jenkinsfile
will only be responsible for building docker container images and
tagging images with git commit short SHA.

If job is ran for master branch, images will be tagged as latest.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,158 +6,76 @@ import groovy.json.*
 // 'global' var to store git info
 def scmVars
 // 'global' var to store application version
-def appversion
+def commitid
 
 try { // massive try{} catch{} around the entire build for failure notifications
-
-node('master'){
-    scmVars = checkout scm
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -6,158 +6,76 @@ import groovy.json.*
...
-        sh 'sudo dnf update -y python3-setuptools'
-    }
-    stage('Determine Version') {
-        // TODO: Set appversion as a label on the image?
-        appversion = sh(returnStdout: true, script: """
-            printf `python setup.py -V`-`git log -1 --pretty=format:%ct`
-        """).trim()
-    }
-    stage('Run unit tests') {
-        sh 'tox'
```
gnaponie, line 67: Why removing this part and the documentation part? I thought it was useful.

qixiang: Because Travis CI now runs these tests and build+publish the docs: https://github.com/redhat-exd-rebuilds/freshmaker/commit/5dd27852c5cd4738fd0a29ed1d6509bfec816d9b

PR #8, opened 2020-06-29
 Modify script to load a cassette

mulaievaRH:
Modify the "find_images_to_rebuild.py" script to be able to use a
cassette downloaded from the Freshmaker production instance rather
than querying Lightblue directly.

### DIFF
--- a/dev_scripts/README.md
+++ b/dev_scripts/README.md
@@ -24,3 +24,21 @@ the errata_id and it submits a build in dry_run mode.
 
 By default, the script is executed on prod. To execute the script on dev the parameter `--dev` 
 should be added.
+
+
+### find_images_to_rebuild
+
+This script can be used to find images to rebuild by an Errata ID Event. The argument 
+`--container-image` can be used to provide the NVRs of images to rebuild. If not specified, all 
+images will be rebuilt that are affected by the advisory. The `--cassette-path` argument can be set 
+to the path of a vcrpy cassette downloaded from a deployed Freshmaker instance. This will use the 
+recorded Lightblue queries instead of querying Lightblue directly.
+ 
+ Example usage:
+ 
+ ```bash
+python3 find_images_to_rebuild.py 53137
+  ```
+ ```bash
+python3 find_images_to_rebuild.py 53137 --cassette-path ~/vcr_test/fixtures/cassettes/1.yml
+ ```
--- a/dev_scripts/find_images_to_rebuild.py
+++ b/dev_scripts/find_images_to_rebuild.py
@@ -2,6 +2,7 @@
 from __future__ import print_function
 import os
 import sys
+import argparse
 from pprint import pprint
 from unittest.mock import patch
 from logging.config import dictConfig
@@ -15,12 +16,19 @@ os.environ["FRESHMAKER_DEVELOPER_ENV"] = "1"
 os.environ["FRESHMAKER_CONFIG_FILE"] = os.path.join(sys.path[0], "config.py")
 os.environ["REQUESTS_CA_BUNDLE"] = "/etc/ssl/certs/ca-bundle.crt"
 
-from freshmaker import db, app
+from freshmaker import db, app, conf
 from freshmaker.errata import Errata, ErrataAdvisory
 from freshmaker.events import (
     ErrataAdvisoryStateChangedEvent, ManualRebuildWithAdvisoryEvent)
 from freshmaker.handlers.koji import RebuildImagesOnRPMAdvisoryChange
 
+parser = argparse.ArgumentParser()
+parser.add_argument('errata_event_id', type=int, help='Errata event ID')
+parser.add_argument('--cassette-path', default=False, dest='cassette',
+                    help='Set a path to a cassette')
+parser.add_argument('--container-image', nargs='+', default=False, dest='container_img',
+                    help='Container images id')
+args = parser.parse_args()
 fedmsg_config = fedmsg.config.load_config()
 dictConfig(fedmsg_config.get('logging', {'version': 1}))
 
@@ -28,8 +36,6 @@ if len(sys.argv) < 2:
     print("Queries Lightblue to find out all the images Freshmaker rebuilds.")
     print("Usage: ./lightblue.py ERRATA_ID [[CONTAINER_IMAGE], ...]")
     sys.exit(1)
-
-container_images = sys.argv[2:]
 
 app_context = app.app_context()
 app_context.__enter__()
@@ -40,15 +46,33 @@ db.session.commit()
 
 errata = Errata()
 kwargs = {}
-if container_images:
+
+if args.container_img:
     EventClass = ManualRebuildWithAdvisoryEvent
-    kwargs['container_images'] = container_images
+    kwargs['container_images'] = args.container_img
 else:
     EventClass = ErrataAdvisoryStateChangedEvent
+
+event_id = None
+if args.cassette:
+    cassette_name = os.path.splitext(os.path.basename(args.cassette))[0]
+    extension = os.path.splitext(os.path.basename(args.cassette))[1]
+    try:
+        event_id = int(cassette_name)
+        if extension != ".yml":
+            raise ValueError
+    except ValueError:
+        print("The input cassette must be named in the format of <event ID>.yml")
+        sys.exit(1)
+
+    conf.vcrpy_path = os.path.dirname(args.cassette)
+    conf.vcrpy_mode = 'none'
+
 event = EventClass(
     "fake_message", ErrataAdvisory.from_advisory_id(errata, sys.argv[1]),
-    dry_run=True, **kwargs)
+    dry_run=True, freshmaker_event_id=event_id, **kwargs)
 
 handler = RebuildImagesOnRPMAdvisoryChange()
+
 with patch("freshmaker.consumer.get_global_consumer"):
     handler.handle(event)


mulaievaRH:
@mprahl , @gnaponie , PTAL :)

gnaponie:
I've tried with advisory "57009", could you try the same? 
It gives me an error: http://pastebin.test.redhat.com/883767

dev_scripts/find_images_to_rebuild.py
```diff
@@ -50,5 +62,21 @@
     dry_run=True, **kwargs)
 
 handler = RebuildImagesOnRPMAdvisoryChange()
+
+if args.cassete_path:
+    VCRPY_MODE = "none"
+    try:
+        file_dict = yaml.safe_load(args.cassete_path)
```
mprahl, line 69: You shouldn't need to load the cassette yourself. As long as you set the Freshmaker `vcrpy_path` configuration to the directory of the cassette and the cassette is named after the Freshmaker event ID, then this will just work.

dev_scripts/find_images_to_rebuild.py
```diff
@@ -50,5 +62,21 @@
     dry_run=True, **kwargs)
 
 handler = RebuildImagesOnRPMAdvisoryChange()
+
+if args.cassete_path:
+    VCRPY_MODE = "none"
```
mprahl, line 67: You need to set the Freshmaker configuration `vcrpy_mode` to `"none"`. I don't see you doing this in the code.

dev_scripts/find_images_to_rebuild.py
```diff
@@ -50,5 +62,21 @@
...
+    try:
+        file_dict = yaml.safe_load(args.cassete_path)
+        response_container_repo = json.loads(
+            file_dict['interactions'][0]['response']['body']['string'])
+        response_container_imgs = json.loads(
+            file_dict['interactions'][1]['response']['body']['string'])
+        with patch("freshmaker.consumer.get_global_consumer"), patch(
```
mprahl, line 74: vcrpy should handle all this for you. You shouldn't need to do any mocking.

mulaievaRH:
> I've tried with advisory "57009", could you try the same?
> It gives me an error: http://pastebin.test.redhat.com/883767

@gnaponie , ptal now after my last changes. It works now for me.

dev_scripts/find_images_to_rebuild.py
mulaievaRH: Oh. didn't know it. Now it looks much better. PTAL now. Did I understood right how it should be.

dev_scripts/find_images_to_rebuild.py
```diff
@@ -15,12 +16,18 @@
...
-from freshmaker import db, app
+from freshmaker import db, app, conf
 from freshmaker.errata import Errata, ErrataAdvisory
 from freshmaker.events import (
     ErrataAdvisoryStateChangedEvent, ManualRebuildWithAdvisoryEvent)
 from freshmaker.handlers.koji import RebuildImagesOnRPMAdvisoryChange
 
+parser = argparse.ArgumentParser()
+parser.add_argument('Errata event ID', metavar='N', type=int, nargs='*',
+                    help='Errata event ID and container images id')
```
mprahl, line 27: I think this will work. You only need to accept a single argument for the event ID.
```suggestion
parser.add_argument('errata_event_id', type=int, help='Errata event ID and container images id')
```

dev_scripts/find_images_to_rebuild.py
```diff
@@ -29,7 +36,8 @@
     print("Usage: ./lightblue.py ERRATA_ID [[CONTAINER_IMAGE], ...]")
     sys.exit(1)
 
-container_images = sys.argv[2:]
+container_images = [container_image for container_image in sys.argv[2:] if
```
mprahl, line 39: Add an argument in argsparse instead to avoid guessing what the container images. This approach is very fragile.

dev_scripts/find_images_to_rebuild.py
```diff
@@ -50,5 +58,15 @@
     dry_run=True, **kwargs)
```
mprahl, line 58: Just above this, you should be able to get the argument from `args` rather than `sys.argv`.

dev_scripts/find_images_to_rebuild.py
```diff
@@ -50,5 +58,15 @@
...
 
 handler = RebuildImagesOnRPMAdvisoryChange()
-with patch("freshmaker.consumer.get_global_consumer"):
-    handler.handle(event)
+
+if args.cassete_path:
+    import vcr
+    my_vcr = vcr.VCR(
```
mprahl, line 64: This code already initializes vcrpy:
https://github.com/redhat-exd-rebuilds/freshmaker/commit/687072cfba2d2474bbf72e4a4359c8c3df794ddf#diff-ca295b25c8157b5a17eb8e3054e8c49dR526-R533

I have not confirmed this, but I think you could create a temporary directory and set `conf.vcrpy_path` to that temporary directory. Then create a symlink to the cassette file in the temporary directory so that it is named after the event created above (e.g. `f'{event.id}.yml'` (verify event.id is actually set)). Then if you set `conf.record_mode` to `"none"`, it should automatically pick up the cassette and use it.

dev_scripts/find_images_to_rebuild.py
```diff
@@ -15,12 +16,18 @@
...
+from freshmaker import db, app, conf
 from freshmaker.errata import Errata, ErrataAdvisory
 from freshmaker.events import (
     ErrataAdvisoryStateChangedEvent, ManualRebuildWithAdvisoryEvent)
 from freshmaker.handlers.koji import RebuildImagesOnRPMAdvisoryChange
 
+parser = argparse.ArgumentParser()
+parser.add_argument('Errata event ID', metavar='N', type=int, nargs='*',
+                    help='Errata event ID and container images id')
+parser.add_argument('--cassete_path', type=argparse.FileType(), default=False,
```
mprahl, line 28: It's customary to use hyphens in CLI arguments.
```suggestion
parser.add_argument('--cassete-path', type=argparse.FileType(), default=False,
```

mulaievaRH:
@mprahl , review please :)

dev_scripts/find_images_to_rebuild.py
```diff
@@ -15,12 +16,19 @@
...
 from freshmaker.errata import Errata, ErrataAdvisory
 from freshmaker.events import (
     ErrataAdvisoryStateChangedEvent, ManualRebuildWithAdvisoryEvent)
 from freshmaker.handlers.koji import RebuildImagesOnRPMAdvisoryChange
 
+parser = argparse.ArgumentParser()
+parser.add_argument('errata_event_id', type=int, help='Errata event ID')
+parser.add_argument('--cassette-path', default=False, dest='cassette',
+                    help='Set a path to a cassette')
+parser.add_argument('--container-images', nargs='+', default=False, dest='container_imgs',
```
mprahl, line 29: I don't think this should be plural since you have to specify it multiple times to pass many container images
```suggestion
parser.add_argument('--container-image', nargs='+', default=False, dest='container_imgs',
```

dev_scripts/README.md
```diff
@@ -24,3 +24,18 @@ the errata_id and it submits a build in dry_run mode.
 
 By default, the script is executed on prod. To execute the script on dev the parameter `--dev` 
 should be added.
+
+
+###find_images_to_rebuild
+
+This script can be used to find images to rebuild by Errata Id Event. With the argument 
+`--container-images` can be set  list of NVRs of images to rebuild otherwise will rebuild all images
```
mprahl, line 32:
```suggestion
`--container-image` can be used to provide the NVRs of images to rebuild. If not specified, all images will be rebuilt that are
```

dev_scripts/README.md
```diff
@@ -24,3 +24,18 @@ the errata_id and it submits a build in dry_run mode.
...
+affected by the advisory. With the argument `--cassette_path` can be set a path to the recorded
```
mprahl, line 33:
```suggestion
affected by the advisory. With the argument `--cassette-path` can be set a path to the recorded 
```

dev_scripts/README.md
```diff
@@ -24,3 +24,18 @@ the errata_id and it submits a build in dry_run mode.
...
 By default, the script is executed on prod. To execute the script on dev the parameter `--dev` 
 should be added.
+
+
+###find_images_to_rebuild
+
+This script can be used to find images to rebuild by Errata Id Event. With the argument 
+`--container-images` can be set  list of NVRs of images to rebuild otherwise will rebuild all images
+affected by the advisory. With the argument `--cassette_path` can be set a path to the recorded 
+cassette downloaded from the Freshmaker production instance rather than querying Lightblue directly.
```
mprahl, line 34:
```suggestion
affected by the advisory. The `--cassette-path` argument can be set to the path to a recorded 
cassette downloaded from the Freshmaker production. This will use the recorded Lightblue queries
instead of querying Lightblue directly.
```

dev_scripts/README.md
```diff
@@ -24,3 +24,18 @@ the errata_id and it submits a build in dry_run mode.
...
+ 
+ Example usage:
+       
+       python3 find_images_to_rebuild.py 53137
+       OR
+       python3 find_images_to_rebuild.py 53137 --cassette_path ~/vcr_test/fixtures/cassettes/1.yml
```
mprahl, line 40:
```suggestion
```bash
python3 find_images_to_rebuild.py 53137
```
```bash
python3 find_images_to_rebuild.py 53137 --cassette_path ~/vcr_test/fixtures/cassettes/1.yml
```

dev_scripts/README.md
```diff
@@ -24,3 +24,18 @@ the errata_id and it submits a build in dry_run mode.
...
+###find_images_to_rebuild
+
+This script can be used to find images to rebuild by Errata Id Event. With the argument 
+`--container-images` can be set  list of NVRs of images to rebuild otherwise will rebuild all images
+affected by the advisory. With the argument `--cassette_path` can be set a path to the recorded 
+cassette downloaded from the Freshmaker production instance rather than querying Lightblue directly.
+ 
+ Example usage:
+       
+       python3 find_images_to_rebuild.py 53137
```
mprahl, line 38: How about using ```bash for each of these?

mprahl: Please remember to change `cassette_path` to `cassette-path`.

mulaievaRH:
@mprahl , all is fixed

dev_scripts/find_images_to_rebuild.py
```diff
@@ -40,15 +46,30 @@
...
 else:
     EventClass = ErrataAdvisoryStateChangedEvent
+
+event_id = None
+if args.cassette:
+    cassette_name = os.path.splitext(os.path.basename(args.cassette))[0]
+    try:
+        event_id = int(cassette_name)
+    except ValueError:
+        print("The input cassette must be named in the format of <event ID>.yaml")
```
mprahl, line 62: This was my mistake. I thought 687072cfba2d2474bbf72e4a4359c8c3df794ddf used `.yaml`, but it's actually `.yml`.
```suggestion
        print("The input cassette must be named in the format of <event ID>.yml")
```

dev_scripts/find_images_to_rebuild.py
```diff
@@ -40,15 +46,30 @@
...
+if args.container_img:
     EventClass = ManualRebuildWithAdvisoryEvent
-    kwargs['container_images'] = container_images
+    kwargs['container_images'] = args.container_img
 else:
     EventClass = ErrataAdvisoryStateChangedEvent
+
+event_id = None
+if args.cassette:
+    cassette_name = os.path.splitext(os.path.basename(args.cassette))[0]
```
mprahl, line 58: Optional: You could verify if the extension is `.yml` to avoid someone having a cassette that has the extension `.yaml` and have it query Lightblue directly.

mulaievaRH:
@mprahl , one more time review please :)

dev_scripts/find_images_to_rebuild.py
```diff
@@ -40,15 +46,33 @@
...
+    extension = os.path.splitext(os.path.basename(args.cassette))[1]
+    try:
+        event_id = int(cassette_name)
+        if extension is not ".yml":
```
mprahl, line 62: `is` will compare the memory address and not if the values are equivalent. For this, use `==`/`!=`.
```suggestion
        if extension != ".yml":
```

This shows the behavior:
```
>>> extension = ".yml"
>>> extension is ".yml"
False
>>> extension is extension
True
```

mprahl:
@mulaievaRH just one small comment to address. Were you able to test this to see if it works?

dev_scripts/find_images_to_rebuild.py
```diff
@@ -40,15 +46,33 @@
...
+    kwargs['container_images'] = args.container_img
 else:
     EventClass = ErrataAdvisoryStateChangedEvent
+
+event_id = None
+if args.cassette:
+    cassette_name = os.path.splitext(os.path.basename(args.cassette))[0]
+    extension = os.path.splitext(os.path.basename(args.cassette))[1]
+    try:
+        event_id = int(cassette_name)
```
mulaievaRH, line 61: > @mulaievaRH just one small comment to address. Were you able to test this to see if it works?

@mprahl , Yes, I've tested it. For the test I've deleted here int() for trying `--cassette-path ~/vcr_test/fixtures/cassettes/fake_message.yml` (because I have not a 'real' cassette ). It works.

PR #9, opened 2020-06-29
Add database section to documentaion

apaplaus:
RESOLVE: CLOUDWF-113

### DIFF
--- a/docs/db.rst
+++ b/docs/db.rst
@@ -1,0 +1,110 @@
+========================
+Freshmaker database
+========================
+
+Freshmaker as an SQL toolkit uses SQLAlchemy extension for `Flask <https://flask-sqlalchemy.palletsprojects.com/en/2.x/>`_. So whenever you want to add some columns or tables you should at least check ``models.py`` file. Every class that has a **FreshmakerBase** as a parent is actually a table with its own columns, relationships etc.
+
+As for database management system, we are using `PostgreSQL <https://www.postgresql.org/>`_ in deployments and `SQLite <https://www.sqlite.org/index.html>`_ in local environment.
+And besides that for testing we are using *in-memory* database, so whenever you are running tests from ``tests/`` folder, new database is created and live while tests are running.
+
+
+Setting up Database
+============================
+When you for the first time want to create freshmaker database, you are supposed to run:
+
+.. sourcecode:: none
+
+    ./manager createdb
+
+This command will create new ``freshmaker.db`` file for local use for you.
+
+
+Database update
+========================
+
+When some changes are applied to database scheme, like changing ``models.py``, you should generate  migration file for a new database scheme by running:
+
+.. sourcecode:: none
+
+    ./manager db migrate
+
+After that you can apply that migration by running:
+
+.. sourcecode:: none
+
+    ./manager db upgrade
+
+After that all changes should be already applied to your local database.
+
+
+Querying database
+==========================
+
+You can query your locally created database by running something like this in the directory with database file:
+
+.. sourcecode:: none
+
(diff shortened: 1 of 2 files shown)


apaplaus:
@gnaponie can you please review if I didn't miss something and if everything I wrote is true)

mprahl:
We use the same stack for interacting with the database in Cachito and Nizi added some pretty good instructions in case you want to draw inspiration:
https://github.com/release-engineering/cachito/commit/6bcae1e41aa3ecddb6b9dfbcc72efbc9fa5a9b68

gnaponie:
@apaplaus all the information look correct :) I'll follow maybe with 1 or 2 comments on the syntax.
Maybe you could also add the info that you can check the context of the local database like this (example):
`sqlite3 freshmaker.db 'SELECT `id` FROM artifact_builds;'`

docs/db.rst
```diff
@@ -0,0 +1,37 @@
+========================
+Freshmaker database
+========================
+
+Freshmaker as an SQL toolkit uses SQLAlchemy extension for `Flask <https://flask-sqlalchemy.palletsprojects.com/en/2.x/>`_. So whenever you want to add some columns or event tables you should at least check ``models.py`` file. Every class that has a **FreshmakerBase** as a parent is actually a table with it's own columns, relationships etc.
```
gnaponie, line 5: "So whenever you want to add some columns or event tables" -- not sure why "event tables", did you mean `new tables`?

gnaponie: s/a table with it's own columns/a table with its own columns/g

docs/db.rst
```diff
@@ -0,0 +1,37 @@
...
+
+As for database management system, we are using `PostgreSQL <https://www.postgresql.org/>`_ in deployments and `SQLite <https://www.sqlite.org/index.html>`_ in local environment.
+And besides that for testing we are using *in-memory* database, so whenever you are running tests from ``tests/`` folder, new database are created and live while tests are running.
```
gnaponie, line 8: s/new database are created/a new database is created/g

apaplaus:
@gnaponie I've made changes proposed by you, and btw I am not sure if I should add info from [here](https://github.com/release-engineering/cachito/commit/6bcae1e41aa3ecddb6b9dfbcc72efbc9fa5a9b68) to our documentation. What do you think?

docs/db.rst
mprahl: @gnaponie I suggest you use the "Insert Suggestion" feature of GitHub instead when suggesting changes like this  :smile:

gnaponie: @mprahl I guess I'm getting old :D

gnaponie:
@apaplaus the changes look good +1 thanks for addressing the comments.
About the question: yeah, if there are things that apply also to Freshmaker and you didn't mention here, maybe it'd make sense to add them.

docs/db.rst
```diff
@@ -0,0 +1,110 @@
...
+When you for the first time want to create freshmaker database, you are supposed to run:
+
+.. sourcecode:: none
+
+    ./manager createdb
+
+This command will create new ``freshmaker.db`` file for local use for you.
+
+
+Database update
```
gnaponie, line 22: You're repeating these information below in the `Database Migrations` section, it might make sense to unify the sections and keep just one of the two.

apaplaus: Yes it's repetitive, but *Database migration* section is more detailed and complex. From my view if I wanted to just make some changes in scheme(as specified in the task) I would like to know just a few steps/commands that are needed. And if I wanted to dig deeper into 'how migrations work' I would use that guide for migration.

gnaponie: Alright then, in that case I'd say it looks good.

PR #16, opened 2020-07-27
[WIP]Remove some old references of pagure and fedoraproject

apaplaus:
Some fedora references are left because it could be useful in future for OIDC authentication.

RESOLVE: CLOUDWF-2666

### DIFF
--- a/freshmaker/config.py
+++ b/freshmaker/config.py
@@ -184,11 +184,11 @@ class Config(object):
             'desc': 'Polling interval, in seconds.'},
         'git_base_url': {
             'type': str,
-            'default': "git://pkgs.fedoraproject.org",
+            'default': "git://pkgs.devel.redhat.com",
             'desc': 'Dist-git base URL.'},
         'git_ssh_base_url': {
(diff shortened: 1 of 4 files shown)


freshmaker/config.py
```diff
@@ -184,19 +184,19 @@ class Config(object):
...
+            'default': "ssh://%s@pkgs.devel.redhat.com/",
             'desc': 'Dist-git ssh base URL.'},
         'git_user': {
             'type': str,
             'default': '',
             'desc': 'User for git operations.'},
         'git_author': {
             'type': str,
-            'default': 'Freshmaker <freshmaker-owner@fedoraproject.org>',
+            'default': 'Freshmaker <freshmaker-owner@github.com>',
```
gnaponie, line 199: *Maybe* we should create this user in github too. Or maybe let's just leave it since this conf will never actually be used.

gnaponie:
I guess it makes sense to make this change since we'll probably never have a freshmaker instance in fedora.

PR #17, opened 2020-07-29
Adding the findutils install in the Dockerfile

mulaievaRH:
Install find in the Freshmaker container for the possibility to create
an OpenShift CronJob to delete cassettes on the persistent volume that
are older than two weeks.

JIRA: https://projects.engineering.redhat.com/browse/CLOUDWF-1937

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ RUN sed -i 's_https://_http://_' /etc/yum.repos.d/rcm-tools-fedora.repo
 COPY yum-packages.txt /tmp/yum-packages.txt
 
 RUN \
-    dnf -y install $(cat /tmp/yum-packages.txt) python3-rhmsg && \
+    dnf -y install $(cat /tmp/yum-packages.txt) python3-rhmsg findutils && \
     dnf clean all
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -22,6 +22,7 @@ COPY yum-packages.txt /tmp/yum-packages.txt
 
 RUN \
     dnf -y install $(cat /tmp/yum-packages.txt) python3-rhmsg && \
+    dnf install -y findutils && \
```
mprahl, line 25:
```suggestion
    dnf -y install $(cat /tmp/yum-packages.txt) python3-rhmsg findutils && \
```

mprahl:
@mulaievaRH could you please fix the typo in the commit message and PR title? Also, could you please explain why you are adding this dependency in the commit message? Linking to the JIRA ticket would also be nice as well for added context.

mulaievaRH:
> @mulaievaRH could you please fix the typo in the commit message and PR title? Also, could you please explain why you are adding this dependency in the commit message? Linking to the JIRA ticket would also be nice as well for added context.

All is done.

PR #18, opened 2020-07-31
Fix a typo: parent image is a tuple instead of None

gnaponie:
ref: CLOUDWF-2810


qixiang:
+1, the issue of parent not get resolved can be addressed in other PR.

PR #19, opened 2020-08-10
Use Errata Tool CVE mapping functionality

qixiang:
Errata now provides an API to get builds by cve for an advisory, it
returns the builds affected by CVEs in advisory, we can use this API to
replace the code of fetching packages affected by CVEs and then compare
the result with builds in current advisory.

JIRA: CLOUDWF-2482


qixiang:
More explanation on the change:

1. Before this change, when an advisory is shipped, we firstly get the CVEs in this advisory, then query sfm2 to get the packages affected by these CVEs, and then compare those packages with the builds in this advisory, in this way we'll know what packages are affected by the CVEs in this advisory.

2.  After this change, we use Errata's new API `get_builds_by_cve` to get the list of builds in this advisory that are affected by the CVEs. One thing need to mention is, for module advisories, the builds in advisory are module builds, they are not srpms, so we need fetch the "SRPMS" in the returned data instead of just "builds".

3. We keeps querying sfm2 to get the highest severity impact for CVEs, because advisory's security impact value may differ from the highest severity impact we get from sfm2, though I only found it happens to 41 of 1087 advisories.

qixiang:
PR is updated to eliminate the sfm2 module, replace the `highest_cve_severity` by advisory's `security_impact`, the severity value from sfm2 for each CVE is a global value, it can be different for a specific product, so `security_impact` from advisory is more precise.

apaplaus:
It looks good for me, but there is some change in functionality I would say. So I suggest to ask someone else to have a look at it too. Probably @mprahl may know something about **severity** and sfm2.

qixiang:
For the change of using "security_impact", I added a comment to jira issue, @mprahl if you're not clear about the change, you can check the jira issue for more details.

freshmaker/errata.py
```diff
@@ -78,6 +84,9 @@ def from_advisory_id(cls, errata, errata_id):
         else:
             cve_list = []
 
+        # security_impact in errata is capitalized string
+        security_impact = erratum_data["security_impact"].lower()
```
mprahl, line 88: Is making this lowercase for backwards compatibility with SFM2?

freshmaker/errata.py
```diff
@@ -379,3 +388,24 @@ def get_pulp_repository_ids(self, errata_id):
...
+        for cve, data_by_product in data.items():
+            for product, product_data in data_by_product.items():
+                for build in product_data.get("builds", []):
+                    for build_nvr, build_info in build.items():
+                        # module builds are not source rpms, we need SRPMS in variants
+                        for variant, variant_data in build_info.get("variant_arch", {}).items():
+                            srpms.extend(variant_data.get("SRPMS", []))
+
+        srpm_nvrs = [x.split(".src.rpm")[0] for x in set(srpms)]
+        return srpm_nvrs
```
mprahl, line 411: Optional: I'd prefer using a set for this for efficiency and less memory usage, but I'll leave it up to you.
You can also just return a set if your calling function doesn't need a list. This tells the function caller that the order doesn't matter and that there are no duplicates, which is nice since this is always be true of this return value.
```suggestion
        srpms = set()
        for cve, data_by_product in data.items():
            for product, product_data in data_by_product.items():
                for build in product_data.get("builds", []):
                    for build_nvr, build_info in build.items():
                        # module builds are not source rpms, we need SRPMS in variants
                        for variant, variant_data in build_info.get("variant_arch", {}).items():
                            for srpm in variant_data.get("SRPMS", []):
                                srpms.add(srpm.split(".src.rpm")[0])

        return list(srpms)
```

freshmaker/errata.py
```diff
@@ -379,3 +388,24 @@ def get_pulp_repository_ids(self, errata_id):
...
+        :return: a list of strings each of them is a source rpm nvr
+        :rtype: list
+        """
+        data = self._errata_rest_get(f"/erratum/{errata_id}/builds_by_cve")
+        srpms = []
+        for cve, data_by_product in data.items():
+            for product, product_data in data_by_product.items():
+                for build in product_data.get("builds", []):
+                    for build_nvr, build_info in build.items():
+                        # module builds are not source rpms, we need SRPMS in variants
```
mprahl, line 406: I don't understand what this comment is trying to say. Could you please rephrase it?

freshmaker/errata.py
qixiang: Yes, comment is updated.

qixiang:
Comments are addressed, updated the in line comments and changed to use set instead of list while getting the srpm nvrs, it's a minor change and I've tested it locally, since I've got the approvals, I'm going to merge this, thanks for reviewing.

PR #21, opened 2020-08-17
Set parent build to the right value

apaplaus:
Previously we were using wrong key for retrieving parent image nvr for
async rebuilds.

RESOLVES: CLOUDWF-2810

### DIFF
--- a/tests/handlers/koji/test_rebuild_images_on_async_manual_build.py
+++ b/tests/handlers/koji/test_rebuild_images_on_async_manual_build.py
@@ -19,6 +19,8 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+from unittest.mock import MagicMock
+
 from freshmaker import db
 from freshmaker.handlers.koji import RebuildImagesOnAsyncManualBuild
 from freshmaker.events import FreshmakerAsyncManualBuildEvent
@@ -80,7 +82,7 @@ class TestRebuildImagesOnAsyncManualBuild(helpers.ModelsTestCase):
         #              +
         #            image_e
 
-        # image_c is unrelated.
+        # image_c and image_f are unrelated.
 
         # image_0 is a base image, with no parent
         self.image_0 = ContainerImage({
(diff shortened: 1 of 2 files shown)


qixiang:
Can we add a test for this?
And I'm thing about adding a feature to resolve parent to latest published version, building an image against the original parent image can result in some regressions if the parent has been updated to include some CVE fixes. It would be great to have this feature enabled by default and can disable it in the request with "resolve_to_latest_parent: false". What do you think?

apaplaus:
@qixiang I've added some tests, and made some improvements to my solution, can you please check it. Tests were not so easy to write)
And about your feature, isn't it the work for some other service, like metaxor or someone else, to set up the latest parent for images? I am not sure if it's safe to user some other images as a parent, then the one setted in image. Because maybe there could be some incompatibilities or something like that between newer parent and child, that we are processing. I don't know, maybe I am wrong, and it's impossible, just thoughts.

tests/handlers/koji/test_rebuild_images_on_async_manual_build.py
```diff
@@ -387,3 +414,42 @@ def test_related_images_are_built(self):
...
+
+    def test_parent_if_image_without_parent(self):
+        """
+        This tests if we get parent as brew build of single image to rebuild
+        when image doesn't have "parent" key
+        """
+        self.mock_find_images_to_rebuild.return_value = [self.image_f]
+        event = FreshmakerAsyncManualBuildEvent(
+            'msg-id-123', 'test_branch', ['image-a-container'])
+        from unittest.mock import MagicMock
```
qixiang, line 426: Can you import it at top of file?

tests/handlers/koji/test_rebuild_images_on_async_manual_build.py
```diff
@@ -387,3 +414,42 @@ def test_related_images_are_built(self):
...
+        from json import dumps as dump
+        for image in [self.image_0, self.image_a]:
+            self.mock_find_images_to_rebuild.return_value = [image]
+            event = FreshmakerAsyncManualBuildEvent(
+                'msg-id-123', 'test_branch', ['image-a-container'])
+
+            with patch('freshmaker.handlers.koji.rebuild_images_on_async_manual_build.json.dumps') as json_mock:
+                json_mock.side_effect = dump
+                RebuildImagesOnAsyncManualBuild().handle(event)
+                json_mock.assert_called_once()
```
qixiang, line 449: Can you check the builds in db instead of checking the `json.dumps` call?

apaplaus:
@qixiang changes are done, if everything else is fine, can I merge it?

tests/handlers/koji/test_rebuild_images_on_async_manual_build.py
```diff
@@ -387,3 +414,51 @@ def test_related_images_are_built(self):
...
+
+    def test_parent_if_image_with_parent(self):
+        """
+        This tests if we get parent of single image to rebuild, when image
+        has "parent" key as None OR as some image
+        """
+        event_id = 'msg-id-'
+        for image in [self.image_0, self.image_a]:
+            self.mock_find_images_to_rebuild.return_value = [image]
+            event_id += '1'
```
qixiang, line 448: optional: you can it with
```
        for index, image in enumerate([self.image_0, self.image_a]):
            ...
            event_id = f'msg-id-{index}'
```

qixiang:
Added a comment, it's optional, LGTM.

PR #22, opened 2020-08-19
Filter images by builds in advisory for non-CVE advisories

qixiang:
Fixed a regression in 3726cd4, after changing to use CVE mapping
feature from errata, the case of non-CVE advisory triggered rebuilds
will not work, this is a case in E2E testing. Put the code back to
fix it.


apaplaus (approved):
lgfm

PR #23, opened 2020-08-31
Use binary rpms for filtering images

apaplaus:
Now we will use binary rpms instead of SRPMs to filter images. It's
possible due to new functionality in ET, allowing us to get binary rpms
 affected by CVEs in an advisory straight from ET.

RESOLVE CLOUDWF-2968

### DIFF
--- a/freshmaker/errata.py
+++ b/freshmaker/errata.py
@@ -55,16 +55,16 @@ class ErrataAdvisory(object):
         self.cve_list = cve_list or []
         self.has_hightouch_bug = has_hightouch_bug
 
-        self._affected_srpm_nvrs = None
+        self._affected_rpm_nvrs = None
 
     @property
-    def affected_srpm_nvrs(self):
-        if self._affected_srpm_nvrs is not None:
-            return self._affected_srpm_nvrs
+    def affected_rpm_nvrs(self):
+        if self._affected_rpm_nvrs is not None:
+            return self._affected_rpm_nvrs
 
         errata = Errata()
-        self._affected_srpm_nvrs = errata.get_cve_affected_srpm_nvrs(self.errata_id)
-        return self._affected_srpm_nvrs
+        self._affected_rpm_nvrs = errata.get_cve_affected_rpm_nvrs(self.errata_id)
+        return self._affected_rpm_nvrs
 
     @classmethod
     def from_advisory_id(cls, errata, errata_id):
@@ -318,10 +318,11 @@ class Errata(object):
 
         return data["rhel_release"]["name"]
 
-    def get_builds(self, errata_id, rhel_release_prefix=None):
+    def _get_rpms(self, errata_id, rhel_release_prefix=None):
         """
-        Returns set of NVRs of builds added to the advisory. These are just
-        brew build NVRs, not the particular RPM NVRs.
+        Returns dictionary of NVRs of builds added to the advisory.
+        "source_rpms" key with SRPMs as a value
+        "binary_rpms" key with binary rpms as a value
 
         If module build is attached to advisory, also all the NVRs of builds
         included in this module build are returned, together with the NVR of
@@ -334,18 +335,9 @@ class Errata(object):
             `rhel_release_prefix`. For example to return only RHEL-7 builds,
             this should be set to "RHEL-7".
             Defaults to conf.errata_rhel_release_prefix.
-        :rtype: set of strings
-        :return: Set of NVR builds.
+        :rtype: dict
+        :return: Dictionary with source and binary rpms.
         """
-        def get_srpms_nvrs(build_dict):
-            """ Gets srpms nvrs from the build dictionary. """
-            for key, val in build_dict.items():
-                if build_dict.get('SRPMS'):
-                    return {nvr.split(".src.rpm")[0] for nvr in build_dict['SRPMS']}
-                if isinstance(val, dict):
-                    return get_srpms_nvrs(val)
-                return
-
         if rhel_release_prefix is None:
             rhel_release_prefix = conf.errata_rhel_release_prefix
 
@@ -353,7 +345,8 @@ class Errata(object):
             "advisory/%s/builds.json" % str(errata_id))
 
         # Store NVRs of all builds in advisory to nvrs set.
-        nvrs = set()
+        source_rpms = set()
+        binary_rpms = set()
         for product_version, builds in builds_per_product.items():
             if rhel_release_prefix:
                 rhel_release = Errata.product_region.get(product_version)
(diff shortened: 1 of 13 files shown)


freshmaker/lightblue.py
```diff
@@ -712,18 +712,19 @@ def _get_default_projection(self, srpm_names=None, include_rpm_manifest=True):
...
                          "$or": [{
-                             "field": "srpm_name",
+                             "field": "name",
                              "op": "=",
-                             "rvalue": srpm_name
-                         } for srpm_name in srpm_names]},
+                             "rvalue": rpm_name
+                         } for rpm_name in rpm_names]},
                      "project": [
                          {"field": "srpm_nevra", "include": True},
```
apaplaus, line 725: Not sure if we still need 'srpm_nevra' and 'srpm_name' fields

freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
```diff
@@ -411,19 +411,19 @@ def _find_images_to_rebuild(self, errata_id):
...
+        # Get binary rpm nvrs which are affected by the CVEs in this advisory
+        affected_nvrs = self.event.advisory.affected_rpm_nvrs
 
-        # If there is no CVE affected srpms, this can be non-RHSA advisory,
+        # If there is no CVE affected binary rpms, this can be non-RHSA advisory,
         # just rebuild images that have the builds in this advisory installed
-        if not srpm_nvrs:
-            srpm_nvrs = list(errata.get_builds(errata_id))
+        if not affected_nvrs:
+            affected_nvrs = list(errata.get_builds(errata_id))
```
qixiang, line 420: `get_builds` returns the srpm nvrs, since we're changing to filtering by binary rpms, this is also required to be changed or add another api to get the binary rpms.

apaplaus: Yes, that was one of questions I wanted to ask you. That method is used in some other places, like **odcsclient.py::prepare_yum_repo** and in **handlers/koji/rebuild_images_on_parent_image_build.py** so I think I should leave the existing implementation with SRPMs for mentioned usecases. And just add new functionality(with binary rpms) to use only in **rebuild_images_on_parent_image_build.py**.
Or should I just change that method to use binary rpms instead or SRPMs? What do you think is better here?

apaplaus: And one more thing, there is a comment to the usage of **get_builds** method "# If there is no CVE affected binary rpms, this can be non-RHSA advisory, just rebuild images that have the builds in this advisory installed". So actually as I understand we will use that method only when there won't be any CVE affected nvrs. So that new ET api call will not give us any binary rpms. And we will just rebuild all builds that are attached to that advisory. So maybe we should leave that method as it is? because it's somehow other use-case then we are working on.

qixiang: You need to keep `get_builds` for odcs `prepare_yum_repo` and `rebuild_images_on_parent_image_build.py` because they're expecting build nvr instead of binary nvr.

So you need another api to get all binary rpm nvrs from advisory, and for the non-RHSA case, you also need binary rpm 
nvrs instead of build/srpm nvrs, because we'll use these nvrs later to query/filter images in lightblue for both the RHSA and non-RHSA cases.

apaplaus: So I've changed that part. You can check it now. And btw I left some comments in the code with some stuff that is not pretty clear for me, so could you comment on them too? Thx)

freshmaker/errata.py
```diff
@@ -392,25 +411,30 @@ def get_pulp_repository_ids(self, errata_id):
...
-                            for srpm in variant_data.get("SRPMS", []):
-                                srpms.add(srpm.split(".src.rpm")[0])
-
-        return list(srpms)
+                            for arch, rpms in variant_data.items():
+                                # Do we need to exclude 'noarch' too?
+                                # Remove '.arch.....' part from rpm's name
+                                # and make a list from them
+                                if arch != 'SRPMS':
+                                    just_nvrs = [rpm.partition("." + arch)[0] for rpm in rpms]
```
qixiang, line 437: You can use `rpm.rsplit('.', 2)[0]`, not sure if there could be some cases that arch doesn't match with the string in rpm.

freshmaker/errata.py
```diff
@@ -392,25 +411,30 @@ def get_pulp_repository_ids(self, errata_id):
...
         :type errata_id: str or int
-        :return: a list of strings each of them is a source rpm nvr
+        :return: a list of strings each of them is a binary rpm nvr
         :rtype: list
         """
         data = self._errata_rest_get(f"/erratum/{errata_id}/builds_by_cve")
-        srpms = set()
-        for cve, data_by_product in data.items():
-            for product, product_data in data_by_product.items():
+        final_rpms = set()
```
qixiang, line 423: I'd like just `rpms` as the var name.

freshmaker/errata.py
```diff
@@ -318,7 +318,7 @@ def _rhel_release_from_product_version(self, errata_id, product_version):
 
         return data["rhel_release"]["name"]
 
-    def get_builds(self, errata_id, rhel_release_prefix=None):
+    def get_builds(self, errata_id, rhel_release_prefix=None, return_srpm=True):
         """
```
qixiang, line 322: Can we have a separate api to get binary rpms to avoid confusion?

apaplaus: Well actually the whole functionality to filter 'srpms' or 'binary rpms' is located inside small function inside that **get_builds** method. So If we want to have separate method, it will have like 75% of code exactly the same as that method, so it will be a lot of duplication.

freshmaker/errata.py
apaplaus: I don't know, I think the arch and the part of string should be the same. And I am not sure that it's safe to just split by dot, because I think there could be different amount of dots at the end of string.

freshmaker/errata.py
```diff
@@ -318,7 +318,7 @@ def _rhel_release_from_product_version(self, errata_id, product_version):
 
         return data["rhel_release"]["name"]
 
-    def get_builds(self, errata_id, rhel_release_prefix=None):
+    def get_builds(self, errata_id, rhel_release_prefix=None, return_srpm=True):
```
qixiang, line 321: btw, this api has a bug that module nvr is returned in the nvr list when we're getting the srpm nvrs, could you fix it as well?

apaplaus: What exactly do you mean by **module nvr**?  For example for that advisory builds https://errata.devel.redhat.com/api/v1/erratum/53134/builds.json you will have **nodejs-10-8010020200318152252.c27ad7f8** in the nvrs list?

qixiang: Yes, that's what I mean, *nodejs-10-8010020200318152252.c27ad7f8* never appears in a systems (s)rpm list, so it should be excluded from the srpm nvrs. And I believe we can simplify the code by removing `get_srpm_nvrs` and `get_binary_nvrs`, only get source rpm nvrs and rpm nvrs from the actual rpms which is under the arches, so it's unnecessary have the code to get the build nvrs.

freshmaker/errata.py
mikebonnet: RPM filenames always end with .<arch>.rpm. There are cases where the arch in the rpm can vary (i386, i586, i686). The `rsplit()` method is more reliable.

freshmaker/lightblue.py
```diff
@@ -767,17 +770,18 @@ def filter_out_images_with_higher_srpm_nvr(self, images, srpm_name_to_nvrs):
                 ret.append(image)
             image_included = False
             for rpm in rpms or []:
-                image_srpm_nvr = kobo.rpmlib.parse_nvr(rpm["srpm_nevra"])
-                for srpm_nvr in srpm_name_to_nvrs.get(rpm.get("srpm_name"), []):
-                    input_srpm_nvr = kobo.rpmlib.parse_nvr(srpm_nvr)
+                # Not sure about which NVR we should parse srpm or common rpm nvr?
+                image_rpm_nvr = kobo.rpmlib.parse_nvr(rpm["nvra"])
```
qixiang, line 774: you need to remove the 'arch' part from nvra before `parse_nvr`.

freshmaker/handlers/koji/rebuild_images_on_parent_image_build.py
```diff
@@ -176,7 +176,7 @@ def _verify_advisory_rpms_in_container_build(self, errata_id, container_build_id
         # and use set of the nvrs as a value.
         advisory_rpms_by_name = {}
         e = Errata()
-        build_nvrs = e.get_builds(errata_id)
+        build_nvrs = e.get_srpm_nvrs(errata_id)
```
qixiang, line 179: Please also comparing with binary rpms instead of srpms.

freshmaker/lightblue.py
apaplaus: I will just replace it with `parse_nvra` method, if it's appropriate.

freshmaker/handlers/koji/rebuild_images_on_parent_image_build.py
apaplaus: Done

freshmaker/errata.py
apaplaus: Ok, I've changed it to `rsplit()`

freshmaker/handlers/koji/rebuild_images_on_parent_image_build.py
```diff
@@ -176,7 +176,7 @@ def _verify_advisory_rpms_in_container_build(self, errata_id, container_build_id
         # and use set of the nvrs as a value.
         advisory_rpms_by_name = {}
         e = Errata()
-        build_nvrs = e.get_builds(errata_id)
+        build_nvrs = e.get_binary_rpm_nvrs(errata_id)
         if build_nvrs:
```
qixiang, line 180: Since we have binary rpms from advisory already, there is no necessary to get binary rpms from brew and then compare. Also you're calling `session.get_build_rpms` with binary rpm nvr provided, which is wrong.

apaplaus: Wait, so we don't need that `_verify_advisory_rpms_in_container_build` method at all or what?

qixiang: We need that, but we don't need to query brew for binary rpms (`build_rpms = session.get_build_rpms(build_nvr)`) now, since we already have the binary rpms which are fetched from errata.

apaplaus: Ok, so I will replace it with this:
```
binary_rpm_nvrs = e.get_binary_rpm_nvrs(errata_id)
if binary_rpm_nvrs:
    for nvr in binary_rpm_nvrs:
         parsed_nvr = rpmlib.parse_nvr(nvr)
         if parsed_nvr['name'] not in advisory_rpms_by_name:
             advisory_rpms_by_name[parsed_nvr['name']] = set()
         advisory_rpms_by_name[parsed_nvr['name']].add(parsed_nvr['nvr'])
```

qixiang:
Can you rebase your PR on top of latest master and fix the CI failure? The ci failure is caused by a new test case in master. Other part looks good to me.

apaplaus:
@qixiang I've done rebasing. Can you please quickly check if it's ok. And check my comment at the top about  **srpm_nevra**' and **srpm_name** fields.

freshmaker/lightblue.py
qixiang: We can keep them at this moment, you can test and remove them in other PR, remember to update the odcs compose releated  code when you're going to remove `srpm_nevra`.

qixiang:
> @qixiang I've done rebasing. Can you please quickly check if it's ok. And check my comment at the top about **srpm_nevra**' and **srpm_name** fields.

Comment added. And now this PR looks good to me.

freshmaker/lightblue.py
apaplaus: Sure, thx

PR #24, opened 2020-09-08
Exclude images from build repositories

qixiang:
Images from build repositories should not be include in our rebuild
list, to filter out these images, we have a new configurable
"IMAGE_BUILD_REPOSITORY_REGISTRIES" which can be set in config.py,
it's a list of build repository registries. If an image's repository
registry is in this list, the image will be excluded.

JIRA: CLOUDWF-3094


qixiang:
Got +1 from Mark, going to merge this.

PR #26, opened 2020-09-18
Fix KeyError when repo has no release_categories

qixiang:
Some repos may not have release_categories, fix KeyError for such repos.


apaplaus:
So, taking into account all the work carried out to eradicate this bug and all the team cooperation necessary to implement the current solution. We can say with confidence and without unnecessary objections that this solution is the most appropriate in this situation. =)
+1

PR #28, opened 2020-09-21
Update README information

mulaievaRH:
Document how to copy the  cassette from OpenShift to the local machine.


mprahl:
@mulaievaRH this seems like documentation that is specific to our internal deployment of Freshmaker. Could you put this in internal documentation instead?

PR #29, opened 2020-09-22
Use publicly available copr repo for python-rhsm

apaplaus:
Now we will use publicly available copr repo for python-rhmsg, because we
need it to use quay.io build. So now we can build Freshmaker image on
quay.io. And so we don't have any internal packages dependencies.

And for now we don't have solution on how to create lable with git commit hash, so I removed that functionality for now. Because Quay doesn't support build-arg for docker build.

RESOLVE: CLOUDWF-2607

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,14 @@
 # TODO: Upgrade to a more recent fedora version
 FROM fedora:32
 
-# The caller should build a Freshmaker RPM package and then pass it in this arg.
-ARG cacert_url=undefined
-ARG commitid=undefined
-
 LABEL \
     name="Freshmaker application" \
     vendor="Freshmaker developers" \
-    license="GPLv2+" \
-    build-date="" \
-    version="$commitid"
+    license="GPLv2+" 
 
 
-# An internal yum repo is needed for rhmsg
-ADD http://download.devel.redhat.com/rel-eng/RCMTOOLS/rcm-tools-fedora.repo /etc/yum.repos.d/
-# ...but the image doesn't have the required root CA installed
-RUN sed -i 's_https://_http://_' /etc/yum.repos.d/rcm-tools-fedora.repo
+# Use Copr repo for python3-rhmsg package
+RUN dnf install -y 'dnf-command(copr)' && dnf copr enable -y qwan/python-rhmsg
 
 COPY yum-packages.txt /tmp/yum-packages.txt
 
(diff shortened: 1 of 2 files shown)


qixiang:
In addition to this change, you need to update the related part of installing ca cert, because we now build the image in quay.io and it can not access internal network. You can install ca cert in container's entrypoint script.

Can you test this change and quay.io build trigger with your forked repo first?

> Now we will use publicly available copr repo for python-rhsm, because we
typo: python-rhmsg

Dockerfile
```diff
@@ -13,10 +13,8 @@ LABEL \
     version="$commitid"
 
 
-# An internal yum repo is needed for rhmsg
-ADD http://download.devel.redhat.com/rel-eng/RCMTOOLS/rcm-tools-fedora.repo /etc/yum.repos.d/
-# ...but the image doesn't have the required root CA installed
-RUN sed -i 's_https://_http://_' /etc/yum.repos.d/rcm-tools-fedora.repo
+# Use Copr repo for python3-rhmsg package
+ADD https://copr.fedorainfracloud.org/coprs/qwan/python-rhmsg/repo/fedora-32/qwan-python-rhmsg-fedora-32.repo /etc/yum.repos.d/
```
qixiang, line 17: Instead of this, you can use "RUN dnf install -y 'dnf-command(copr)' && dnf copr enable -y qwan/python-rhmsg", then when we update our image to latest fedora, we don't need to make any change to this.

apaplaus: > In addition to this change, you need to update the related part of installing ca cert, because we now build the image in quay.io and it can not access internal network. You can install ca cert in container's entrypoint script.

And what script is entrypoint? In Dockerfile it says **fedms-hub-3**, and as I understand it will start consumer, but where this configuration located?

mprahl: @apaplaus I think @qixiang is suggesting that you create a wrapper script  that does this and then runs `exec fedmsg-hub-3` at the end. Greenwave does this:
https://pagure.io/greenwave/blob/master/f/docker/docker-entrypoint.sh

My suggestion is to remove this functionality entirely and just mount the certificate bundle with the internal CAs in the container image at runtime. We do this for many deployments already.

I'll leave it up to you to chose which route you prefer.

Dockerfile
```diff
@@ -1,22 +1,15 @@
...
 
-# The caller should build a Freshmaker RPM package and then pass it in this arg.
-ARG cacert_url=undefined
-ARG commitid=undefined
-
 LABEL \
     name="Freshmaker application" \
     vendor="Freshmaker developers" \
     license="GPLv2+" \
     build-date="" \
```
qixiang, line 8: The trailing "\" is unnecessary now. we can remove the `build-date` line as well since it's just an empty string.

apaplaus: Ok, I will do it

qixiang:
:+1:

PR #31, opened 2020-10-05
Check advisories missed by freshmaker

mulaievaRH:
Check if the RHSA in SHIPPED state  were catched by freshmaker.

JIRA: https://projects.engineering.redhat.com/browse/CLOUDWF-555

### DIFF
--- a/contrib/freshmaker-cli
+++ b/contrib/freshmaker-cli
@@ -18,6 +18,8 @@ Example usage:
         ./freshmaker-cli --redhat --stage --name httpd24-docker
     - Search for events in particular state
         ./freshmaker-cli --redhat --stage --state skipped
+    - Search for RHSA advisories from ET ignored by Freshmaker in the past 5 days
+        ./freshmaker-cli --redhat  --errata --days-ago 5
 """
 
 from __future__ import print_function
@@ -25,7 +27,11 @@ from __future__ import print_function
 import textwrap
 import argparse
 import sys
+import urllib
+from datetime import date, timedelta, datetime
+
 import requests
+from requests_kerberos import HTTPKerberosAuth
 from tabulate import tabulate
 
 # Disable InsecureRequestWarning
@@ -50,6 +56,7 @@ EVENT_TYPES = {
     12: {"name": "ODCS compose state changed", "url": ""},
     13: {"name": "Manual rebuild with advisory",
          "url": "https://errata.devel.redhat.com/advisory/{search_key}"},
+    None: {"name": "No event", "url": ""}
 }
 
 
@@ -59,7 +66,10 @@ EVENT_STATES = [
     "complete",
     "failed",
     "skipped",
-    ]
+    "changed",
+    "no event"
+]
+
 
 def split_nvr_epoch(nvre):
     """Split nvre to N-V-R and E.
@@ -121,6 +131,7 @@ def parse_nvr(nvre):
     result["epoch"] = epoch
     return result
 
+
 def get_api_url(deployment, env):
     API_URLS = {
         "redhat": {
@@ -140,6 +151,19 @@ def get_api_url(deployment, env):
         return None
 
     return envs[env]
+
+
+def get_errata_filter(env):
+    """
+    Returns the address of custom precreated filter which shows the last 100 RHSA in SHIPPED_LIVE
+    state depending on the environment
+    """
+    ERRATA_FILTERS = {
+        "prod": "https://errata.engineering.redhat.com/filter/2964.json",
+        "stage": "https://errata.stage.engineering.redhat.com/filter/1471.json",
+        "dev": "https://errata.devel.redhat.com/filter/2971.json",
+    }
+    return ERRATA_FILTERS.get(env)
 
 
 def _get_freshmaker_items(url, params, limit=None):
@@ -222,6 +246,55 @@ def get_freshmaker_events_by_ids(api_url, ids):
             print(ret["error"])
 
     events = sorted(events, key=lambda k: k['id'])
+    return events
+
+
+def get_advisories_from_errata(errata_url, days=7):
+    """
+    Sends the GET request to the preset Errata filter. Filter shows RHSA in SHIPPED_LIVE state
+    ordered by date of shipping (from the latest to the oldest). Returns the list of advisories
+    that are IMPORTANT or CRITICAL and the date of shipping is older given number of days before the
+    current date.
+    """
+    if days > 7:
+        print("The interval between dates should not exceed 7 days")
+        sys.exit(1)
+    page = 1
+    krb_auth = HTTPKerberosAuth()
+    advisories = []
+    date_from = date.today() - timedelta(days)
+    while True:
+        print("Getting advisories from  Errata page", page)
+        r = requests.get(errata_url, auth=krb_auth, timeout=30, params={page: page})
+        r.raise_for_status()
+        data = r.json()
+        if not data:
+            return advisories
+        for advisory in data:
+            ship_date = datetime.strptime(advisory["timestamps"]["actual_ship_date"], "%Y-%m-%dT%H:%M:%SZ").date()
+            if ship_date < date_from:
+                return advisories
+            if advisory["security_impact"] in ("Important", "Critical"):
+                advisories.append(advisory["id"])
+        page += 1
+
+
+def get_errata_advisories_ignored_by_freshmaker(api_url, advisories):
+    print("Getting Freshmaker events")
+    events = []
+    for advisory in advisories:
+        url = urllib.parse.urljoin(api_url, f'/api/1/events/?search_key={advisory}')
+        r = requests.get(url, timeout=30)
+        r.raise_for_status()
+        ret = r.json()
+        if not ret["items"]:
+            events += [{"id": None,
+                        "state": 6,
+                        "builds": None,
+                        "state_reason": None,
+                        "event_type_id": None,
+                        "search_key": str(advisory),
+                        'builds': []}]
     return events
 
 
@@ -357,6 +430,9 @@ parser.add_argument(
     '--redhat', default=False, action='store_true',
     help="Use the redhat instances.")
 parser.add_argument(
+    '--errata', default=False, action='store_true',
+    help="Check RHSA if freshmaker reacted on them")
+parser.add_argument(
     'events', metavar='event', nargs='*',
     help='Event ids to be queried')
 parser.add_argument(
@@ -377,6 +453,10 @@ parser.add_argument(
 parser.add_argument(
     '--limit', default=10, type=int,
     help="Return only last N events")
+parser.add_argument('--days-ago', default=7, dest="days", type=int, choices=range(1, 8),
+                    help="Number of days in the past for searching from today. Used only with the "
+                         "--errata flag")
+
 args = parser.parse_args()
 
 if args.redhat:
@@ -398,7 +478,20 @@ if not api_url:
     print("There is no Freshmaker instance for deployment %s and env %s" %( deployment, env))
     sys.exit(1)
 
-if not args.events:
+if args.errata:
+    if args.events:
+        print("The flags --errata and --event can't be used simultaneously")
+        sys.exit(1)
+
+    api_errata = get_errata_filter(env)
+    if not api_errata:
+        print("There is no Errata instance for env %s" % (env))
+        sys.exit(1)
+
+    advisories = get_advisories_from_errata(api_errata, args.days)
+    events = get_errata_advisories_ignored_by_freshmaker(api_url, advisories)
+    print_events(events)
+elif not args.events:
     if not args.state:
         args.state = ["init", "building", "complete", "failed"]
     events = get_freshmaker_events(api_url, search_key=args.search_key,


mulaievaRH:
@mprahl , @gnaponie , PTAL

contrib/freshmaker-cli
```diff
@@ -224,6 +229,45 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+    """
+    krb_auth = HTTPKerberosAuth()
+    r = requests.get("https://errata.devel.redhat.com/filter/2943.json", verify=False, auth=krb_auth)
+    r.raise_for_status()
+    data = r.json()
+    advisories = []
+    for advisory in data:
+        if advisory['security_impact'] == "Important" or advisory['security_impact'] == "Critical":
+            advisories.append(advisory['id'])
+    return advisories
```
hluk, line 246: A bit easier to read:

```python
    return [
        advisory["id"]
        for advisory in data
        if advisory["security_impact"] in ("Important", "Critical")
    ]
```

contrib/freshmaker-cli
```diff
@@ -224,6 +229,45 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
     return events
 
+def get_advisories_from_errata():
+    """
+    Sends the GET request to the Errata filter for the latest RHSA, in SHIPPED state. Returns the
+    list of advisories that are IMPORTANT or CRITICAL.
+
+    """
+    krb_auth = HTTPKerberosAuth()
+    r = requests.get("https://errata.devel.redhat.com/filter/2943.json", verify=False, auth=krb_auth)
```
hluk, line 239: The URL should go to configuration.

contrib/freshmaker-cli
```diff
@@ -224,6 +229,45 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+    return advisories
+
+
+def get_freshmaker_events_by_search_keys(api_url, search_keys):
+    print("Getting Freshmaker event id..")
+    events = []
+    for i in search_keys:
+        url = '%s/api/1/events/?search_key=%s' % (api_url, i)
+
+        r = requests.get(url, verify=False)
```
hluk, line 255: Why use `verify=False`?

Why not use request session?

contrib/freshmaker-cli
mulaievaRH: @hluk , what do mean? 
We have no API to list advisories with filters. So I've created the filter.

contrib/freshmaker-cli
mulaievaRH: I have the SSL error without it

contrib/freshmaker-cli
hluk: I thought that there is a configuration, e.g. for different environments (env/stage/prod). The URL should be at least a constant at the top of the script.

mulaievaRH: Added

contrib/freshmaker-cli
mulaievaRH: Thanks. Fixed

contrib/freshmaker-cli
```diff
@@ -356,6 +413,9 @@ parser.add_argument(
 parser.add_argument(
     '--redhat', default=False, action='store_true',
     help="Use the redhat instances.")
+parser.add_argument(
```
mprahl, line 416: This would be more intuitive as a [subcommand](https://docs.python.org/3/library/argparse.html#sub-commands) since this flag changes the entire behavior of the command. It doesn't feel like an option to me.

mprahl: Never mind. I misread the PR.

contrib/freshmaker-cli
```diff
@@ -398,7 +458,12 @@ if not api_url:
     print("There is no Freshmaker instance for deployment %s and env %s" %( deployment, env))
     sys.exit(1)
 
-if not args.events:
+if args.errata:
```
mprahl, line 461: Could you please print an error and exit if both `args.events` and `args.errata` are set?

contrib/freshmaker-cli
mprahl: This is not needed. This is likely because you installed python-requests from pip instead of the Fedora/Red Hat repos. This means that it doesn't use the certificate authorities trusted on your system to determine if the certificate can be trusted. As a work-around, you can run the command with this environment variable set:
```
REQUESTS_CA_BUNDLE=/etc/ssl/certs/ca-bundle.crt
```

contrib/freshmaker-cli
```diff
@@ -224,6 +239,48 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+    """
+    Sends the GET request to the Errata filter for the latest RHSA, in SHIPPED state. Returns the
+    list of advisories that are IMPORTANT or CRITICAL.
+
+    """
+    krb_auth = HTTPKerberosAuth()
+    r = requests.get(errata_url, verify=False, auth=krb_auth)
+    #https://errata.engineering.redhat.com/filter/2958.json - prod
+    #https://errata.devel.redhat.com/filter/2943.json - dev
+    #https://errata.stage.engineering.redhat.com/filter/1469.json - stage
```
mprahl, line 252: This information is in `get_errata_url`. This comment is not needed.
```suggestion
    #https://errata.engineering.redhat.com/filter/2958.json - prod
    #https://errata.devel.redhat.com/filter/2943.json - dev
    #https://errata.stage.engineering.redhat.com/filter/1469.json - stage
```

contrib/freshmaker-cli
```diff
@@ -141,6 +146,16 @@ def get_api_url(deployment, env):
...
+def get_errata_api(env):
+    ERRATA_APIS = {
+                "prod": "https://errata.engineering.redhat.com/filter/2958.json",
+                "stage": "https://errata.stage.engineering.redhat.com/filter/1469.json",
+                "dev": "https://errata.devel.redhat.com/filter/2943.json",
+            }
+    if env not in ERRATA_APIS:
+        return None
+
+    return ERRATA_APIS[env]
```
mprahl, line 158: return ERRATA_APIS.get(env)

contrib/freshmaker-cli
```diff
@@ -141,6 +146,16 @@ def get_api_url(deployment, env):
 
     return envs[env]
 
+def get_errata_api(env):
+    ERRATA_APIS = {
+                "prod": "https://errata.engineering.redhat.com/filter/2958.json",
```
mprahl, line 151: I don't love internal URLs being hard-coded in an upstream repo, but I'll let it be since the precedent has already been set in a prior commit for this script.

contrib/freshmaker-cli
```diff
@@ -141,6 +146,16 @@ def get_api_url(deployment, env):
 
     return envs[env]
 
+def get_errata_api(env):
```
mprahl, line 149: This is a misleading function name since it returns the API for the custom filter you've created. Could you please rename this to make this more clear?

Also, could you please describe what the filter does in the docstring of this function?

contrib/freshmaker-cli
```diff
@@ -398,7 +458,12 @@ if not api_url:
     print("There is no Freshmaker instance for deployment %s and env %s" %( deployment, env))
     sys.exit(1)
 
-if not args.events:
+if args.errata:
+    api_errata = get_errata_api(env)
```
mprahl, line 462: If `api_errata` is `None`, you should print an error and exit.

contrib/freshmaker-cli
```diff
@@ -224,6 +239,48 @@ def get_freshmaker_events_by_ids(api_url, ids):
     events = sorted(events, key=lambda k: k['id'])
     return events
 
+def get_advisories_from_errata(errata_url):
+    """
+    Sends the GET request to the Errata filter for the latest RHSA, in SHIPPED state. Returns the
```
mprahl, line 244:
```suggestion
    Sends the GET request to the Errata filter for the latest RHSA, in SHIPPED_LIVE state. Returns the
```

mprahl: What does "latest" mean? Is it in the past 24 hours?

contrib/freshmaker-cli
```diff
@@ -224,6 +239,48 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+    ]
+
+def get_freshmaker_events_by_search_keys(api_url, search_keys):
+    print("Getting Freshmaker events")
+    events = []
+    for i in search_keys:
+        url = '%s/api/1/events/?search_key=%s' % (api_url, i)
+
+        r = requests.get(url, verify=False)
+        ret = r.json()
```
mprahl, line 269: The rest of the script does this for HTTP request errors.
```suggestion
        r.raise_for_status()
        ret = r.json()
```

Then you don't need the `if not "error" in ret:` if statement.

contrib/freshmaker-cli
```diff
@@ -224,6 +239,48 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+        advisory["id"]
+        for advisory in data
+        if advisory["security_impact"] in ("Important", "Critical")
+    ]
+
+def get_freshmaker_events_by_search_keys(api_url, search_keys):
+    print("Getting Freshmaker events")
+    events = []
+    for i in search_keys:
+        url = '%s/api/1/events/?search_key=%s' % (api_url, i)
```
mprahl, line 266: Optional: The preferred "Pythonic" way to doing this is:
```suggestion
        url = urllib.parse.urljoin(api_url, f'/api/1/events/?search_key={i}')
```

contrib/freshmaker-cli
```diff
@@ -224,6 +239,48 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+    return [
+        advisory["id"]
+        for advisory in data
+        if advisory["security_impact"] in ("Important", "Critical")
+    ]
+
+def get_freshmaker_events_by_search_keys(api_url, search_keys):
+    print("Getting Freshmaker events")
+    events = []
+    for i in search_keys:
```
mprahl, line 265: Optional: This makes it easier to read:
```suggestion
    for search_key in search_keys:
```

contrib/freshmaker-cli
```diff
@@ -224,6 +239,48 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+        url = '%s/api/1/events/?search_key=%s' % (api_url, i)
+
+        r = requests.get(url, verify=False)
```
mprahl, line 268: It'd be nice if this timed out after 30 seconds of not hearing a response. If you don't specify a timeout and the server doesn't respond, this line of code will just hang forever.
```suggestion
        r = requests.get(url, timeout=30)
```

contrib/freshmaker-cli
mulaievaRH: I had also this question, what does "latest" mean. It's written like this in the JIRA card. 
But as we use the custom filter, which is sorted by newest and this filter shows us only 20 advisories (on the first page which basically we use ) I've decided don't check any dates and check all of this 20 advisories.

mprahl: @mulaievaRH who created the filter? You can ask them and then document it.

mprahl: Never mind. It looks like you documented it. Thanks!

contrib/freshmaker-cli
```diff
@@ -59,6 +65,7 @@ EVENT_STATES = [
     "complete",
     "failed",
     "skipped",
+    "no event"
```
mprahl, line 68: Why is this added?

I can't find this in the Freshmaker codebase.

contrib/freshmaker-cli
```diff
@@ -50,6 +55,7 @@ EVENT_TYPES = {
     12: {"name": "ODCS compose state changed", "url": ""},
     13: {"name": "Manual rebuild with advisory",
          "url": "https://errata.devel.redhat.com/advisory/{search_key}"},
+    100: {"name": "No event", "url": ""}
```
mprahl, line 58: This looks like a fake event type. How about making the key `None`? Then you can set `"event_type_id": 100,` to `"event_type_id": None,` which seems more natural.

contrib/freshmaker-cli
```diff
@@ -225,6 +244,42 @@ def get_freshmaker_events_by_ids(api_url, ids):
     return events
 
 
+def get_advisories_from_errata(errata_url):
+    """
+    Sends the GET request to the Errata filter for the latest RHSA, in SHIPPED_LIVE state. Returns
+    the list of advisories that are IMPORTANT or CRITICAL.
+    """
+    krb_auth = HTTPKerberosAuth()
+    r = requests.get(errata_url, verify=False, auth=krb_auth)
```
mprahl, line 253: Please remove `verify=False` and add a timeout.
```suggestion
    r = requests.get(errata_url, auth=krb_auth, timeout=30)
```

contrib/freshmaker-cli
```diff
@@ -225,6 +244,42 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+    for search_key in search_keys:
+        url = urllib.parse.urljoin(api_url, f'/api/1/events/?search_key={search_key}')
+        r = requests.get(url, timeout=30)
+        r.raise_for_status()
+        ret = r.json()
+        if ret["items"]:
+           events += ret["items"]
+        else:
+           events+=[{"id": None,
+                     "state": 5,
```
mprahl, line 275: Doesn't `5` mean `CANCELED`? This doesn't seem right in this instance.

contrib/freshmaker-cli
```diff
@@ -398,7 +456,20 @@ if not api_url:
     print("There is no Freshmaker instance for deployment %s and env %s" %( deployment, env))
     sys.exit(1)
 
-if not args.events:
+if args.errata:
+    if args.events:
+        print("Parameters --errata and event can't be used simultaneously")
```
mprahl, line 461:
```suggestion
        print("Parameters --errata and --event can't be used simultaneously")
```

gnaponie:
@mulaievaRH could you squash the commits and add the card reference to the commit message?

contrib/freshmaker-cli
```diff
@@ -225,6 +244,42 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+def get_freshmaker_events_by_search_keys(api_url, search_keys):
+    print("Getting Freshmaker events")
+    events = []
+    for search_key in search_keys:
+        url = urllib.parse.urljoin(api_url, f'/api/1/events/?search_key={search_key}')
+        r = requests.get(url, timeout=30)
+        r.raise_for_status()
+        ret = r.json()
+        if ret["items"]:
+           events += ret["items"]
```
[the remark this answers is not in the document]
gnaponie: IMHO it's maybe clearer to get just the ones that were missed, that was the point of the card. This should substitute the manual process of going to ET, check the latest shipped advisories that are important and critical and then for each one of them check Freshmaker's API.... The outcome should be just the ones that were ignored in the end, I don't care that much about the ones that were processed, otherwise I'd still have to "manually" filter out those ones.

contrib/freshmaker-cli
```diff
@@ -18,14 +18,19 @@ Example usage:
         ./freshmaker-cli --redhat --stage --name httpd24-docker
     - Search for events in particular state
         ./freshmaker-cli --redhat --stage --state skipped
+    - Search if freshmaker reacted on all  RHSA, in SHIPPED_LIVE state that are IMPORTANT or CRITICAL
```
gnaponie, line 21: I guess it would be preferable something like "Search for advisories ignored by Freshmaker"

contrib/freshmaker-cli
mulaievaRH: I added in EVENT_STATES the fake state "no event". "state":5 corresponds to this value

contrib/freshmaker-cli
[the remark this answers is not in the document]
mulaievaRH: @gnaponie . PTAL now

contrib/freshmaker-cli
```diff
@@ -18,14 +18,20 @@ Example usage:
         ./freshmaker-cli --redhat --stage --name httpd24-docker
     - Search for events in particular state
         ./freshmaker-cli --redhat --stage --state skipped
+    - Search for RHSA ignored by Freshmaker
+        ./freshmaker-cli --redhat  --errata
```
gnaponie, line 22: Maybe it would be nice to show the example of how to pass the range dates.
Also it would be nice if it would be clearer that with this combination you get only the ignored advisories.

contrib/freshmaker-cli
```diff
@@ -142,6 +152,19 @@ def get_api_url(deployment, env):
     return envs[env]
 
 
+def get_errata_filter(env):
+    """
+    Returns the address of custom precreated filter which shows the last 20 RHSA in SHIPPED_LIVE
```
gnaponie, line 157: The comment says 20, but I believe it is 100.

contrib/freshmaker-cli
```diff
@@ -225,6 +248,49 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+    """
+    if (date_to - date_from).days > 7:
+        print("The interval between dates should not exceed 7 days")
+        sys.exit(1)
+
+    krb_auth = HTTPKerberosAuth()
+    r = requests.get(errata_url, auth=krb_auth, timeout=30)
+    r.raise_for_status()
+    data = r.json()
+    return [
```
gnaponie, line 266: With the current amount of advisories shipped by ET it seems unlikely that FM will perform more than one request to get the latest advisories (within the maximum set to one week). But in the future this amount might increase... We could consider at least checking if we got all the results and see if a new request would be necessary. @mprahl wdyt? Should we still check the second page or is it so unlikely that this will happen that we shouldn't bother?

contrib/freshmaker-cli
```diff
@@ -225,6 +248,49 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+        if advisory["security_impact"] in ("Important", "Critical") and
+           date_from <= datetime.strptime(advisory["timestamps"]["issue_date"],
+                                          "%Y-%m-%dT%H:%M:%SZ").date() <= date_to
+    ]
+
+
+def get_errata_advisories_ignored_by_freshmaker(api_url, search_keys):
+    print("Getting Freshmaker events")
+    events = []
+    for search_key in search_keys:
```
gnaponie, line 278: Optional: rename `search_key` and `search_keys` to `advisory` and `advisories`. I personally hate this `search_key` :D and I think it creates a lot of confusion in new developers, so maybe we should try using it as less as possible in the future. But this is optional.

contrib/freshmaker-cli
mulaievaRH: Wanted to add. As we have in the filter 100 advisories, the oldest one is 1.5 month old. Will we need some older advisories ?

gnaponie: I don't think we'll need older advisories. But my concern is that with time the number of advisories filed in ET could increase and maybe we'll need pagination. That's why I'd be interested in hearing @mprahl on this. If we think it's unlikely that ET will speed up so fast in years to come, then we can just leave it like this.

contrib/freshmaker-cli
```diff
@@ -225,6 +248,49 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
 
 
+def get_advisories_from_errata(errata_url, date_from=date.today() - timedelta(7),
+                               date_to=date.today()):
+    """
+    Sends the GET request to the Errata filter for the latest RHSA, in SHIPPED_LIVE state. Returns
+    the list of advisories that are IMPORTANT or CRITICAL and the issue date is between specified
+    dates.
+    """
+    if (date_to - date_from).days > 7:
```
mprahl, line 258: Why is this restricted?

contrib/freshmaker-cli
mprahl: Can the filter be set to be ordered by the timestamp of when it shipped? Then you can keep going to the next page until you encounter an advisory that was shipped outside of your date range.

mprahl, line 266: Optional: This is a very long list comprehension and makes it hard to read. Consider turning this into a normal for loop for readability.

contrib/freshmaker-cli
```diff
@@ -50,6 +56,7 @@ EVENT_TYPES = {
     12: {"name": "ODCS compose state changed", "url": ""},
     13: {"name": "Manual rebuild with advisory",
          "url": "https://errata.devel.redhat.com/advisory/{search_key}"},
+    "None": {"name": "No event", "url": ""}
```
mprahl, line 59: This is more natural
```suggestion
    None: {"name": "No event", "url": ""}
```

contrib/freshmaker-cli
```diff
@@ -225,6 +248,49 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+        url = urllib.parse.urljoin(api_url, f'/api/1/events/?search_key={advisory}')
+        r = requests.get(url, timeout=30)
+        r.raise_for_status()
+        ret = r.json()
+        if not ret["items"]:
+            events += [{"id": None,
+                        "state": 5,
+                        "builds": None,
+                        "state_reason": None,
+                        "event_type_id": "None",
```
mprahl, line 288: Based on my above suggestion
```suggestion
                        "event_type_id": None,
```

contrib/freshmaker-cli
mulaievaRH: We've discussed with @gnaponie , that it's not necessary find advisories in time period more than 7 days

gnaponie: @mprahl The purpose of this feature is to be able to check if FM missed any advisory. When you do this process manually you don't need too back in time. So we figured one week might be enough for this purpose. Do you think we should remove the restriction?

gnaponie: If we introduce pagination we can get rid of this restriction.

contrib/freshmaker-cli
mulaievaRH: Done

mulaievaRH: This filters can be set to be ordered by Release date or Last modified. Or The Newest/Oldest of course (now they are ordered by Newest )

contrib/freshmaker-cli
mprahl: I don't really get the need for a date range. Why not just allow the script to specify how many days to go back and max it out to 7?

contrib/freshmaker-cli
```diff
@@ -225,6 +248,49 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+
+    krb_auth = HTTPKerberosAuth()
+    r = requests.get(errata_url, auth=krb_auth, timeout=30)
+    r.raise_for_status()
+    data = r.json()
+    advisories = []
+    for advisory in data:
+        if advisory["security_impact"] in ("Important", "Critical") and \
+                date_from <= datetime.strptime(advisory["timestamps"]["issue_date"],
+                                               "%Y-%m-%dT%H:%M:%SZ").date() <= date_to:
```
mprahl, line 270:
```suggestion
        issue_date = datetime.strptime(advisory["timestamps"]["issue_date"], "%Y-%m-%dT%H:%M:%SZ").date()
        if advisory["security_impact"] in ("Important", "Critical") and date_from <= issue_date <= date_to:
```

contrib/freshmaker-cli
mprahl: @mulaievaRH I understand, but 5 means Canceled in Freshmaker:
https://github.com/redhat-exd-rebuilds/freshmaker/blob/40e36b671fe123f2e1427ef89b00fd4896f613f2/freshmaker/types.py#L91-L92

You'll need a value that doesn't overlap.

mulaievaRH: Oh. Understood now. Will be changed.

contrib/freshmaker-cli
gnaponie: Imagine that you are debugging some new card (that was filed a couple of days ago), you are interested only to that date range, not all days. But, yeah, it's not a big deal, also the num of days to go back is also good.

mprahl: @gnaponie unless you can filter by a date range in ET, you're not gaining much efficiency by restricting the number of days since you have to go through all the pages to find the page that starts with that date anyways. Since ET sorts the dates (I think that was said earlier in this PR), you could do a slightly modified version of the binary search algorithm, but that seems a little complex for a simple script.

This is why I proposed just having x number of days to go back with a maximum.

mulaievaRH: @mprahl . @gnaponie . PTAL. I've changed it now

contrib/freshmaker-cli
```diff
@@ -142,6 +153,19 @@ def get_api_url(deployment, env):
     return envs[env]
 
 
+def get_errata_filter(env):
+    """
+    Returns the address of custom precreated filter which shows the last 100 RHSA in SHIPPED_LIVE
+    state depending on the environment
+    """
+    ERRATA_FILTERS = {
+        "prod": "https://errata.engineering.redhat.com/filter/2964.json",
```
mprahl, line 162: When looking at the JSON, I don't see it as ordered by the date the advisory was shipped. How is it ordered?

The ordering must be correct in order for your code to find the advisories to work.

contrib/freshmaker-cli
```diff
@@ -225,6 +249,48 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+    Sends the GET request to the Errata filter for the latest RHSA, in SHIPPED_LIVE state. Returns
+    the list of advisories that are IMPORTANT or CRITICAL and the issue date is between specified
+    dates.
+    """
+    if days > 7:
+        print("The interval between dates should not exceed 7 days")
+        sys.exit(1)
+
+    krb_auth = HTTPKerberosAuth()
+    r = requests.get(errata_url, auth=krb_auth, timeout=30)
```
mprahl, line 263: This doesn't perform pagination. I thought we wanted to do that to ensure all advisories in the specified date range exists.

contrib/freshmaker-cli
```diff
@@ -377,6 +446,9 @@ parser.add_argument(
 parser.add_argument(
     '--limit', default=10, type=int,
     help="Return only last N events")
+parser.add_argument('--days-behind', default=7, dest="days", type=int,
```
mprahl, line 449:
```suggestion
parser.add_argument('--days-ago', default=7, dest="days", type=int,
```

contrib/freshmaker-cli
```diff
@@ -377,6 +446,9 @@ parser.add_argument(
...
+                    help="Number of days in the past for searching from today")
```
mprahl, line 450: Could you document that this is only used with the `--errata` flag?

contrib/freshmaker-cli
```diff
@@ -225,6 +249,48 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+        sys.exit(1)
+
+    krb_auth = HTTPKerberosAuth()
+    r = requests.get(errata_url, auth=krb_auth, timeout=30)
+    r.raise_for_status()
+    data = r.json()
+    advisories = []
+    for advisory in data:
+        issue_date = datetime.strptime(advisory["timestamps"]["issue_date"], "%Y-%m-%dT%H:%M:%SZ").date()
+        date_from = date.today()-timedelta(days)
```
mprahl, line 269:
```suggestion
        date_from = date.today() - timedelta(days)
```

contrib/freshmaker-cli
hluk: Pagination can be done with simple helper function:

```python
def get_paginated_items(session, url):
    page = 0
    while True:
        page += 1
        r = session.get(url, params={page: page})
        r.raise_for_status()
        data = r.json()
        if not data:
            break
        for advisory in data:
            yield data()
```

gnaponie:
Tested locally and it looks good. :+1: @mprahl do you want to review?
BTW I saw there's no testing for the whole cli, I've filed a tech debt card to fix that in the future -- even though I'm not sure we'll ever have the capacity to pick it up. :)

contrib/freshmaker-cli
```diff
@@ -225,6 +250,53 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+        print("The interval between dates should not exceed 7 days")
+        sys.exit(1)
+    page = 1
+    while True:
+        print("Getting advisories from  Errata page", page)
+        krb_auth = HTTPKerberosAuth()
+        r = requests.get(errata_url, auth=krb_auth, timeout=30, params={page: page})
+        r.raise_for_status()
+        data = r.json()
+        advisories = []
```
mprahl, line 270: You need to define this outside of the while loop or else you only get the advisories that matched on the last page that was looked at.

contrib/freshmaker-cli
```diff
@@ -225,6 +250,53 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+    that are IMPORTANT or CRITICAL and the date of shipping is older given number of days before the
+    current date.
+    """
+    if days > 7:
+        print("The interval between dates should not exceed 7 days")
+        sys.exit(1)
+    page = 1
+    while True:
+        print("Getting advisories from  Errata page", page)
+        krb_auth = HTTPKerberosAuth()
```
mprahl, line 266: I would move this outside of the while loop to avoid having to redefine it all the time.

contrib/freshmaker-cli
```diff
@@ -225,6 +250,53 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+        r = requests.get(errata_url, auth=krb_auth, timeout=30, params={page: page})
+        r.raise_for_status()
+        data = r.json()
+        advisories = []
+        date_from = date.today() - timedelta(days)
+        for advisory in data:
```
mprahl, line 272: We should guard against the Errata Tool returning no advisories on the page we are querying.
```suggestion
        if not data:
            return advisories
        for advisory in data:
```

contrib/freshmaker-cli
```diff
@@ -18,14 +18,21 @@ Example usage:
         ./freshmaker-cli --redhat --stage --name httpd24-docker
     - Search for events in particular state
         ./freshmaker-cli --redhat --stage --state skipped
+    - Search for RHSA advisories from ET ignored by Freshmaker from 5 days before to current date
+       REQUESTS_CA_BUNDLE=/etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt ./freshmaker-cli --redhat  --errata --days-ago 5
+       (REQUESTS_CA_BUNDLE should be passed to avoid SSL errors)
```
mprahl, line 23: This only applies if the requests library was installed via pip and not via a system package manager such as dnf. I would remove mentioning this environment variable since it's system specific.

mulaievaRH:
@mprahl , all is fixed. Review please

contrib/freshmaker-cli
```diff
@@ -18,14 +18,20 @@ Example usage:
         ./freshmaker-cli --redhat --stage --name httpd24-docker
     - Search for events in particular state
         ./freshmaker-cli --redhat --stage --state skipped
+    - Search for RHSA advisories from ET ignored by Freshmaker from 5 days before to current date
```
mprahl, line 21:
```suggestion
    - Search for RHSA advisories from ET ignored by Freshmaker in the past 5 days
```

contrib/freshmaker-cli
```diff
@@ -225,6 +249,55 @@ def get_freshmaker_events_by_ids(api_url, ids):
...
+        sys.exit(1)
+    page = 1
+    krb_auth = HTTPKerberosAuth()
+    advisories = []
+    while True:
+        print("Getting advisories from  Errata page", page)
+        r = requests.get(errata_url, auth=krb_auth, timeout=30, params={page: page})
+        r.raise_for_status()
+        data = r.json()
+        date_from = date.today() - timedelta(days)
```
mprahl, line 270: Could you move this outside of the loop? Sorry for missing this earlier.

contrib/freshmaker-cli
```diff
@@ -377,6 +453,10 @@ parser.add_argument(
 parser.add_argument(
     '--limit', default=10, type=int,
     help="Return only last N events")
+parser.add_argument('--days-ago', default=7, dest="days", type=int,
```
mprahl, line 456: Rather than validate the value in `get_advisories_from_errata`, you can just do the following. It would also guard against someone entering a value less than 1.
```suggestion
parser.add_argument('--days-ago', default=7, dest="days", type=int, choices=range(1, 8),
```

mprahl (changes requested):
Just a few comments to address. Once addressed, this will be good to merge.

contrib/freshmaker-cli
mulaievaRH: Fixed :)

mprahl (approved):

PR #33, opened 2020-10-13
Botas advisory handler

apaplaus:
Add new handler to handle advisories in SHIPPED_LIVE state
that were filed by BOTAS. We need that for future support of Operators CVE
rebuild functionality. To support rebuilding of Bundle images.
+ small badge to see container build status in Quay

RESOLVE: CLOUDWF-3611

### DIFF
--- a/dev_scripts/botas_advisory_test.py
+++ b/dev_scripts/botas_advisory_test.py
@@ -1,0 +1,53 @@
+#!/usr/bin/env python
+"""
+Example usage:
+    ./botas_advisory_test.py ADVISORY_ID
+    ./botas_advisory_test.py 59808
+"""
+import os
+import sys
+from unittest.mock import patch
+from logging.config import dictConfig
+import fedmsg.config
+
+# Allow imports from parent directory.
+sys.path.insert(1, os.path.join(sys.path[0], '..'))
+
+# Set the FRESHMAKER_DEVELOPER_ENV variable.
+os.environ["FRESHMAKER_DEVELOPER_ENV"] = "1"
+os.environ["FRESHMAKER_CONFIG_FILE"] = os.path.join(sys.path[0], "config.py")
+os.environ["REQUESTS_CA_BUNDLE"] = "/etc/ssl/certs/ca-bundle.crt"
+
+from freshmaker import db, app
+from freshmaker.events import BotasErrataShippedEvent
+from freshmaker.handlers.botas.botas_shipped_advisory import HandleBotasAdvisory
+from freshmaker.errata import Errata, ErrataAdvisory
+
+
+fedmsg_config = fedmsg.config.load_config()
+dictConfig(fedmsg_config.get('logging', {'version': 1}))
+
+if len(sys.argv) != 2:
+    print("Usage: ./botas_advisory_test.py ADVISORY_ID")
+    sys.exit(1)
+
+app_context = app.app_context()
+app_context.__enter__()
+
+db.drop_all()
+db.create_all()
+db.session.commit()
+
+advisory_id = sys.argv[1]
+errata = Errata()
+advisory = ErrataAdvisory.from_advisory_id(errata, advisory_id)
+
+event = BotasErrataShippedEvent(msg_id='fake-msg', advisory=advisory,
+                                dry_run=True)
+
+handler = HandleBotasAdvisory()
+with patch("freshmaker.consumer.get_global_consumer"):
+    if handler.can_handle(event):
+        handler.handle(event)
+    else:
+        print("Can't handle that event")
--- a/freshmaker/events.py
+++ b/freshmaker/events.py
@@ -452,3 +452,10 @@ class FreshmakerAsyncManualBuildEvent(BaseEvent):
         self.brew_target = brew_target
         self.requester = requester
         self.requester_metadata_json = requester_metadata_json
+
+
+class BotasErrataShippedEvent(ErrataBaseEvent):
+    """ Event triggered, when BOTAS pushes advisory to SHIPPED_LIVE state """
+
+    def __init__(self, msg_id, advisory, dry_run=False):
+        super().__init__(msg_id, advisory, dry_run=dry_run)
--- a/freshmaker/handlers/botas/botas_shipped_advisory.py
+++ b/freshmaker/handlers/botas/botas_shipped_advisory.py
@@ -1,0 +1,76 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) 2020  Red Hat, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from freshmaker import db
+from freshmaker.handlers import ContainerBuildHandler
+from freshmaker.events import BotasErrataShippedEvent
+from freshmaker.models import Event
+from freshmaker.types import EventState
+
+
+class HandleBotasAdvisory(ContainerBuildHandler):
+    """
+    Handles event that was created by transition of an advisory filed by
+    BOTAS to SHIPPED_LIVE state
+    """
+    name = "HandleBotasAdvisory"
+
+    def can_handle(self, event):
+        if (isinstance(event, BotasErrataShippedEvent) and
+                'docker' in event.advisory.content_types):
+            return True
+
+        return False
+
+    def handle(self, event):
+        if event.dry_run:
+            self.force_dry_run()
+        self.event = event
+
+        # Get event from database or create new one.
+        # Then we can get original NVRs from it.
+        db_event = Event.get_or_create_from_event(db.session, event)
+
+        self.set_context(db_event)
+
+        # Check if event is allowed by internal policies
+        if not self.event.is_allowed(self):
+            msg = ("This image rebuild is not allowed by internal policy. "
+                   f"message_id: {event.msg_id}")
+            db_event.transition(EventState.SKIPPED, msg)
+            db.session.commit()
+            self.log_info(msg)
+            return []
+
+        # Get original nvrs of all builds in the advisory
+        original_nvrs = [build.original_nvr for build in db_event.builds]
+        self.log_info(
+            "Orignial nvrs of build in the advisory #{0} are: {1}".format(
+                db_event.search_key, " ".join(original_nvrs)))
+
+        msg = "Skipping due to being blocked on further implementation for now."
+
+        # Skip that event because we can't proceed with processing it.
+        # TODO
+        # Next step would be queries Pyxis and getting the digests of these nvrs
+        db_event.transition(EventState.SKIPPED, msg)
+        db.session.commit()
+        return []
--- a/freshmaker/parsers/errata/state_change.py
+++ b/freshmaker/parsers/errata/state_change.py
@@ -20,13 +20,14 @@
 # SOFTWARE.
 
 from freshmaker.parsers import BaseParser
-from freshmaker.events import ErrataAdvisoryStateChangedEvent
+from freshmaker.events import ErrataAdvisoryStateChangedEvent, BotasErrataShippedEvent
 from freshmaker.errata import Errata, ErrataAdvisory
 
 
 class ErrataAdvisoryStateChangedParser(BaseParser):
     """
     Parser parsing errata.activity.status.
+    If event produced by BOTAS we will generate specific type of event
     """
 
     name = "ErrataAdvisoryStateChangedParser"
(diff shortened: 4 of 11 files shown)


freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -0,0 +1,70 @@
...
+
+        db.session.commit()
+        self.set_context(db_event)
+
+        # Get original nvrs of all builds in the advisory
+        original_nvrs = [build.original_nvr for build in db_event.builds]
+        self.log_info("Orignial nvrs of build in the advisory #{0} are: {1}".format(
+            db_event.search_key, " ".join(original_nvrs)))
+
+        msg = "We are skipping that event, because we are blocked in further" \
```
mprahl, line 62:
```suggestion
        msg = "Skipping due to being blocked for further" \
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -0,0 +1,70 @@
...
+from freshmaker.types import EventState
+
+
+class HandleBotasAdvisory(ContainerBuildHandler):
+    """
+    Handles event that was created by transition of an advisory filed by
+    BOTAS to SHIPPED_LIVE state
+    """
+    name = "HandleBotasAdvisory"
+    order = 0
```
mprahl, line 35: The default order seems to be `50` and it just needs to be lower than `RebuildImagesOnRPMAdvisoryChange` I believe. Can you put it a higher number but less than 50? This will give us some more room for adding handlers that need to go ahead of this handler but not having to go negative on the order.

freshmaker/events.py
```diff
@@ -452,3 +452,11 @@ def __init__(self, msg_id, dist_git_branch, container_images,
         self.brew_target = brew_target
         self.requester = requester
         self.requester_metadata_json = requester_metadata_json
+
+
+class BotasChangeStateEvent(ErrataBaseEvent):
```
mprahl, line 457: Since this event is only produced when the advisory is shipped, let's rename this event so it's a little clearer.
```suggestion
class BotasErrataShippedEvent(ErrataBaseEvent):
```

freshmaker/events.py
```diff
@@ -452,3 +452,11 @@ def __init__(self, msg_id, dist_git_branch, container_images,
...
+    """ Event triggered, when BOTAS pushes advisory to SHIPPED_LIVE state """
+
+    def __init__(self, msg_id, advisory, dry_run=False):
+        super(BotasChangeStateEvent, self).__init__(msg_id, advisory,
```
mprahl, line 461: Optional: Since we are using Python 3, we don't need to specify the class and the instance.
```suggestion
        super().__init__(msg_id, advisory,
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -0,0 +1,70 @@
...
+        # Generate the Database representation of `event`, it can be
+        # triggered by user, we want to track what happened
+        db_event = Event.get(db.session, event.msg_id)
+
+        if not db_event:
+            self.log_warn("Can't find in DB event with id {}".format(
+                event.msg_id))
+            return
+
+        db.session.commit()
```
mprahl, line 54: Is there a reason that you are committing the transaction here even though you haven't made any database changes in this method?

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -0,0 +1,70 @@
...
+        return isinstance(event, BotasChangeStateEvent)
+
+    def handle(self, event):
+        if event.dry_run:
+            self.force_dry_run()
+        self.event = event
+
+        # Generate the Database representation of `event`, it can be
+        # triggered by user, we want to track what happened
+        db_event = Event.get(db.session, event.msg_id)
```
mprahl, line 47: Will this event exist in the database here? This would be the first time that the event is seen by Freshmaker. 

You'll likely need to perform something like this:
https://github.com/redhat-exd-rebuilds/freshmaker/blob/40e36b671fe123f2e1427ef89b00fd4896f613f2/freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py#L70-L73

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -0,0 +1,70 @@
...
+
+        msg = "We are skipping that event, because we are blocked in further" \
+              " implementation for now."
+
+        # Skip that event because we can't proceed with processing it.
+        # TODO
+        # Next step would be queries Pyxis and getting the digests of these nvrs
+        db_event.transition(EventState.SKIPPED, msg)
+        db.session.commit()
+        return original_nvrs
```
mprahl, line 70: This method should return an empty list or a list of event objects to process:
https://github.com/redhat-exd-rebuilds/freshmaker/blob/40e36b671fe123f2e1427ef89b00fd4896f613f2/freshmaker/handlers/__init__.py#L255-L265
https://github.com/redhat-exd-rebuilds/freshmaker/blob/40e36b671fe123f2e1427ef89b00fd4896f613f2/freshmaker/consumer.py#L159-L174

FYI, this is a bad practice since the messages are added to a queue that lives in memory so they will survive a restart., but that's current architecture and there's nothing we can do about it right now.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -0,0 +1,70 @@
...
+class HandleBotasAdvisory(ContainerBuildHandler):
+    """
+    Handles event that was created by transition of an advisory filed by
+    BOTAS to SHIPPED_LIVE state
+    """
+    name = "HandleBotasAdvisory"
+    order = 0
+
+    def can_handle(self, event):
+        return isinstance(event, BotasChangeStateEvent)
```
mprahl, line 38: Could you also add a check to ensure that `docker` is in the content types of the advisory? This is to be defensive in the event that the scope of BOTAS changes to also ship other content types.

freshmaker/parsers/errata/state_change.py
```diff
@@ -41,6 +42,12 @@ def parse(self, topic, msg):
...
         errata = Errata()
-        return ErrataAdvisoryStateChangedEvent(
-            msg_id, ErrataAdvisory.from_advisory_id(errata, errata_id)
-        )
+        advisory = ErrataAdvisory.from_advisory_id(errata, errata_id)
+        # If advisory created by BOTAS and it's shipped,
+        # then return BotasChangeStateEvent event
+        if advisory.state == "SHIPPED_LIVE" and \
+           advisory.reporter.startswith('botas'):
+            event = BotasChangeStateEvent(msg_id, advisory)
```
mprahl, line 50: @gnaponie do you like this approach of having a separate event? Alternatively, the new handler could have a lower `order` to be processed first and return `False` in its `can_handle` method if the reporter is not BOTAS.

freshmaker/events.py
apaplaus: Yes, but since we are using that kind of call everywhere I thought it would be better to have it the same way. But you are right it's more convenient to have it with just 'super()'. I will change it.

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: So as I understood, it should be in DB, that's why there is only 'get()' without creating new event. But I am not sure. Maybe @gnaponie could help with this one?

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: Forgot to remove it, as you said it would be useful only if we created new event in db.

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: Ok, didn't see it. I thought the return value isn't used so I can use it for testing, I will remove it. Btw do you know some way how to test that 'original_nvrs' variable except logging?

apaplaus:
And do you know why travis is complaining about my script in **dev_scripts**. Doesn't he supposed to ignore everything there?

freshmaker/events.py
mprahl: You can keep it the other way for consistency. Whichever way you prefer.

freshmaker/handlers/botas/botas_shipped_advisory.py
mprahl: I can't really think of a good way. It's fine to keep that part untested for now since it's just a simple list comprehension. It'll just be tested when this handler is expanded.

mprahl:
> And do you know why travis is complaining about my script in **dev_scripts**. Doesn't he supposed to ignore everything there?

I think pytest detects it since it has `test` in the module name. You can modify this to ignore the `dev_scripts` directory:
https://github.com/redhat-exd-rebuilds/freshmaker/blob/40e36b671fe123f2e1427ef89b00fd4896f613f2/tox.ini#L65

gnaponie:
You might want to add somewhere something like:
`        if not self.event.is_allowed(self, ArtifactType.IMAGE):
            msg = ("This image rebuild is not allowed by internal policy. "
                   f"message_id: {event.msg_id}")
            db.session.commit()
            self.log_info(msg)
            return []
`

This should check Freshmaker configuration and check if we disabled that for some reason.

dev_scripts/botas_advisory_test.py
```diff
@@ -0,0 +1,53 @@
...
+os.environ["FRESHMAKER_CONFIG_FILE"] = os.path.join(sys.path[0], "config.py")
+os.environ["REQUESTS_CA_BUNDLE"] = "/etc/ssl/certs/ca-bundle.crt"
+
+from freshmaker import db, app
+from freshmaker.events import BotasErrataShippedEvent
+from freshmaker.handlers.botas.botas_shipped_advisory import HandleBotasAdvisory
+from freshmaker.errata import Errata, ErrataAdvisory
+
+
+fedmsg_config = fedmsg.config.conf
```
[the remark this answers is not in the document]
apaplaus: If I use `fedmsg.config.load_config()` pycharm says **Using "load_config" is deprecated and will be removed in a future release; use the "fedmsg.config.conf" dictionary instead. ** But you are right, without that it doesn't work as expected. I will change it.

tests/handlers/botas/test_botas_shipped_advisory.py
```diff
@@ -0,0 +1,75 @@
...
+        self.assertFalse(handler.can_handle(event))
+
+    @patch("freshmaker.handlers.botas.botas_shipped_advisory.HandleBotasAdvisory.log_warn")
+    def test_no_db_event(self, log_warn):
+        event = BotasErrataShippedEvent("123", self.botas_advisory)
+        handler = HandleBotasAdvisory()
+        handler.handle(event)
+
+        self.assertIn("Can't find in DB event with id 123",
+                      log_warn.call_args[0])
```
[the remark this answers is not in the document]
apaplaus: Yes it worked for me. Because that was actually test to check that if there is no event in db it will print error. I will update it.

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: Well maybe I misunderstood something, but we were talking about situation, when 'FM rebuild operator/operand -> advisory created by BOTAS -> msg published to umb when that advisory moved to SHIPPED_LIVE state -> FM gets this message -> new event created -> botas handler handles that event'. But will that new event have information about builds made in before? So when I call something like  `new_db_event.builds`, will there be builds from previous event/advisory?

gnaponie: @apaplaus If I understood correctly your point: the builds made for the operand/operator are not the same event/builds that will be done for the bundle images. Even though they are connected, they are still different container images.
The correct flow is: FM rebuild operator/operand -> advisory created by BOTAS -> msg published to umb when that advisory moved to SHIPPED_LIVE state -> FM gets this message (with this new botas handler that you wrote) -> new event created -> FM builds the bundle image

apaplaus: Ok, so if they are connected and I can get nvrs in that way, how I am doing it now, everything is fine.

freshmaker/parsers/errata/state_change.py
gnaponie: I guess it was just a logic separation, also because the card says "For this handler a new type of Freshmaker event should be created".

gnaponie:
The MR looks good to me now. Perhaps wait for @mprahl's review, but I'm ok with merging this.

gnaponie (approved):

tests/test_errata.py
```diff
@@ -79,7 +79,10 @@ def __init__(self, errata_rest_get, errata_http_get=None):
             "product": {
                 "id": 89,
                 "short_name": "product",
-            }
+            },
+            "people": {
+                "reporter": "botas/pnt-devops-rad-jenkins.rhev-ci-vms.eng.rdu2.redhat.com@REDHAT.COM"
```
qixiang, line 84: We should avoid having internal info like this in code.

apaplaus: Ok I will change it

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -0,0 +1,76 @@
...
+    BOTAS to SHIPPED_LIVE state
+    """
+    name = "HandleBotasAdvisory"
+
+    def can_handle(self, event):
+        if (isinstance(event, BotasErrataShippedEvent) and
+                'docker' in event.advisory.content_types):
+            return True
+
+        return False
```
mprahl, line 41: Optional:
```suggestion
        return (isinstance(event, BotasErrataShippedEvent) and
                   'docker' in event.advisory.content_types)
```

tests/handlers/botas/test_botas_shipped_advisory.py
```diff
@@ -0,0 +1,64 @@
...
+
+class TestBotasShippedAdvisory(helpers.ModelsTestCase):
+
+    def setUp(self):
+        super(TestBotasShippedAdvisory, self).setUp()
+
+        # Each time when recording a build into database, freshmaker has to
+        # request a pulp repo from ODCS. This is not necessary for running
+        # tests.
+        self.patcher = helpers.Patcher(
```
mprahl, line 39: Is this patch needed?

tests/handlers/botas/test_botas_shipped_advisory.py
```diff
@@ -0,0 +1,64 @@
...
+            'freshmaker.handlers.botas.botas_shipped_advisory.')
+
+        # We do not want to send messages to message bus while running tests
+        self.mock_messaging_publish = self.patcher.patch(
+            'freshmaker.messaging.publish')
+
+        self.botas_advisory = ErrataAdvisory(
+            123, "RHBA-2020", "SHIPPED_LIVE", ['docker'])
+        self.botas_advisory._reporter = "botas/pnt-devops-jenkins@REDHAT.COM"
+        self.rhba_event = ErrataAdvisoryRPMsSignedEvent(
```
mprahl, line 49: This is not needed

mprahl (approved):

tests/handlers/botas/test_botas_shipped_advisory.py
apaplaus: Yes, I used it in deleted test, and thought that it could be useful for future tests.

PR #34, opened 2020-10-14
Query images from Lightblue only with RH vendor

apaplaus:
Add 'vendor' parameter to Lightblue queries, so now we will get container images only with Red Hat vendor.

Resolves: CLOUDWF-3527

### DIFF
--- a/freshmaker/lightblue.py
+++ b/freshmaker/lightblue.py
@@ -631,7 +631,8 @@ class LightBlue(object):
 
     def _set_container_repository_filters(
             self, request, published=True,
-            release_categories=conf.lightblue_release_categories):
+            release_categories=conf.lightblue_release_categories,
+            vendors=conf.lightblue_repo_vendors):
         """
         Sets the additional filters to containerRepository request
         based on the self.published, self.release_categories attributes.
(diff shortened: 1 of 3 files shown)


freshmaker/lightblue.py
```diff
@@ -906,6 +906,11 @@ def find_images_with_included_rpms(
                             "rvalue": tag
                         } for tag in auto_rebuild_tags]
                     },
+                    {
+                        "field": "parsed_data.labels.*.vendor",
+                        "op": "=",
+                        "rvalue": "Red Hat, Inc."
```
mprahl, line 912: Do you think this should be configurable? This would allow Freshmaker to be used for other vendors in the future (e.g. partners).

freshmaker/lightblue.py
```diff
@@ -631,7 +631,8 @@ def find_container_images(self, request):
 
     def _set_container_repository_filters(
             self, request, published=True,
-            release_categories=conf.lightblue_release_categories):
+            release_categories=conf.lightblue_release_categories,
+            vendors=conf.lightblue_repo_vendors):
```
mprahl, line 635: Could you add `vendors` to the docstring?

mprahl:
@apaplaus were you able to test the lightblue query?

apaplaus:
Yes, I've tested it using the script I wrote. And it was working, it queried repos with redhat label as expected.

freshmaker/lightblue.py
apaplaus: sure

gnaponie (approved):
Looks good to me. I've also tested it locally and it works.
@qixiang not sure if you want to review this too since you've filed the card.

apaplaus:
Ops, I saw your approve and merged it =(

gnaponie:
:D That's ok. Qixiang is allowed to be mad at you now, but I'm sure he won't :)

qixiang:
No worry, it looks good to me :)

PR #37, opened 2020-10-29
Consider latest released images for "dependency" images

mprahl:
When Freshmaker has to rebuild parent images that are not directly affected, it now detects if it can use the latest released image at the pull specification declared by the child image. If it can, this parent image is used instead of rebuilding the original parent image.

Resolves CLOUDWF-541

### DIFF
--- a/freshmaker/lightblue.py
+++ b/freshmaker/lightblue.py
@@ -840,7 +840,7 @@ class LightBlue(object):
                     break
             else:
                 log.info(
-                    "Will not rebuild %s because there is a modularity mismatch between the RPMs "
+                    "Filtered out %s because there is a modularity mismatch between the RPMs "
                     "from the image and the advisory: %r" % (
                         image.nvr, rpm_name_to_nvrs.values()))
         return ret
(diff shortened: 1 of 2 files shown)


freshmaker/lightblue.py
```diff
@@ -1579,5 +1579,239 @@ def _get_images_to_rebuild(image):
...
+            if not_directly_affected_index is None:
+                continue
+
+            # Try replacing all the not directly affected images starting from the first one
+            for i in range(not_directly_affected_index, len(image_group)):
+                # On the first iteration, this is the last directly affected image in image_group
+                child_image = image_group[i - 1]
+                # Find the pullspec of the parent from the perspective of the child image
+                parent_pull_spec = None
+                for pull_spec, pull_spec_info in (child_image["parent_image_builds"] or {}).items():
```
qixiang, line 1632:
```suggestion
                for pull_spec, pull_spec_info in child_image.get("parent_image_builds", {}).items():
```

freshmaker/lightblue.py
```diff
@@ -1579,5 +1579,239 @@ def _get_images_to_rebuild(image):
...
+                if not parent_image.get_rpms():
+                    log.warning("The parent image %s does not have an RPM manifest", parent_image.nvr)
+                    continue
+
+                for rpm in parent_image.get_rpms():
+                    if rpm_name_to_nvrs.get(rpm["name"]):
+                        parent_applicable_rpm_nvrs.add(rpm_name_to_nvrs[rpm["name"]])
+
+                fixed_published_image = self.get_fixed_published_image(
+                    parent_pull_spec, parent_applicable_rpm_nvrs, content_sets
```
qixiang, line 1654: I'm wondering if it's a good idea to get latest published image by the `pull_spec`, sometimes the parent image can be published later and has different pull_spec with the data in brew (that was generated while building). For example, a parent image (no matter rebuilt due to dependency or affected directly) rebuilt in a freshmaker event always has internal osbs pull specs without registry.

mprahl: Good point. I suppose I can get the original `FROM` line from dist-git but that's hacky.

The only other option I thought of was searching for the latest published name and version that matches, however, it has a few drawbacks. For some images, they just build against the latest RHEL/UBI of a particular major version (e.g. rhel/rhel7:lastest). Upgrading to a higher release with the same name and version as the previously used parent image may not be correct. Additionally, it's possible that the image is built against a tag with a specific NVR (though not super likely), so upgrading this would not be correct. I don't think the current deduplication code takes this into account. Lastly, the highest NVR is not guaranteed to be the latest image for a tag.

What are your thoughts?

qixiang: > The only other option I thought of was searching for the latest published name and version that matches, however, it has a few drawbacks.

That's what I was thinking of and as you said, it's not guaranteed to be safe to just use the highest release image of the name-version. 

An interesting thing is some people want us to use latest published parent image instead of the original parent image, because the original parent can have some outdated configs (e.g. contest sets, CVP testing scripts), those layered images are not updated so frequently, to make Freshmaker rebuilds their images with the latest published parent, they need to rebuild a new image with the latest parent by manual and publish it. Currently we don't have a good solution to support this, because images can have different workflow or rules for building and updating there NVRs.

freshmaker/lightblue.py
```diff
@@ -1579,5 +1579,239 @@ def _get_images_to_rebuild(image):
...
+            )
+            return
+
+        registries = [registry]
+        if conf.lightblue_alt_registry_names.get(registry):
+            registries.append(conf.lightblue_alt_registry_names[registry])
+
+        # It is too slow to also filter by the expected RPMs. This is done outside of the lightblue
+        # query instead.
+        request = {
```
mprahl, line 1744: @mprahl to verify that this is supported in Pyxis

freshmaker/lightblue.py
mprahl: It's a nice suggestion, however, since this key gets assigned with `data["parent_image_builds"] = extra_image.get("parent_image_builds")`, I can't rely on that.

freshmaker/config.py
```diff
@@ -401,6 +401,14 @@ class Config(object):
             'type': str,
             'default': 'all',
             'desc': 'vcr mode for recording lightblue queries'
+        },
+        'lightblue_alt_registry_names': {
```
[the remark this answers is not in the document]
mprahl: In lightblue, the registries are named registry.access.redhat.com but most pullspecs in the `FROM` line of child images are registry.redhat.io.

freshmaker/lightblue.py
```diff
@@ -1579,5 +1579,239 @@ def _get_images_to_rebuild(image):
...
+        """
+        Replace images in ``to_rebuild`` that are not directly affected and have published fixes.
+
+        When an image and its parents in ``to_rebuild`` are not directly affected, it's possible
+        that the image had been rebuilt outside of Freshmaker and published with the fix applied.
+        In this case, Freshmaker should not rebuild the image and its parents. The latest published
+        image is found using the pull spec of the parent image from the perspective of
+        the last directly affected image (FROM line).
+
+        :param Iterable to_rebuild: the list of images to rebuild
```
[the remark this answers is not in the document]
mprahl: Yes, how would you document that in the docstring?

freshmaker/lightblue.py
```diff
@@ -1579,5 +1579,239 @@ def _get_images_to_rebuild(image):
...
+        :param Iterable directly_affected_nvrs: the set of image NVRs in ``to_rebuild`` that are
+            marked as directly affected
+        :param Iterable rpm_nvrs: the list of RPM NVRs with the fixes in the advisory
+        :param Iterable content_sets: the list of content sets that the RPMs in ``rpm_nvrs`` are
+            released in
+        """
+        for image_group in to_rebuild:
+            # Find the first index in image_group of an image that is not directly
```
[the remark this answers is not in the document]
mprahl: I'll add this to the description in the docstring.

freshmaker/lightblue.py
```diff
@@ -1579,5 +1579,239 @@ def _get_images_to_rebuild(image):
...
+            # affected with parents that are also not directly affected
+            not_directly_affected_index = None
+            # Skip the first image in the group since it is always directly affected
+            for i, image in enumerate(image_group[1:]):
+                if image.nvr in directly_affected_nvrs:
+                    not_directly_affected_index = None
+                elif not_directly_affected_index is None:
+                    # We must add 1 since the first index was skipped in the for loop
+                    not_directly_affected_index = i + 1
```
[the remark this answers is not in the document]
mprahl: That's a nice suggestion. Thanks!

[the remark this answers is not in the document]
mprahl: I think it would be possible since the base image could have the RPM installed, the intermediate image uninstalls it, and the child image reinstalls the same RPM. It's very unlikley but it's not very wasteful to guard against this.

What do you think?

freshmaker/lightblue.py
```diff
@@ -1579,5 +1579,239 @@ def _get_images_to_rebuild(image):
...
+
+            # Try replacing all the not directly affected images starting from the first one
+            for i in range(not_directly_affected_index, len(image_group)):
+                # On the first iteration, this is the last directly affected image in image_group
+                child_image = image_group[i - 1]
+                # Find the pullspec of the parent from the perspective of the child image
+                parent_pull_spec = None
+                for pull_spec, pull_spec_info in (child_image["parent_image_builds"] or {}).items():
+                    if pull_spec_info["id"] == child_image["parent_build_id"]:
+                        parent_pull_spec = pull_spec
```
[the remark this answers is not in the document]
mprahl: As discussed, I'll refactor this to find the latest published image of the same name and version instead. It's less accurate but it's safer since an image maintainer may be aggressive at build time but be more conservative when autorebuilds occur.

mprahl:
@lcarva @qixiang @gnaponie @apaplaus I refactored the PR to only consider the latest release of the same name and version after discussing this with Luiz. He had pointed out that using the pullspec of the parent image from the perspective of the child image is likely too aggressive. This is because maintainers may be more conservative when doing auto-rebuilds since there is more testing during a normal product release.

freshmaker/lightblue.py
```diff
@@ -1579,5 +1579,249 @@ def _get_images_to_rebuild(image):
...
+                    candidate_image_group,
+                    image,
+                )
+                continue
+
+            # Due to filtering by installed RPMs taking too long in lightblue, perform the filter
+            # here since the projection (returned RPM manifest from lightblue) has the filtering
+            # applied. This is to be conservative in the event a child image relies on the RPM but
+            # it is no longer installed
+            if {rpm["name"] for rpm in image.get_rpms() or []} != rpm_name_to_nvrs.keys():
```
qixiang, line 1772: An image doesn't need to have all the rpms from the cve fix rpm advisory installed. We can remove this check.

qixiang:
One minor comment, beside of that, it looks good to me.

gnaponie (approved):

freshmaker/lightblue.py
mprahl: @qixiang `rpm_name_to_nvrs` just contains the RPMs that were present in the vulnerable parent that were fixed by the advisory. So this check verifies that those same RPMs are still present in the fixed version of the parent image. This is to ensure that in the event the child image relies on those RPMs already being installed by the parent image, that they continue to be installed.

I admit that it's very cautious, but this is what @gnaponie and I had agreed to when I discussed this with her a few weeks ago.

@gnaponie do you still agree? I can remove this check if you'd like.

qixiang: I see, I misread it, I thought the `rpm_name_to_nvrs` are rpms from the rpm advisory. I agree it's ok to keep this check.

qixiang (approved):

lcarva (commented):
Nice!

PR #39, opened 2020-11-12
Auto rebuilt filter

apaplaus:
New Pyxis module to query bundle images and ContainerImages by thier
nvrs. So we can find bundle images for the operator/operand images that
were rebuilt by us previously.

Resolve CLOUDWF-3955

### DIFF
--- a/freshmaker/handlers/botas/botas_shipped_advisory.py
+++ b/freshmaker/handlers/botas/botas_shipped_advisory.py
@@ -81,7 +81,6 @@ class HandleBotasAdvisory(ContainerBuildHandler):
             msg = f"The are no digests for NVRs: {','.join(original_nvrs)}"
             log.warning(msg)
             db_event.transition(EventState.SKIPPED, msg)
-            db.session.commit()
             return []
 
         index_images = self._pyxis.get_operator_indices()
@@ -100,9 +99,13 @@ class HandleBotasAdvisory(ContainerBuildHandler):
             bundle_digests.add(bundle['bundle_path_digest'])
         bundle_images = self._pyxis.get_images_by_digests(bundle_digests)
 
+        # Filter image nvrs that don't have or never had auto_rebuild tag
+        # in repos, where image is published
+        auto_rebuild_nvrs = self._pyxis.get_auto_rebuild_tagged_images(bundle_images)
+
         # get NVRs only of those bundles, which have OSBS pinning
         bundles_nvrs = self._filter_bundles_by_pinned_related_images(
-            bundle_images)
+            auto_rebuild_nvrs)
 
         # Skip that event because we can't proceed with processing it.
         # TODO
@@ -112,24 +115,23 @@ class HandleBotasAdvisory(ContainerBuildHandler):
         db_event.transition(EventState.SKIPPED, msg)
         return []
 
-    def _filter_bundles_by_pinned_related_images(self, bundle_images):
+    def _filter_bundles_by_pinned_related_images(self, bundle_image_nvrs):
         """
-        If the digests were not pinned by OSBS, the bundle image
+        If the digests were not pinned by OSBS, the bundle image nvr
         will be filtered out.
 
         There is no need in checking pinning for every of related images,
         because we already know that digest points to the manifest list,
         because of previous filtering.
 
-        :param list bundle_images: ContainerImages of operator bundles
+        :param set bundle_image_nvrs: NVRs of operator bundles
         :return: list of NVRs of bundle images that have at least one
             original related image that was rebuilt
         """
         ret_bundle_images_nvrs = set()
         with koji_service(conf.koji_profile, log, dry_run=self.dry_run,
                           login=False) as session:
-            for bundle in bundle_images:
-                nvr = bundle['brew']['build']
+            for nvr in bundle_image_nvrs:
                 build = session.get_build(nvr)
                 if not build:
                     log.error("Could not find the build %s in Koji", nvr)
--- a/freshmaker/pyxis.py
+++ b/freshmaker/pyxis.py
@@ -215,6 +215,121 @@ class Pyxis(object):
         :return: bundle ContainerImages
         :rtype: list
         """
-        request_params = {'include': 'data.brew',
-                          'filter': f"repositories.manifest_list_digest=in=({','.join(digests)})"}
-        return self._pagination("images", request_params)
+        request_params = {'include': 'data.brew,data.repositories',
+                          'filter': f'repositories.manifest_list_digest=in=({",".join(digests)})'}
+        return self._pagination('images', request_params)
+
+    def _add_repositories_info(self, reg_repo_info):
+        """
+        For every pair of registry-repository add information about it's
+        auto_rebuild tags. To decrease amount of queries to Pyxis, only one
+        query is performed with the filter set to proper registry-repository
+        pair.
+
+        A list of tags for each repository will be added to the
+        'auto_rebuild_tags' key in the input info about the repository.
+
+        Entries about repos without 'auto_rebuild_tags' will be deleted from
+        the mapping.
+
+        :param dict reg_repo_info: map of pairs (registry, repository) to a
+            dict containing nvrs of bundle images from that repo and
+            auto_rebuild tags of that repo
+        """
+        if not reg_repo_info:
+            return None
+        fltr = ""
+        # Construct filter for future request to Pyxis with registry-repository pairs
+        for reg, repo in reg_repo_info.keys():
+            if fltr:
+                fltr += ','
+            fltr += f'(registry=={reg};repository=={repo})'
+        params = {'include': ','.join(['data.auto_rebuild_tags',
+                                       'data.registry', 'data.repository']),
+                  'filter': fltr}
+        repos = self._pagination('repositories', params)
+
+        # For every repo add it's auto_rebuild_tags info
+        for repo in repos:
+            reg_repo_pair = (repo['registry'], repo['repository'])
+            # one of repos isn't in previously constructed map,
+            # so there is inconsistency
+            if reg_repo_pair not in reg_repo_info:
+                log.warning('There is inconsistency in naming for: %s/%s',
+                            reg_repo_pair[0], reg_repo_pair[1])
+                continue
+            tags = repo.get('auto_rebuild_tags')
+            # If the repository doesn't have 'auto_rebuild_tags', don't proceed with it
+            if tags:
+                reg_repo_info[reg_repo_pair]['auto_rebuild_tags'] = set(tags)
+            else:
+                del reg_repo_info[reg_repo_pair]
+
+    def _filter_auto_rebuild_nvrs(self, reg_repo_info):
+        """
+        For every pair in mapping get history of every auto_rebuild tag
+        filtering all nvrs from that registry/repo pair.
+
+        :param dict reg_repo_info: map of pairs (registry, repository) to a
+            dict containing nvrs of bundle images from that repo and
+            auto-rebuild tags of that repo
+        :return: NVRs that were at least once tagged with an auto-rebuild tag
+        :rtype set(str):
+        """
+        params = {'include': 'data.brew.build'}
+        ret_nvrs = set()
+        for reg_repo_pair, info in reg_repo_info.items():
+            if not info.get('nvrs'):
+                continue
+            fltr = f'brew.build=in=({",".join(info["nvrs"])})'
+            params['filter'] = fltr
+            reg, repo = reg_repo_pair
+            for tag in info.get('auto_rebuild_tags', []):
+                tag_history = \
+                    self._pagination(f'repositories/registry/{reg}/'
+                                     f'repository/{repo}/tag/{tag}', params)
+                # If there is at least one record with nvr(brew.build) of the
+                # image, it means that image was published to a repo with
+                # auto_rebuild tag
+                if tag_history:
+                    for temp_tag in tag_history:
+                        ret_nvrs.add(temp_tag['brew']['build'])
+        return ret_nvrs
+
+    def get_auto_rebuild_tagged_images(self, bundle_images):
+        """
+        Determine which bundle images are published to a container repository
+        and was at least once tagged with an auto-rebuild tag.
+
+        :param list bundle_images: Images of operator bundles that should
+            be filtered
+        :return: Image NVRs that where published to repositories with auto_rebuild tag
+        :rtype: set(str)
+        """
+        reg_repo_info = {}
+        for bundle_image in bundle_images:
+            bundle_nvr = bundle_image.get('brew', {}).get('build')
+            if not bundle_nvr:
+                log.warning('One of bundle images doesn\'t have brew.build')
+                continue
+            if not bundle_image.get('repositories'):
+                log.warning('Bundle image %s doesn\'t have repositories set',
+                            bundle_nvr)
+                continue
+            # construct mapping of (registry, repository) -> {'nvrs': {bundles_nvrs}}
+            for repo in bundle_image.get('repositories'):
+                if not (repo.get('registry') and repo.get('repository')):
+                    log.warning('"registry" or "repository" isn\'t set in %s',
+                                bundle_nvr)
+                    continue
+                reg_repo = (repo['registry'], repo['repository'])
+                reg_repo_info.setdefault(reg_repo, {})\
+                    .setdefault('nvrs', set()).add(bundle_nvr)
+
+        # Add auto_rebuild_tags to info structures for every repo
+        self._add_repositories_info(reg_repo_info)
+        # Get tag history for every repo and get nvrs tagged with auto_rebuild tag
+        nvrs = self._filter_auto_rebuild_nvrs(reg_repo_info)
+        if not nvrs:
+            log.warning('Can\'t find any nvr tagged with an auto-rebuild tag')
+        return nvrs
--- a/tests/test_pyxis.py
+++ b/tests/test_pyxis.py
@@ -219,11 +219,17 @@ class TestQueryPyxis(helpers.FreshmakerTestCase):
                 "repositories": [
                     {
                         "manifest_list_digest": "sha256:1111",
-                        "published": False
+                        "published": False,
+                        "registry": "reg1",
+                        "repository": "repo1",
+                        "tags": [{"name": "tag0"}]
                     },
                     {
                         "manifest_list_digest": "sha256:1112",
-                        "published": True
+                        "published": True,
+                        "registry": "reg2",
+                        "repository": "repo2",
+                        "tags": [{"name": "tag1"}, {"name": "tag2"}]
                     }
                 ]
             },
@@ -237,7 +243,10 @@ class TestQueryPyxis(helpers.FreshmakerTestCase):
                 "repositories": [
                     {
                         "manifest_list_digest": "sha256:2222",
-                        "published": True
+                        "published": True,
+                        "registry": "reg2",
+                        "repository": "repo2",
+                        "tags": [{"name": "tag2"}]
                     }
                 ]
             },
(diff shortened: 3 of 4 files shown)


apaplaus:
@mprahl please take a look also on that PR which is about CLOUDWF-3955. It doesn't have any tests yet. Just check if direction where I am going is fine. And It would be better to close previous PR before that, because it's complex to base new work on something that isn't finished yet.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -100,6 +99,10 @@ def handle(self, event):
             bundle_digests.add(bundle['bundle_path_digest'])
         bundle_images = self._pyxis.get_images_by_digests(bundle_digests)
 
+        # Filter images that don't have or never had aut_rebuild tag in repos,
```
mprahl, line 102:
```suggestion
        # Filter images that don't have or never had auto_rebuild tag in repos,
```

freshmaker/pyxis.py
```diff
@@ -215,6 +215,129 @@ def get_images_by_digests(self, digests):
...
+            be filtered
+        :return: Images that where published to repository with auto_rebuild tag
+        :rtype: list
+        """
+        reg_repo_info = {}
+        ret_images = []
+        for bundle_image in bundle_images:
+            bundle_nvr = bundle_image.get('brew', {}).get('build')
+            if not bundle_nvr:
+                log.warning("One of bundle images doesn't have brew.build")
```
mprahl, line 311: Optional: It'd be nice to be consistent within the method of using single quotes vs double quotes. It's bit of a pain when modifying code when you keep needing to switch back and forth.

freshmaker/pyxis.py
```diff
@@ -215,6 +215,129 @@ def get_images_by_digests(self, digests):
...
+                continue
+            # construct mapping of (registry, repository) -> {'nvrs': {bundles_nvrs}}
+            for repo in bundle_image.get('repositories'):
+                reg_repo = (repo.get('registry'), repo.get('repository'))
+                if reg_repo not in reg_repo_info:
+                    reg_repo_info[reg_repo] = {'nvrs': {bundle_nvr}}
+                elif 'nvrs' not in reg_repo_info[reg_repo]:
+                    reg_repo_info[reg_repo]['nvrs'] = {bundle_nvr}
+                else:
+                    reg_repo_info[reg_repo]['nvrs'].add(bundle_nvr)
```
mprahl, line 325:
```suggestion
                reg_repo_info.setdefault(reg_repo, {'nvrs': set()}).add(bundle_nvr)
```

freshmaker/pyxis.py
```diff
@@ -215,6 +215,129 @@ def get_images_by_digests(self, digests):
...
+            if not bundle_nvr:
+                log.warning("One of bundle images doesn't have brew.build")
+                continue
+            if not bundle_image.get('repositories'):
+                log.warning("Bundle image %s doesn't have repositories set",
+                            bundle_nvr)
+                continue
+            # construct mapping of (registry, repository) -> {'nvrs': {bundles_nvrs}}
+            for repo in bundle_image.get('repositories'):
+                reg_repo = (repo.get('registry'), repo.get('repository'))
```
mprahl, line 319: Can `registry` or `repository` not be set? If so, that seems like something we should ignore. If it's always set, then you should just access the keys directly to insinuate that it's always set and no handling is required for when it is not set.

freshmaker/pyxis.py
```diff
@@ -215,6 +215,129 @@ def get_images_by_digests(self, digests):
...
+            else:
+                fltr += f",(registry=={reg};repository=={repo})"
+        params = {'include': ','.join(['data.auto_rebuild_tags',
+                                       'data.registry', 'data.repository']),
+                  'filter': fltr}
+        repos = self._pagination('repositories', params)
+
+        # For every repo add it's auto_rebuild_tags info
+        for repo in repos:
+            reg_repo_pair = (repo.get('registry'), repo.get('repository'))
```
mprahl, line 251: Same comment as https://github.com/redhat-exd-rebuilds/freshmaker/pull/39/files#r528941742

freshmaker/pyxis.py
```diff
@@ -215,6 +215,129 @@ def get_images_by_digests(self, digests):
...
+                else:
+                    reg_repo_info[reg_repo]['nvrs'].add(bundle_nvr)
+
+        # Add auto_rebuild_tags to info structures for every repo
+        self._add_repositories_info(reg_repo_info)
+        # Get tag history for every repo and filter out repos that didn't have
+        # bundle images in their history
+        repos = self._filter_auto_rebuild_repos(reg_repo_info)
+        if not repos:
+            log.warning("Can't find any repo with aut_rebuild tags")
```
mprahl, line 333:
```suggestion
            log.warning("Can't find any repo with auto_rebuild tags")
```

freshmaker/pyxis.py
```diff
@@ -215,6 +215,129 @@ def get_images_by_digests(self, digests):
...
+                elif 'nvrs' not in reg_repo_info[reg_repo]:
+                    reg_repo_info[reg_repo]['nvrs'] = {bundle_nvr}
+                else:
+                    reg_repo_info[reg_repo]['nvrs'].add(bundle_nvr)
+
+        # Add auto_rebuild_tags to info structures for every repo
+        self._add_repositories_info(reg_repo_info)
+        # Get tag history for every repo and filter out repos that didn't have
+        # bundle images in their history
+        repos = self._filter_auto_rebuild_repos(reg_repo_info)
```
mprahl, line 331: If I understand the code correctly, this is returning all the repos which have at least one vulnerable bundle image that was at least once tagged with an auto rebuild tag. Later on, the code determines which bundle images that are published to these repos to then rebuild. This is not sufficient since it's possible one of the vulnerable bundle images was tagged to "registry/ns/repo:v4.5" but auto rebuilds are only set to the `v4.6` tag on that repo.

apaplaus: @mprahl  In that case will it be enough to check in addition if at least one image tag(that **repositories->tags->name**) is in **auto_rebuild_tags** list of the repo? I will push changes so you can take a look.

mprahl: @apaplaus I think what you'll need to do is map the NVRs of the bundle images to the repositories like you are doing. Then check the tag history of each autorebuild tag to see of any of the NVRs on that repository were or are currently tagged with one of the autorebuild tags.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -81,7 +81,6 @@ def handle(self, event):
             msg = f"The are no digests for NVRs: {','.join(original_nvrs)}"
             log.warning(msg)
             db_event.transition(EventState.SKIPPED, msg)
-            db.session.commit()
```
mprahl, line 84: I think this is still needed unless the `transition` method calls this.

freshmaker/pyxis.py
```diff
@@ -215,6 +215,135 @@ def get_images_by_digests(self, digests):
...
+            be filtered
+        :return: Images that where published to repository with auto_rebuild tag
+        :rtype: list
+        """
+        reg_repo_info = {}
+        ret_images = []
+        for bundle_image in bundle_images:
+            bundle_nvr = bundle_image.get('brew', {}).get('build')
+            if not bundle_nvr:
+                log.warning("One of bundle images doesn't have brew.build")
```
mprahl, line 312: Is there an identifier that you can use in the log message to denote which bundle image doesn't have an NVR?

freshmaker/pyxis.py
mprahl: I reread your comment and I think I originally misunderstood. Let me check the code to see.

freshmaker/pyxis.py
```diff
@@ -215,6 +215,135 @@ def get_images_by_digests(self, digests):
...
+        return self._pagination('images', request_params)
+
+    def _add_repositories_info(self, reg_repo_info):
+        """
+        For every pair of registry-repository add information about it's
+        auto_rebuild tags. To decrease amount of queries to Pyxis, only one
+        query is performed with the filter set to proper registry-repository
+        pair.
+
+        Tags list is added to the parameter object of the method.
```
mprahl, line 229: Could you please reword this to be clearer? How about?
```suggestion
        A list of tags for each repository will be added to the 'tags' key in the input info about the repository.
```

freshmaker/pyxis.py
```diff
@@ -215,6 +215,135 @@ def get_images_by_digests(self, digests):
...
+            # one of repos isn't in previously constructed map,
+            # so there is inconsistency
+            if reg_repo_pair not in reg_repo_info:
+                log.warning("There is inconsistency in naming for: %s/%s",
+                            reg_repo_pair[0], reg_repo_pair[1])
+                continue
+            tags = repo.get('auto_rebuild_tags')
+            # If the repository doesn't have 'auto_rebuild_tags', don't proceed with it
+            if tags:
+                reg_repo_info[reg_repo_pair]['tags'] = set(tags)
```
mprahl, line 261: Could you please use the key `auto_rebuild_tags` to make this clearer when processing this data outside of this method?

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: That was actually the reason why I deleted it, `transition` calls it before publishing messages and returning.

freshmaker/pyxis.py
apaplaus: For now I am querying Pyxis only for two fields and they are **data.brew** and **data.repositories**. So this check is more for preventing someone calling it with incorrect data structures. I could use **brew.nvra** to print that warning, but it doesn't have much sense since it's almost the same data that is missing.

freshmaker/pyxis.py
```diff
@@ -215,6 +215,135 @@ def get_images_by_digests(self, digests):
...
+
+        :param dict reg_repo_info: map of pairs (registry, repository) to a
+            dict containing nvrs of bundle images from that repo and
+            auto_rebuild tags of that repo
+        :return: pairs of (registry, repository) of every repo with images ever
+            published to a repo with auto_rebuild tag + info about it
+        :rtype dict(tuple(str, str): dict()):
+        """
+        params = {'include': 'data.brew.build'}
+        ret_repos = dict()
```
mprahl, line 278:
```suggestion
        ret_repos = {}
```

freshmaker/pyxis.py
mprahl: @apaplaus I think you could modify `_filter_auto_rebuild_repos` to create and return a set of all the NVRs that were once tagged with an autorebuild tag. I think in the section that you do `if tag_history:`, you could just loop through the NVRs returned and add them to the set since we know that the NVRs are limited to just those that are bundle images of interest. Then the additional for loop of `for bundle_image in bundle_images:` in this method could be removed and it could just return the set of NVRs.

freshmaker/pyxis.py
```diff
@@ -215,6 +215,121 @@ def get_images_by_digests(self, digests):
...
+        return self._pagination('images', request_params)
+
+    def _add_repositories_info(self, reg_repo_info):
+        """
+        For every pair of registry-repository add information about it's
+        auto_rebuild tags. To decrease amount of queries to Pyxis, only one
+        query is performed with the filter set to proper registry-repository
+        pair.
+
+        A list of tags for each repository will be added to the 'tags' key in
```
mprahl, line 229:
```suggestion
        A list of tags for each repository will be added to the 'auto_rebuild_tags' key in
```

freshmaker/pyxis.py
```diff
@@ -215,6 +215,121 @@ def get_images_by_digests(self, digests):
...
+                tag_history = \
+                    self._pagination(f"repositories/registry/{reg}/"
+                                     f"repository/{repo}/tag/{tag}", params)
+                # If there is at least one record with nvr(brew.build) of the
+                # image, it means that image was published to a repo with
+                # auto_rebuild tag
+                if tag_history:
+                    auto_rebuild_nvrs = \
+                        [temp_tag['brew']['build'] for temp_tag in tag_history]
+                    ret_nvrs.update(auto_rebuild_nvrs)
```
mprahl, line 295: This is easier to read, more memory efficient, and should be slightly faster. For future reference, you can define a set with the list comprehension syntax by using curly braces (e.g. `{temp_tag['brew']['build'] for temp_tag in tag_history}`).
```suggestion
                    for temp_tag in tag_history:
                        ret_nvrs.add(temp_tag['brew']['build'])
```

freshmaker/pyxis.py
```diff
@@ -215,6 +215,121 @@ def get_images_by_digests(self, digests):
...
+        return ret_nvrs
+
+    def get_auto_rebuild_taged_images(self, bundle_images):
```
mprahl, line 298:
```suggestion
    def get_auto_rebuild_tagged_images(self, bundle_images):
```

freshmaker/pyxis.py
```diff
@@ -215,6 +215,121 @@ def get_images_by_digests(self, digests):
...
+                reg_repo_info.setdefault(reg_repo, {})\
+                    .setdefault('nvrs', set()).add(bundle_nvr)
+
+        # Add auto_rebuild_tags to info structures for every repo
+        self._add_repositories_info(reg_repo_info)
+        # Get tag history for every repo and get nvrs tagged with auto_rebuild tag
+        nvrs = self._filter_auto_rebuild_nvrs(reg_repo_info)
+        if not nvrs:
+            log.warning("Can't find any nvr tagged with an auto_rebuild tag")
+            return set()
```
mprahl, line 334: `nvrs` would be an empty set in this case, so this line is not necessary.
```suggestion
```

freshmaker/pyxis.py
```diff
@@ -215,6 +215,121 @@ def get_images_by_digests(self, digests):
...
+                if tag_history:
+                    auto_rebuild_nvrs = \
+                        [temp_tag['brew']['build'] for temp_tag in tag_history]
+                    ret_nvrs.update(auto_rebuild_nvrs)
+        return ret_nvrs
+
+    def get_auto_rebuild_taged_images(self, bundle_images):
+        """
+        Determine which bundle images are part of container repository and tag
+        that has auto-rebuilds enabled
```
mprahl, line 301:
```suggestion
        Determine which bundle images are published to a container repository and was at least once tagged with an auto-rebuild tag.
```

freshmaker/pyxis.py
```diff
@@ -215,6 +215,121 @@ def get_images_by_digests(self, digests):
...
+            if not bundle_nvr:
+                log.warning("One of bundle images doesn't have brew.build")
+                continue
+            if not bundle_image.get('repositories'):
+                log.warning("Bundle image %s doesn't have repositories set",
+                            bundle_nvr)
+                continue
+            # construct mapping of (registry, repository) -> {'nvrs': {bundles_nvrs}}
+            for repo in bundle_image.get('repositories'):
+                if 'registry' not in repo or 'repository' not in repo:
```
mprahl, line 320: This is safer to guard against falsy (e.g. None) values.
```suggestion
                if not (repo.get('registry') and repo.get('repository')):
```

freshmaker/pyxis.py
```diff
@@ -215,6 +215,121 @@ def get_images_by_digests(self, digests):
...
+        """
+        if not reg_repo_info:
+            return None
+        fltr = ""
+        # Construct filter for future request to Pyxis with registry-repository pairs
+        for reg, repo in reg_repo_info.keys():
+            if not fltr:
+                fltr = f"(registry=={reg};repository=={repo})"
+            else:
+                fltr += f",(registry=={reg};repository=={repo})"
```
mprahl, line 244: This reduces code duplication.
```suggestion
            if fltr:
                fltr += ","
            fltr = f"(registry=={reg};repository=={repo})"

```

freshmaker/pyxis.py
```diff
@@ -215,6 +215,121 @@ def get_images_by_digests(self, digests):
...
+        params = {'include': ','.join(['data.auto_rebuild_tags',
```
mprahl, line 245: Can you be consistent with your use of single quotes vs double quotes within the same method please?

freshmaker/pyxis.py
```diff
@@ -215,6 +215,121 @@ def get_images_by_digests(self, digests):
...
+            if reg_repo_pair not in reg_repo_info:
+                log.warning("There is inconsistency in naming for: %s/%s",
+                            reg_repo_pair[0], reg_repo_pair[1])
+                continue
+            tags = repo.get('auto_rebuild_tags')
+            # If the repository doesn't have 'auto_rebuild_tags', don't proceed with it
+            if tags:
+                reg_repo_info[reg_repo_pair]['auto_rebuild_tags'] = set(tags)
+            else:
+                del reg_repo_info[reg_repo_pair]
```
mprahl, line 264: Could you add document this behavior in a separate section in the docstring please? I was not expecting this to occur when first reviewing this method.

freshmaker/pyxis.py
```diff
@@ -215,6 +215,121 @@ def get_images_by_digests(self, digests):
...
+
+    def _filter_auto_rebuild_nvrs(self, reg_repo_info):
+        """
+        For every pair in mapping get history of every auto_rebuild tag
+        filtering all nvrs from that registry/repo pair.
+
+        :param dict reg_repo_info: map of pairs (registry, repository) to a
+            dict containing nvrs of bundle images from that repo and
+            auto_rebuild tags of that repo
```
mprahl, line 273:
```suggestion
            auto-rebuild tags of that repo
```

freshmaker/pyxis.py
```diff
@@ -215,6 +215,121 @@ def get_images_by_digests(self, digests):
...
+        :return: NVRs that were once tagged with an autorebuild tag
```
mprahl, line 274:
```suggestion
        :return: NVRs that were once tagged with an auto-rebuild tag
```

freshmaker/pyxis.py
```diff
@@ -215,6 +215,121 @@ def get_images_by_digests(self, digests):
...
+                reg_repo = (repo['registry'], repo['repository'])
+                reg_repo_info.setdefault(reg_repo, {})\
+                    .setdefault('nvrs', set()).add(bundle_nvr)
+
+        # Add auto_rebuild_tags to info structures for every repo
+        self._add_repositories_info(reg_repo_info)
+        # Get tag history for every repo and get nvrs tagged with auto_rebuild tag
+        nvrs = self._filter_auto_rebuild_nvrs(reg_repo_info)
+        if not nvrs:
+            log.warning("Can't find any nvr tagged with an auto_rebuild tag")
```
mprahl, line 333:
```suggestion
            log.warning("Can't find any nvr tagged with an auto-rebuild tag")
```

tests/test_pyxis.py
```diff
@@ -424,10 +439,125 @@ def test_get_images_by_digests(self, page):
...
+                 {"include": "data.brew.build",
+                  "filter": f"brew.build=in=({','.join({'nvr4','nvr5'})})"}),
+            call("repositories/registry/reg2/repository/rep1/tag/tag2",
+                 {"include": "data.brew.build",
+                  "filter": f"brew.build=in=({','.join({'nvr4','nvr5'})})"}),
+        ])
+
+    @patch('freshmaker.pyxis.Pyxis._filter_auto_rebuild_nvrs')
+    @patch('freshmaker.pyxis.Pyxis._add_repositories_info')
+    def test_get_auto_rebuild_taged_images(self, info, tag_filter):
```
mprahl, line 536:
```suggestion
    def test_get_auto_rebuild_tagged_images(self, info, tag_filter):
```

apaplaus:
@mprahl PTAL

mprahl (approved):
Looks great!

PR #40, opened 2020-11-25
Exclude operator indices that haven't been released

qixiang:
Operator index can exists in Pyxis even the product has not been
released, these indices should be excluded. By checking Product
Page, we can get release date of an OpenShift version, and then
can compare the release date with current date to know whether
the product is released.

JIRA: CLOUDWF-4084

### DIFF
--- a/freshmaker/config.py
+++ b/freshmaker/config.py
@@ -409,6 +409,11 @@ class Config(object):
             'type': str,
             'default': '',
             'desc': 'Query Pyxis for index images only with this organization'
+        },
+        'product_pages_api_url': {
+            'type': str,
+            'default': '',
+            'desc': 'The API URL of the Product Pages service'
         }
     }
 
--- a/freshmaker/pyxis.py
+++ b/freshmaker/pyxis.py
@@ -1,9 +1,12 @@
+import dogpile.cache
 import requests
 import urllib
+from datetime import datetime
 from requests_kerberos import HTTPKerberosAuth, OPTIONAL
 from packaging import version
 
 from freshmaker import log, conf
+from freshmaker.utils import get_ocp_release_date
 
 
 class PyxisRequestError(Exception):
@@ -33,6 +36,8 @@ class PyxisRequestError(Exception):
 
 class Pyxis(object):
     """ Interface for querying Pyxis"""
+
+    region = dogpile.cache.make_region().configure(conf.dogpile_cache_backend)
 
     def __init__(self, server_url):
         self._server_url = server_url
@@ -98,8 +103,30 @@ class Pyxis(object):
         organization = conf.pyxis_index_image_organization
         if organization:
             request_params["filter"] = "organization==" + organization
+        indices = self._pagination("operators/indices", request_params)
 
-        return self._pagination("operators/indices", request_params)
+        # Operator indices can be available in pyxis prior to the Openshift version
+        # is released, so we need to filter out such indices
+        indices = list(filter(lambda x: self.ocp_is_released(x["ocp_version"]), indices))
+        return indices
+
+    @region.cache_on_arguments()
+    def ocp_is_released(self, ocp_version):
+        """ Check if ocp_version is released by comparing the GA date with current date
+
+        :param str ocp_version: the OpenShift Version
+        :return: True if GA date in Product Pages is in the past, otherwise False
+        :rtype: bool
+        """
+        ga_date_str = get_ocp_release_date(ocp_version)
+        # None is returned if GA date is not found
+        if not ga_date_str:
+            log.warning(
+                f"GA date of OpenShift {ocp_version} is not found in Product Pages, ignore it"
+            )
+            return False
+
+        return datetime.now() > datetime.strptime(ga_date_str, "%Y-%m-%d")
 
     def _get_bundles_per_index_image(self, index_images):
         """
--- a/freshmaker/utils.py
+++ b/freshmaker/utils.py
@@ -21,6 +21,7 @@
 #
 
 import functools
+import requests
 import subprocess
 import sys
 import tempfile
@@ -189,3 +190,30 @@ def _run_command(command, logger=None, rundir=None, output=subprocess.PIPE, erro
 def is_pkg_modular(nvr):
     """ Returns True if the package is modular, False otherwise. """
     return "module+" in nvr
+
+
+def get_ocp_release_date(ocp_version):
+    """ Get the OpenShift version release date via the Product Pages API
+
+    :param str ocp_version: the OpenShift version
+    :return: None or date in format of "%Y-%m-%d", example: 2021-02-23.
+    :rtype: str or None
+    """
+    if not conf.product_pages_api_url:
+        raise RuntimeError("Product Pages API url is not set in config")
+
+    ocp_release = f"openshift-{ocp_version}"
+
+    url = f"{conf.product_pages_api_url.rstrip('/')}/releases/{ocp_release}/schedule-tasks"
+    resp = requests.get(
+        url,
+        params={"name": "GA", "fields": "name,date_finish"},
+        timeout=conf.net_timeout,
+    )
+
+    if resp.status_code == 404:
+        log.warning(f"GA date of {ocp_release} is not found via {resp.url}: {resp.reason}")
+        return None
+    if not resp.ok:
+        resp.raise_for_status()
+    return resp.json()[0]['date_finish']
--- a/tests/test_pyxis.py
+++ b/tests/test_pyxis.py
@@ -21,7 +21,10 @@
 # SOFTWARE.
 
 import requests
+import requests_mock
 
+from datetime import datetime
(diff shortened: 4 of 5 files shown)


freshmaker/utils.py
```diff
@@ -189,3 +190,21 @@ def _run_command(command, logger=None, rundir=None, output=subprocess.PIPE, erro
 def is_pkg_modular(nvr):
     """ Returns True if the package is modular, False otherwise. """
     return "module+" in nvr
+
+
+def get_ocp_release_date(ocp_version):
+    """ Get product release data via Product Page API
+
+    :param str ocp_version: ocp version
+    :return: date in format of "%Y-%d-%m", example: 2021-02-23
```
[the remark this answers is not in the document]
qixiang: Good catch, yes, it should be "%Y-%m-%d".

gnaponie (approved):

gnaponie:
Looks good +1

freshmaker/utils.py
```diff
@@ -189,3 +190,21 @@ def _run_command(command, logger=None, rundir=None, output=subprocess.PIPE, erro
 def is_pkg_modular(nvr):
     """ Returns True if the package is modular, False otherwise. """
     return "module+" in nvr
+
+
+def get_ocp_release_date(ocp_version):
+    """ Get product release data via Product Page API
```
mprahl, line 196:
```suggestion
    """ Get product release date via Product Page API
```

freshmaker/utils.py
```diff
@@ -189,3 +190,21 @@ def _run_command(command, logger=None, rundir=None, output=subprocess.PIPE, erro
...
+
+    :param str ocp_version: ocp version
```
mprahl, line 198:
```suggestion
    :param str ocp_version: the OpenShift version
```

freshmaker/utils.py
```diff
@@ -189,3 +190,21 @@ def _run_command(command, logger=None, rundir=None, output=subprocess.PIPE, erro
...
+    :return: date in format of "%Y-%m-%d", example: 2021-02-23
+    :rtype: str
+    """
+    if not conf.product_page_api_url:
+        raise RuntimeError("Product Page API url is not set in config")
+
+    ocp_release = f"openshift-{ocp_version}"
+
+    url = f"{conf.product_page_api_url.rstrip('/')}/releases/{ocp_release}/schedule-tasks"
```
mprahl, line 207: Optional: Could you use [urljoin](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin)?

freshmaker/utils.py
```diff
@@ -189,3 +190,21 @@ def _run_command(command, logger=None, rundir=None, output=subprocess.PIPE, erro
...
+    resp = requests.get(url, params={"name": "GA", "fields": "name,date_finish"})
```
mprahl, line 208: Could you please specify a [timeout](https://requests.readthedocs.io/en/master/user/quickstart/#timeouts)?

freshmaker/config.py
```diff
@@ -409,6 +409,11 @@ class Config(object):
             'type': str,
             'default': '',
             'desc': 'Query Pyxis for index images only with this organization'
+        },
+        'product_page_api_url': {
```
mprahl, line 413:
```suggestion
        'product_pages_api_url': {
```

freshmaker/config.py
```diff
@@ -409,6 +409,11 @@ class Config(object):
...
+            'type': str,
+            'default': '',
+            'desc': 'API url of Product Page service'
```
mprahl, line 416:
```suggestion
            'desc': 'API url of the Product Pages service'
```

freshmaker/pyxis.py
```diff
@@ -98,8 +103,26 @@ def get_operator_indices(self):
...
+
+        :param str ocp_version: ocp version
+        :return: True if GA date in Product Page is in past, othewise False
+        :rtype: bool
+        """
+        ga_date_str = get_ocp_release_date(ocp_version)
+        ga_date = datetime.strptime(ga_date_str, "%Y-%m-%d")
+        if datetime.now() > ga_date:
+            return True
+        return False
```
mprahl, line 125:
```suggestion
        return datetime.now() > ga_date
```

freshmaker/pyxis.py
```diff
@@ -98,8 +103,26 @@ def get_operator_indices(self):
...
+        # is released, so we need to filter out such indices
+        indices = list(filter(lambda x: self.ocp_is_released(x["ocp_version"]), indices))
+        return indices
+
+    @region.cache_on_arguments()
+    def ocp_is_released(self, ocp_version):
+        """ Check if ocp_version is released by comparing the GA date with current date
+
+        :param str ocp_version: ocp version
+        :return: True if GA date in Product Page is in past, othewise False
```
mprahl, line 118:
```suggestion
        :return: True if GA date in Product Page is in the past, otherwise False
```

freshmaker/pyxis.py
```diff
@@ -98,8 +103,26 @@ def get_operator_indices(self):
...
+        # Operator indices can be available in pyxis prior to the Openshift version
+        # is released, so we need to filter out such indices
+        indices = list(filter(lambda x: self.ocp_is_released(x["ocp_version"]), indices))
+        return indices
+
+    @region.cache_on_arguments()
+    def ocp_is_released(self, ocp_version):
+        """ Check if ocp_version is released by comparing the GA date with current date
+
+        :param str ocp_version: ocp version
```
mprahl, line 117:
```suggestion
        :param str ocp_version: the OpenShift version
```

freshmaker/utils.py
```diff
@@ -189,3 +190,21 @@ def _run_command(command, logger=None, rundir=None, output=subprocess.PIPE, erro
...
+    resp.raise_for_status()
```
mprahl, line 209: Optional: I'm thinking it shouldn't raise an exception if the version is not found. Perhaps just a warning and ignore it as a bogus version in Pyxis.

tests/test_pyxis.py
```diff
@@ -345,6 +348,47 @@ def test_get_operator_indices(self, page):
         page.assert_called_once_with(
             'operators/indices', {'filter': 'organization==org'})
 
+    @patch.object(conf, 'product_page_api_url', new='http://pp.abc.org/api')
```
mprahl, line 351: Could you use double quotes to be consistent with the test please?

freshmaker/utils.py
qixiang: `product_pages_api_url` (the base URL) contains the path to API, not just hostname, `urljoin` is used for get the full URL from a base URL and a relative URL, it can be easy to make mistake when we have path in baseurl and just want to concat the urls, so I'll just leave this as it is.

qixiang:
Hi @mprahl , all your comments have been addressed except one about using `urljoin`, PTAL.

freshmaker/utils.py
mprahl: @qixiang good point.

freshmaker/utils.py
```diff
@@ -189,3 +190,27 @@ def _run_command(command, logger=None, rundir=None, output=subprocess.PIPE, erro
...
+
+    url = f"{conf.product_pages_api_url.rstrip('/')}/releases/{ocp_release}/schedule-tasks"
+    resp = requests.get(
+        url,
+        params={"name": "GA", "fields": "name,date_finish"},
+        timeout=conf.net_timeout,
+    )
+    if not resp.ok:
+        log.warning(f"GA date of {ocp_release} is not found via {resp.url}: {resp.reason}")
+        return None
```
mprahl, line 215: I think we still want to raise for status if the error is anything but it being not found.
```suggestion
    if resp.status_code == 404:
        log.warning(f"GA date of {ocp_release} is not found via {resp.url}: {resp.reason}")
        return None
    resp.raise_for_status()
```

freshmaker/utils.py
```diff
@@ -189,3 +190,27 @@ def _run_command(command, logger=None, rundir=None, output=subprocess.PIPE, erro
...
     """ Returns True if the package is modular, False otherwise. """
     return "module+" in nvr
+
+
+def get_ocp_release_date(ocp_version):
+    """ Get the OpenShift version release date via the Product Pages API
+
+    :param str ocp_version: the OpenShift version
+    :return: None or date in format of "%Y-%m-%d", example: 2021-02-23.
+    :rtype: str
```
mprahl, line 200:
```suggestion
    :rtype: str or None
```

qixiang:
Patch is updated to address the last 2 comments from Matt.

qixiang:
Hi @apaplaus @gnaponie could you help to review this? I'm trying to merge this before our sprint close, Matt has two minor comments, and have been addressed in the latest version.

apaplaus:
Looks nice

apaplaus (approved):

gnaponie:
+1

PR #42, opened 2020-12-02
Migrate to koji gssapi_login API

qixiang:
Use koji's gssapi_login API instead of krb_login

krb_login() is deprecated, and for ClientSession objects, krb_login()
is currently redirected to gssapi_login() with a printed warning.

More details:
https://docs.pagure.org/koji/migrations/migrating_to_1.22/#migration-krbv

And add python requests-kerberos package explicitly in yum-packages.txt
and requirements.txt.


gnaponie:
No old requirement can be removed? It seems not...
Looks good +1

gnaponie (approved):

apaplaus (approved):

PR #44, opened 2020-12-08
Get original nvrs from incoming message

apaplaus:
Fix bug when original nvrs were extracted from new event instead of
searching for advisory builds and then searching Freshmaker db for older
build events.

### DIFF
--- a/freshmaker/handlers/botas/botas_shipped_advisory.py
+++ b/freshmaker/handlers/botas/botas_shipped_advisory.py
@@ -22,7 +22,7 @@
 from freshmaker import db, conf, log
 from freshmaker.handlers import ContainerBuildHandler
 from freshmaker.events import BotasErrataShippedEvent
-from freshmaker.models import Event
+from freshmaker.models import ArtifactBuild, ArtifactType, Event
 from freshmaker.types import EventState
 from freshmaker.pyxis import Pyxis
 from freshmaker.kojiservice import koji_service
@@ -56,8 +56,6 @@ class HandleBotasAdvisory(ContainerBuildHandler):
             self.force_dry_run()
         self.event = event
 
-        # Get event from database or create new one.
-        # Then we can get original NVRs from it.
         db_event = Event.get_or_create_from_event(db.session, event)
 
         self.set_context(db_event)
@@ -70,15 +68,25 @@ class HandleBotasAdvisory(ContainerBuildHandler):
             self.log_info(msg)
             return []
 
-        # Get original nvrs of all builds in the advisory
-        original_nvrs = [build.original_nvr for build in db_event.builds]
+        # Get builds NVRs from the advisory attached to the message/event and
+        # then get original NVR for every build
+        original_nvrs = set()
+        for product_info in event.advisory.builds.values():
+            for build in product_info['builds']:
+                # Search for the first build that triggered the chain of rebuilds
+                # for every shipped NVR to get original NVR from it
+                original_nvr = self.get_published_original_nvr(build['nvr'])
+                if original_nvr is None:
+                    continue
+                original_nvrs.add(original_nvr)
+
         self.log_info(
             "Orignial nvrs of build in the advisory #{0} are: {1}".format(
-                db_event.search_key, " ".join(original_nvrs)))
+                event.advisory.errata_id, " ".join(original_nvrs)))
         # Get images by nvrs and then get their digests
-        original_images_digests = set(self._pyxis.get_digests_by_nvrs(original_nvrs))
+        original_images_digests = self._pyxis.get_digests_by_nvrs(original_nvrs)
         if not original_images_digests:
-            msg = f"The are no digests for NVRs: {','.join(original_nvrs)}"
+            msg = f"There are no digests for NVRs: {','.join(original_nvrs)}"
             log.warning(msg)
             db_event.transition(EventState.SKIPPED, msg)
             return []
@@ -149,3 +157,39 @@ class HandleBotasAdvisory(ContainerBuildHandler):
                     continue
                 ret_bundle_images_nvrs.add(nvr)
         return ret_bundle_images_nvrs
+
+    def get_published_original_nvr(self, rebuilt_nvr):
+        """
+        Search for an original build, that has been built and published to a
+            repository, and get original_nvr from it
+
+        :param str rebuilt_nvr: rebuilt NVR to look build by
+        :rtype: str or None
+        :return: original NVR from the first published FM build for given NVR
+        """
+        original_nvr = None
+        # artifact build should be only one in database, or raise an error
+        artifact_build = db.session.query(ArtifactBuild).filter(
+            ArtifactBuild.rebuilt_nvr == rebuilt_nvr,
+            ArtifactBuild.type == ArtifactType.IMAGE.value,
+        ).one_or_none()
+        # recursively search for original artifact build
+        if artifact_build is not None:
+            original_nvr = artifact_build.original_nvr
+
+            # check if image is published
+            request_params = {'include': 'data.repositories',
+                              'page_size': 1}
+            images = self._pyxis._pagination(f'images/nvr/{original_nvr}',
+                                             request_params)
+            if not images:
+                return None
+            # stop recursion if the image is published in some repo
+            if any(repo['published'] for repo in images[0].get('repositories')):
+                return original_nvr
+
+            next_nvr = self.get_published_original_nvr(original_nvr)
+            if next_nvr is not None:
+                original_nvr = next_nvr
+
+        return original_nvr
--- a/tests/handlers/botas/test_botas_shipped_advisory.py
+++ b/tests/handlers/botas/test_botas_shipped_advisory.py
@@ -24,10 +24,11 @@ from unittest.mock import patch, call, MagicMock
 from freshmaker import db, conf
 from freshmaker.events import (
     BotasErrataShippedEvent,
-    ManualRebuildWithAdvisoryEvent)
+    ManualRebuildWithAdvisoryEvent,
+    TestingEvent)
 from freshmaker.handlers.botas import HandleBotasAdvisory
 from freshmaker.errata import ErrataAdvisory
-from freshmaker.models import Event
+from freshmaker.models import Event, ArtifactBuild
 from freshmaker.types import EventState
 from tests import helpers
 
@@ -101,16 +102,39 @@ class TestBotasShippedAdvisory(helpers.ModelsTestCase):
                 'advisory_name': 'RHBA-2020'
             }
         }})
+    @patch("freshmaker.handlers.botas.botas_shipped_advisory.HandleBotasAdvisory.get_published_original_nvr")
+    def test_get_original_nvrs(self, get_build):
+        event = BotasErrataShippedEvent("test_msg_id", self.botas_advisory)
+        self.botas_advisory._builds = {
+            "product_name": {
+                "builds": [{"nvr": "some_name-2-2"},
+                           {"nvr": "some_name_two-2-2"}]
+            }
+        }
+        get_build.return_value = "some_name-1-0"
+
+        self.handler.handle(event)
+
+        self.pyxis().get_digests_by_nvrs.assert_called_with({"some_name-1-0"})
+
+    @patch.object(conf, 'dry_run', new=True)
+    @patch.object(conf, 'handler_build_allowlist', new={
+        'HandleBotasAdvisory': {
+            'image': {
+                'advisory_name': 'RHBA-2020'
+            }
+        }})
     def test_handle_no_digests_error(self):
         event = BotasErrataShippedEvent("test_msg_id", self.botas_advisory)
         self.pyxis().get_digests_by_nvrs.return_value = set()
+        self.botas_advisory._builds = {}
 
         self.handler.handle(event)
         db_event = Event.get(db.session, message_id='test_msg_id')
 
         self.assertEqual(db_event.state, EventState.SKIPPED.value)
         self.assertTrue(
-            db_event.state_reason.startswith("The are no digests for NVRs:"))
+            db_event.state_reason.startswith("There are no digests for NVRs:"))
 
     @patch.object(conf, 'dry_run', new=True)
     @patch.object(conf, 'handler_build_allowlist', new={
@@ -150,6 +174,7 @@ class TestBotasShippedAdvisory(helpers.ModelsTestCase):
             },
         ]
         self.pyxis().filter_bundles_by_related_image_digests.return_value = bundles
+        self.botas_advisory._builds = {}
 
         self.handler.handle(event)
         db_event = Event.get(db.session, message_id='test_msg_id')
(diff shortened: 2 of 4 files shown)


freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -70,11 +68,37 @@ def handle(self, event):
             self.log_info(msg)
             return []
 
-        # Get original nvrs of all builds in the advisory
-        original_nvrs = [build.original_nvr for build in db_event.builds]
+        # Get builds NVRs from the advisory attached to the message/event
+        affected_nvrs = set()
```
qixiang, line 72: Both `original_nvrs` and `affected_nvrs` are misleading, can we rename it to `shipped_nvrs`?

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -70,11 +68,37 @@ def handle(self, event):
...
+        for product_info in event.advisory.builds.values():
+            for build in product_info['builds']:
+                affected_nvrs.add(build['nvr'])
+
+        # Search for the first event that triggered the chain of rebuilds
```
qixiang, line 77: I don't think we need to get the event first, why not just get the original nvr in `models.ArtifactBuild`?

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -70,11 +68,37 @@ def handle(self, event):
...
+                msg = f"Can't find old build event for nvr : {original_nvr}"
+                self.log_warn(msg)
+                continue
+            old_db_events_list.append(olb_db_event)
+        if not old_db_events_list:
+            msg = f"Can't find any previous build event for nvrs: {affected_nvrs}"
+            self.log_warn(msg)
+            return []
+
+        # Get original nvrs of all old builds of rpms in the advisory
```
qixiang, line 94: An advisory filed by BOTAS can contains only part of the rebuilt images in an event, we only want the original NVRs for the shipped NVRs in this advisory, not all the original NVRs in the event.

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: It's because our models.Event probably will not have any builds attached(since we didn't do any? I am not sure). So by calling `event.advisory.builds` I am calling errata api to get builds for the attached advisory.

qixiang: Which event do you mean? When an advisory filed by botas is shipped, it always contains some images, an advisory can not be moved to QE or shipped without any build (except Text only advisories, but we're dealing with container advisories). In this shipped advisory, the attached images are always rebuilt by Freshmaker, for example, it contains an image `foobar-1.2-1.1607508530`, this image was rebuilt by Freshmaker and we can find the original NVR by  `ArtifactBuild.query.filter_by(rebuilt_nvr="foobar-1.2-1.1607508530").one()`.

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: Yes you are right, didn't think about it. I will fix it.

freshmaker/models.py
```diff
@@ -617,6 +617,33 @@ def get_lowest_build_id(cls, session):
...
+        original_nvr = None
+        # build event should be only one in database, or raise an error
+        build_event = db.session.query(ArtifactBuild).filter(
+            ArtifactBuild.rebuilt_nvr == rebuilt_nvr,
+            ArtifactBuild.type == ArtifactType.IMAGE.value,
+        ).distinct().one_or_none()
+        # recursively search for original build event
+        if build_event is not None:
+            original_nvr = build_event.original_nvr
+            next_nvr = cls.get_original_nvr_by_rebuilt_nvr(session,
```
qixiang, line 640: Any reason of get the original image recursively? as an example: original or original -> original -> rebuilt_nvr, this will ignore the original image and is not what we want.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -70,15 +68,24 @@ def handle(self, event):
...
-        # Get original nvrs of all builds in the advisory
-        original_nvrs = [build.original_nvr for build in db_event.builds]
+        # Get builds NVRs from the advisory attached to the message/event and
+        # then get original NVR for every build
+        original_nvrs = set()
+        for product_info in event.advisory.builds.values():
+            for build in product_info['builds']:
+                # Search for the first build that triggered the chain of rebuilds
+                # for every shipped NVR to get original NVR from it
+                original_nvr = ArtifactBuild.get_original_nvr_by_rebuilt_nvr(
```
qixiang, line 78: I don't think this class method is necessary, we can just filter `ArtifactBuild` with the `rebuild_nvr` directly

qixiang:
Added comments, could you also update the commit message?

freshmaker/models.py
apaplaus: I am not quite understand. There is always suppose to be previous rebuild event with the **olderEvent.rebuilt_nvr** == **currentlyProcessedEvent.orignial_nvr**. So if there is at least one event with that condition, this method will return **olderEvent.original_nvr**. And the reason of getting it recursively is because it reflects logic how we suppose to search for **original_nvr**, just searching every time previous event untill we can't find any, and return original_nvr of the last found build event.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -70,15 +68,24 @@ def handle(self, event):
...
+                    db.session, build['nvr'])
+                original_nvrs.add(original_nvr)
```
mprahl, line 80: What if `original_nvr` is `None`? This can happen if someone attached a non-Freshmaker build to a BOTAS advisory. It's an unsupported behavior but we've seen it happen.

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: I can move it from the class but I would at least make a separate method for it, just to not overextend the `handle()` method. And why it's a class method, I thought that all logic including querying particular tables should be in `models.` class of that table, as in `models.Event` for example

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: Yes you are right, I should check for None.

freshmaker/models.py
```diff
@@ -617,6 +617,33 @@ def get_lowest_build_id(cls, session):
...
+        :param str rebuilt_nvr: rebuilt NVR to look build by
+        :rtype: str or None
+        :return: original NVR from the first FM build for given NVR
+        """
+        original_nvr = None
+        # build event should be only one in database, or raise an error
+        build_event = db.session.query(ArtifactBuild).filter(
+            ArtifactBuild.rebuilt_nvr == rebuilt_nvr,
+            ArtifactBuild.type == ArtifactType.IMAGE.value,
+        ).distinct().one_or_none()
```
mprahl, line 636: Could you please explain why `distinct` is required here?

apaplaus: Oh yes, I have it because there was a join previously, I will remove it.

mprahl:
This looks good! I just left a couple of small comments.

freshmaker/models.py
mprahl: @qixiang the reason is that if Release Driver has rebuilt this image for conflict resolution, the `original_nvr` can actually still be a `rebuilt_nvr`. The recursion will ensure the actual `original_nvr` that is released.

That being said, I noticed a bug in this logic that I had recommended. If `original_nvr` was a previous Freshmaker rebuild that was released, then we actually don't want to continue the recursion. Can you stop the recursion if the event type that triggered the build is `ErrataAdvisoryRPMsSignedEvent` (I think that is the correct name)?

apaplaus: No I don't, @mprahl so if I see that event is of type **ErrataAdvisoryRPMsSignedEvent** I should stop and use that event's **original_nvr** or the **original_nvr** of the previously found event?

mprahl: @apaplaus sorry, I meant to say "Could you please stop the recursion....".

That's what I'm thinking. The recursion stops if no build is found or the build was triggered by an event of type `ErrataAdvisoryRPMsSignedEvent`.

apaplaus: So will it be enough?
```
        if build_event is not None:
            # check if event that triggered the build is released
            if build_event.event.event_type_id == ErrataAdvisoryRPMsSignedEvent:
                return None
            original_nvr = build_event.original_nvr
            next_nvr = cls.get_original_nvr_by_rebuilt_nvr(session, original_nvr)
```

mprahl: `if build_event.event.event_type_id == ErrataAdvisoryRPMsSignedEvent:` => `if original_nvr and build_event.event.event_type_id == ErrataAdvisoryRPMsSignedEvent:`

apaplaus: I didn't get it. I assign **original_nvr** after this conditional so at that point it always will be None. And event if I change their places, why would I check orignal_nvr if I still returning None? You are suggesting to do it like this?:
```
 if build_event is not None:
            original_nvr = build_event.original_nvr
            # check if event that triggered the build is released
            if original_nvr and build_event.event.event_type_id == ErrataAdvisoryRPMsSignedEvent:
                return None
 
            next_nvr = cls.get_original_nvr_by_rebuilt_nvr(session, original_nvr)
```

mprahl: Yes, except that now you can go back to your original proposed if statement of `if build_event.event.event_type_id == ErrataAdvisoryRPMsSignedEvent:`, but return `original_nvr` and not `None`.

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: @qixiang so you think it's better to move it outside of `ArtifactBuild` class?

qixiang: If we can just get the original nvr with one line query filter, it doesn't worth to have separate class method, now I'm thinking about the case of looking for the orignal nvr recursively, I'll add comment in the other thread.

freshmaker/models.py
qixiang: If we have the following case:

1. `foobar-1-2` is an image rebuilt by product team
2. ` foobar-1-2.xxx` is an image rebuilt for `foobar-1-2`
3. `foobar-1-2.yyy` is an image rebulit for `foobar-1-2.xxx`

I believe when `foobar-1-2.yyy` is shipped, `foobar-1-2.xxx` should always been shipped already, and the bundle image was rebuilt to replace `foobar-1-2` with `foobar-1-2.xxx` already in previous this type of event, so I don't think we need to get the original_nvr recursively.

apaplaus: I think we should ask @mprahl. It was his thought originally, so I am not competent enough in this question =)

mprahl: @qixiang the Release Driver case I'm thinking is as follows. Please correct me if I'm wrong.
CVE2 (operator-1.2.124) -> CVE1 (operator-1.2.123)

Release Driver would request a rebuild of operator-1.2.124 to include the CVE fix from operator-1.2.123. CVE2 now has the build of operator-1.2.125 attached. Assume no other changes and the advisories ship.

Once shipped, Freshmaker queries for the original NVR of operator-1.2.125, but the original NVR would be operator-1.2.124 which was never shipped. Freshmaker would then determine no bundle image references the digest of operator-1.2.124 and thus skip any bundle image rebuilds of that. 

Does that make sense?

qixiang: Maybe we should check whether the original_nvr is a published image instead of checking the event type. Because before all this happens, the latest published image can be an image rebuilt by freshmaker, not an image built by product team, so we should stop looking for the original_nvr when an image is a published image.

Even with this, I still see a gap here in your example, both `operator-1-2.123` and `operator-1-2.124` were rebuilt for `operator-1-2`, when `operator-1-2.123` was shipped, the `operator-1-2` in bundle image was replaced by `operator-1-2.123`, and then when `operator-1-2.124` or even `operator-1-2.125` is shipped, the bundle image will not be updated to replace `operator-1-2.123` with `operator-1-2.124` or `operator-1-2.125`, because `operator-1-2.123` is not the original nvr of `operator-1-2.124` or `operator-1-2.125`.

mprahl: @qixiang yes... very good point.

Let's implement your idea of recursively getting the original NVR until the original NVR is published. That is the safest approach.

In addition, lets have Freshmaker look at the advisories that the advisory currently being processed depends on. For all of those advisories, find the container images with the same "package" name and also consider those as operator/operand images to be replaced by the operator/operand image on the advisory being processed.

What do you think?

qixiang: @mprahl I agree, this approach looks good to me.

mprahl: @apaplaus is this approach clear? If not, feel free to reach out over chat.

mprahl: @apaplaus I filed CLOUDWF-4548 to address the second part (traversing the advisory dependency graph). If you can, please implement @qixiang's suggestion of modifying the recursion to stop when there is either no original NVR or an original NVR is found that is published.

apaplaus: Gotcha =)

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -149,3 +158,39 @@ def _filter_bundles_by_pinned_related_images(self, bundle_image_nvrs):
                     continue
                 ret_bundle_images_nvrs.add(nvr)
         return ret_bundle_images_nvrs
+
+    def get_published_original_nvr(self, session, rebuilt_nvr):
+        """
+        Search for an original build, that has been built and published to a
+            repository, and get original_nvr from it
+
+        :param session: database session
```
mprahl, line 167: Why are you passing in session as an argument but using `db.session` directly? I would just remove this argument.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -149,3 +158,39 @@ def _filter_bundles_by_pinned_related_images(self, bundle_image_nvrs):
...
+            original_nvr = build_event.original_nvr
+
+            # check if image is published
+            request_params = {'include': 'data.repositories'}
+            images = self._pyxis._pagination(f'images/nvr/{original_nvr}',
+                                             request_params)
+            if not images:
+                return None
+            # stop recursion if the image is published in some repo
+            if any([repo['published'] for repo in images[0].get('repositories')]):
```
mprahl, line 189: This is a minor optimization that allows `any` to stop processing the generator as soon as it matches a repo that is published.
```suggestion
            if any(repo['published'] for repo in images[0].get('repositories')):
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -149,3 +158,39 @@ def _filter_bundles_by_pinned_related_images(self, bundle_image_nvrs):
...
+        build_event = db.session.query(ArtifactBuild).filter(
+            ArtifactBuild.rebuilt_nvr == rebuilt_nvr,
+            ArtifactBuild.type == ArtifactType.IMAGE.value,
+        ).one_or_none()
+        # recursively search for original build event
+        if build_event is not None:
+            original_nvr = build_event.original_nvr
+
+            # check if image is published
+            request_params = {'include': 'data.repositories'}
```
mprahl, line 183: I would add `page_size` to `1` for a minor optimization here since you only look at the first entry anyways.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -149,3 +158,39 @@ def _filter_bundles_by_pinned_related_images(self, bundle_image_nvrs):
...
+            repository, and get original_nvr from it
+
+        :param session: database session
+        :param str rebuilt_nvr: rebuilt NVR to look build by
+        :rtype: str or None
+        :return: original NVR from the first published FM build for given NVR
+        """
+        original_nvr = None
+        # build event should be only one in database, or raise an error
+        build_event = db.session.query(ArtifactBuild).filter(
```
mprahl, line 174: `build_event` is misleading. It sounds like a Freshmaker event but it's not. Could you please use `artifact_build` or something like that?

tests/handlers/botas/test_botas_shipped_advisory.py
```diff
@@ -207,3 +232,47 @@ def test_filter_bundles_by_pinned_related_images(self, service):
         temp_mock.get_build.assert_has_calls([call(bundle_list[0]),
                                              call(bundle_list[1])])
         self.assertEqual(nvrs, {bundle_list[0]})
+
+    def test_no_original_nvr_found(self):
```
mprahl, line 236: I don't understand why this test is named this way. Could you please explain?

tests/handlers/botas/test_botas_shipped_advisory.py
```diff
@@ -207,3 +232,47 @@ def test_filter_bundles_by_pinned_related_images(self, service):
...
+                             rebuilt_nvr="nvr1-0-2")
+        db.session.commit()
+        self.pyxis()._pagination.return_value = [
+            {"repositories": [{"published": True}]}
+        ]
+
+        ret_nvr = self.handler.get_published_original_nvr(db.session, "nvr1-0-2")
+        self.assertEqual(ret_nvr, 'nvr1-0-1')
+
+    def test_get_published_original_nvr(self):
```
mprahl, line 249: To make the test better, could you assign each build to a separate event? Bonus points if you use the `ManualRebuildWithAdvisoryEvent` event type for the last two builds.

tests/handlers/botas/test_botas_shipped_advisory.py
apaplaus: It's tests that `get_published_original_nvr` will return the first entry when there are no other events to test. That's why I thought about that name.

tests/handlers/botas/test_botas_shipped_advisory.py
apaplaus: Previously I had it like that, but I thought that since I don't query Events table at all there is no need in separate events.

tests/handlers/botas/test_botas_shipped_advisory.py
mprahl: How about renaming this to `test_get_published_original_nvr_single_event` or something similar?

mprahl (approved):

qixiang (approved):

PR #45, opened 2021-01-11
Add pullspec overrides data for related images in each bundle image

qixiang:
For each bundle image we're going to rebuild when operator/operand
images are shipped, pullspec overrides data is required, the general
workflow is: 

1. Get the current image NVRs in advisory, these are rebuilt images
2. Get the original image NVRs for rebuilt images
3. Get digests of original images
4. Get digests of rebuilt images
5. With digests of original images, we can search in all latest bundle
   images to get the bundles which have original images are connected.
6. For each bundle image, if this image is not enabled for rebuild, skip
   it.
7. For each bundle image, if a related image is in original digests (step
   3), and osbs-pinning is used, update the digest with digest of rebuilt
   image.

JIRA: CLOUDWF-2902

### DIFF
--- a/freshmaker/handlers/botas/botas_shipped_advisory.py
+++ b/freshmaker/handlers/botas/botas_shipped_advisory.py
@@ -18,6 +18,7 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
+import copy
 
 from freshmaker import db, conf, log
 from freshmaker.handlers import ContainerBuildHandler
@@ -25,7 +26,7 @@ from freshmaker.events import BotasErrataShippedEvent
 from freshmaker.models import ArtifactBuild, ArtifactType, Event
 from freshmaker.types import EventState
 from freshmaker.pyxis import Pyxis
-from freshmaker.kojiservice import koji_service
+from freshmaker.kojiservice import KojiService
 
 
 class HandleBotasAdvisory(ContainerBuildHandler):
@@ -70,7 +71,9 @@ class HandleBotasAdvisory(ContainerBuildHandler):
 
         # Get builds NVRs from the advisory attached to the message/event and
         # then get original NVR for every build
-        original_nvrs = set()
+
+        # Mapping of original build nvrs to rebuilt nvrs in advisory
+        nvrs_mapping = {}
         for product_info in event.advisory.builds.values():
             for build in product_info['builds']:
                 # Search for the first build that triggered the chain of rebuilds
@@ -78,15 +81,50 @@ class HandleBotasAdvisory(ContainerBuildHandler):
                 original_nvr = self.get_published_original_nvr(build['nvr'])
                 if original_nvr is None:
                     continue
-                original_nvrs.add(original_nvr)
+                nvrs_mapping[original_nvr] = build['nvr']
 
+        original_nvrs = nvrs_mapping.keys()
         self.log_info(
             "Orignial nvrs of build in the advisory #{0} are: {1}".format(
                 event.advisory.errata_id, " ".join(original_nvrs)))
-        # Get images by nvrs and then get their digests
-        original_images_digests = self._pyxis.get_digests_by_nvrs(original_nvrs)
-        if not original_images_digests:
-            msg = f"There are no digests for NVRs: {','.join(original_nvrs)}"
+
+        # Get image manifest_list_digest for all original images, manifest_list_digest is used
+        # in pullspecs in bundle's related images
+        original_digests_by_nvr = {}
+        original_nvrs_by_digest = {}
+        for nvr in original_nvrs:
+            digest = self._pyxis.get_manifest_list_digest_by_nvr(nvr)
+            if digest:
+                original_digests_by_nvr[nvr] = digest
+                original_nvrs_by_digest[digest] = nvr
+            else:
+                log.warning(
+                    f"Image manifest_list_digest not found for original image {nvr} in Pyxis, "
+                    "skip this image"
+                )
+
+        if not original_digests_by_nvr:
+            msg = f"None of the original images have digests in Pyxis: {','.join(original_nvrs)}"
+            log.warning(msg)
+            db_event.transition(EventState.SKIPPED, msg)
+            return []
+
+        # Get image manifest_list_digest for all rebuilt images, manifest_list_digest is used
+        # in pullspecs of bundle's related images
+        rebuilt_digests_by_nvr = {}
+        rebuilt_nvrs = nvrs_mapping.values()
+        for nvr in rebuilt_nvrs:
+            digest = self._pyxis.get_manifest_list_digest_by_nvr(nvr)
+            if digest:
+                rebuilt_digests_by_nvr[nvr] = digest
+            else:
+                log.warning(
+                    f"Image manifest_list_digest not found for rebuilt image {nvr} in Pyxis, "
+                    "skip this image"
+                )
+
+        if not rebuilt_digests_by_nvr:
+            msg = f"None of the rebuilt images have digests in Pyxis: {','.join(rebuilt_nvrs)}"
             log.warning(msg)
             db_event.transition(EventState.SKIPPED, msg)
             return []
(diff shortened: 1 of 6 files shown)


apaplaus (commented):
LGFM except for some little comments, but I would still wait for Matt and for what he is famous for.... _his review_

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -96,15 +127,29 @@ def handle(self, event):
...
+                image_digest, all_bundles
+            )
+            if not bundles:
+                log.info(f"No latest bundle image found for {image_nvr}")
                 continue
-            bundle_digests.add(bundle['bundle_path_digest'])
+
+            for bundle in bundles:
+                bundle_digest = bundle.get('bundle_path_digest')
+                if not bundle_digest:
```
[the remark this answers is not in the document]
qixiang: ok, I checked bundles data in pyxis, all bundles in redhat registry have `bundle_path_digest`, quay.io doesn't have such field, so yes, I agree we can remove this, for quay.io bundles, that's another story if we need to care about.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -115,6 +160,27 @@ def handle(self, event):
...
+            for image_nvr, digests in bundle_digests_by_related_nvr.items():
+                if bundle_digest not in digests:
+                    continue
+
+                # this original image NVR connects to bundle_nvr
+                original_digest = original_digests_by_nvr[image_nvr]
+                rebuilt_digest = rebuilt_digests_by_nvr[nvrs_mapping[image_nvr]]
+
+                pullspec_overrides_by_nvr.setdefault(bundle_nvr, {})
+                pullspec_overrides_by_nvr[bundle_nvr][original_digest] = rebuilt_digest
```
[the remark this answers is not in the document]
qixiang: This is because we're adding multiple `original_digest: rebuilt_digest` to a single bundle image within the loop of 
```
for image_nvr, digests in bundle_digests_by_related_nvr.items():
```
multiple rebuilt images in an advisory can connect to a same bundle image, this is a common case.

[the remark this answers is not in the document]
apaplaus: Oh ok, now I got it. Thx

qixiang:
Hi @mprahl could you help to review this?

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -70,23 +70,54 @@ def handle(self, event):
...
+            digest = self._pyxis.get_digest_by_nvr(nvr)
+            if digest:
+                original_digests_by_nvr[nvr] = digest
+            else:
+                log.warning(
+                    f"Image manifest_list_digest not found for {nvr} in Pyxis, skip this image"
+                )
+
+        if not original_digests_by_nvr:
+            msg = f"None of the original images have digest in Pyxis: {','.join(original_nvrs)}"
```
mprahl, line 102:
```suggestion
            msg = f"None of the original images have digests in Pyxis: {','.join(original_nvrs)}"
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -70,23 +70,54 @@ def handle(self, event):
...
+            digest = self._pyxis.get_digest_by_nvr(nvr)
+            if digest:
+                rebuilt_digests_by_nvr[nvr] = digest
+            else:
+                log.warning(
+                    f"Image manifest_list_digest not found for {nvr} in Pyxis, skip this image"
+                )
+
+        if not rebuilt_digests_by_nvr:
+            msg = f"None of the rebuilt images have digest in Pyxis: {','.join(rebuilt_nvrs)}"
```
mprahl, line 120:
```suggestion
            msg = f"None of the rebuilt images have digests in Pyxis: {','.join(rebuilt_nvrs)}"
```

freshmaker/pyxis.py
```diff
@@ -190,48 +190,39 @@ def get_latest_bundles(self, index_images):
...
-        # If bundle doesn't have any of digests of images, don't add it to return list
+        ret = []
         for bundle in bundles:
-            for image in bundle.get('related_images', []):
-                # If same digest within images' digests is found, it will be in return list
-                if image.get('digest') in original_digests:
-                    ret_bundles.append(bundle)
-                    break
+            related_digests = [img.get('digest') for img in bundle.get('related_images', [])]
+            if image_digest in related_digests:
```
mprahl, line 222: This would avoid looping through all the related images when there is a match.
```suggestion
            if any (image_digest == img.get('digest') for img in bundle.get('related_images', [])):
```

freshmaker/pyxis.py
```diff
@@ -190,48 +190,39 @@ def get_latest_bundles(self, index_images):
...
+            for repo in image.get('repositories'):
+                if repo['published'] and 'manifest_list_digest' in repo:
+                    return repo['manifest_list_digest']
+        return None
+
+    def get_bundles_by_related_image_digest(self, image_digest, bundles):
         """
-        Filter out bundles that don't have at least one 'related_image'
-        with the same manifest list digest as those in 'original_digests'
+        Get latest bundles that have the specified image digest in related images.
```
mprahl, line 212: I suggest this change in wording since the method does not determine what is latest, the input to is what determines that.
```suggestion
        Get bundles that have the specified image digest in related images.
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -96,15 +127,23 @@ def handle(self, event):
...
+            bundles = self._pyxis.get_bundles_by_related_image_digest(
+                image_digest, all_bundles
+            )
+            if not bundles:
+                log.info(f"No latest bundle image found for {image_nvr}")
                 continue
-            bundle_digests.add(bundle['bundle_path_digest'])
+
+            for bundle in bundles:
+                bundle_digest = bundle.get('bundle_path_digest')
```
mprahl, line 141: If I recall correctly, there is a bug that causes `bundle_path_digest` not to be set in the data returned by Pyxis. So we should guard against this like the previous code. If you've confirmed that this is no longer the case, then access the key directly to indicate that it is always set (i.e. don't use `.get`).

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -115,6 +154,27 @@ def handle(self, event):
         bundles_nvrs = self._filter_bundles_by_pinned_related_images(
             auto_rebuild_nvrs)
 
+        # Now we have the bundle NVRs needs to be rebuilt, for each bundle NVR,
```
mprahl, line 157:
```suggestion
        # Now we have the bundle NVRs that need to be rebuilt, for each bundle NVR,
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -96,15 +127,23 @@ def handle(self, event):
...
+            if not bundles:
+                log.info(f"No latest bundle image found for {image_nvr}")
                 continue
-            bundle_digests.add(bundle['bundle_path_digest'])
+
+            for bundle in bundles:
+                bundle_digest = bundle.get('bundle_path_digest')
+                bundle_digests_by_related_nvr.setdefault(image_nvr, []).append(bundle_digest)
+
+        bundle_digests = set()
```
mprahl, line 144: It'd probably be safer to just skip the event if `bundle_digests_by_related_nvr` is empty at this point. We have no guarantee that the original NVRs of the images in the BOTAS advisory were ever part of a bundle image.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -115,6 +154,27 @@ def handle(self, event):
         bundles_nvrs = self._filter_bundles_by_pinned_related_images(
             auto_rebuild_nvrs)
 
+        # Now we have the bundle NVRs needs to be rebuilt, for each bundle NVR,
+        # construct the data of related images pullspec overrides
+        pullspec_overrides_by_nvr = {}
+        for bundle_nvr in bundles_nvrs:
+            images = [i for i in bundle_images if i['brew']['build'] == bundle_nvr]
```
mprahl, line 161: Rather than loop through `bundle_images` for each `nvr`, could you loop through it once outside of this for loop and create a dictionary of NVR to image. Then this for loop can be changed to iterate through that dictionary.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -115,6 +154,27 @@ def handle(self, event):
         bundles_nvrs = self._filter_bundles_by_pinned_related_images(
             auto_rebuild_nvrs)
 
+        # Now we have the bundle NVRs needs to be rebuilt, for each bundle NVR,
+        # construct the data of related images pullspec overrides
+        pullspec_overrides_by_nvr = {}
```
mprahl, line 159: We need to end up with the same format that OSBS uses in the Brew build extra field of `image.operator_manifests.related_images.pullspecs`. You can see an example of this with the Brew build with the ID of 1464389.

It also needs to include the pullspecs that Freshmaker did not replace. So in summary, use the dictionary of `image.operator_manifests.related_images.pullspecs` as a starting point. Then update the references of the `new` fields for pullspecs that were updated to be the rebuilt digest instead of the vulnerable one.

mprahl:
@qixiang I think this PR could be simplified in general by starting from the original code and add a mapping of original_nvr digest to rebuilt_nvr digest (like you did) and then refactor `_filter_bundles_by_pinned_related_images` so that it returns a dictionary that has a mapping of bundle NVR to `image.operator_manifests.related_images.pullspecs` (from the Brew build extra data).

From there, loop through all the items in the new dictionary just mentioned and loop through each `pullspecs` list (i.e. the value of the dictionary item). If one of the `new` fields contains a pullspec with the digest of an `original_nvr`, replace just the digest with the new digest of the `rebuilt_nvr`.

The dictionary now contain the pullspec replacement metadata that OSBS needs.

freshmaker/migrations/versions/fcba8824bf8d_add_bundle_pullspec_overrides.py
```diff
@@ -0,0 +1,26 @@
...
+# revision identifiers, used by Alembic.
+revision = 'fcba8824bf8d'
+down_revision = '2358b6f55f24'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
mprahl, line 18: Could you remove these automatic comments please?

qixiang:
Hi @mprahl @apaplaus I refactored the code in this handler and pyxis module, and updated the patch, please help to review this.

The APIs are refactored because they return a overall list of images or NVRs without the mapping of digest to images, now we get the bundles which the images in advisory connect to first, then for each bundle, check auto rebuild tag and osbs pinning, finally for each bundle we're going to rebuild, update the digest in pullspec (`new`) to digest of rebuilt images. 

And other changes:
1. `get_images_by_digests` -> `get_images_by_digest`: 
- Instead of getting a overall list of images, we need to know the mapping of digest to images, so refactor it to call with a single digest.
- `manifest_list_digest` is used in bundle's related images, but for bundles, `manifest_schema2_digest` is used, so we need to get image by `manifest_list_digest` **or** `manifest_schema2_digest`.

2. Get brew extra metadata: `buildinfo.get("build", {}).get("extra", {})` -> `buildinfo.get("extra", {})`, there is no `build` key.
3. Get repository auto rebuild tags via endpoint `repositories/registry/{registry}/repository/{repository}` instead of `repositories`.

apaplaus (commented):
Except few small comments it seems fine

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -70,23 +71,59 @@ def handle(self, event):
...
+        # in pullspecs of bundle's related images
+        rebuilt_digests_by_nvr = {}
+        rebuilt_nvrs = nvrs_mapping.values()
+        for nvr in rebuilt_nvrs:
+            digest = self._pyxis.get_manifest_list_digest_by_nvr(nvr)
+            if digest:
+                rebuilt_digests_by_nvr[nvr] = digest
+            else:
+                log.warning(
+                    f"Image manifest_list_digest not found for {nvr} in Pyxis, skip this image"
```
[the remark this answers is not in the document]
qixiang: Updated.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -96,63 +133,144 @@ def handle(self, event):
...
+            buildinfo = koji_api.get_build(bundle_nvr)
+            related_images = (
+                buildinfo.get("extra", {})
+                .get("image", {})
+                .get("operator_manifests", {})
+                .get("related_images", {})
+            )
+            bundle_data['osbs_pinning'] = related_images.get('created_by_osbs', False)
+            # Save the original pullspecs
+            bundle_data['pullspecs'] = related_images.get('pullspecs', [])
```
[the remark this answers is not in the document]
qixiang: I don't think it can happen, every published bundle image should has pullspecs of related images, and even with that can happen, we'll skip this bundle in later for loop:
```
    for pullspec in bundle_data['pullspecs']:
    ...
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -70,23 +71,61 @@ def handle(self, event):
...
+                )
+
+        if not original_digests_by_nvr:
+            msg = f"None of the original images have digests in Pyxis: {','.join(original_nvrs)}"
+            log.warning(msg)
+            db_event.transition(EventState.SKIPPED, msg)
+            return []
+
+        # Convert to mapping of digest to NVR for later use
+        original_nvrs_by_digest = {v: k for k, v in original_digests_by_nvr.items()}
```
mprahl, line 111: Optional: You could just build this dictionary as part of the `for nvr in original_nvrs:` loop to avoid looping through the values again. It also makes it easier to read in my opinion.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -96,63 +135,144 @@ def handle(self, event):
...
-        bundles = self._pyxis.filter_bundles_by_related_image_digests(
-            original_images_digests, all_bundles)
+        # get bundle digests for original images
+        bundle_digests_by_related_nvr = {}
+        for image_nvr, image_digest in original_digests_by_nvr.items():
+            bundles = self._pyxis.get_bundles_by_related_image_digest(
+                image_digest, all_bundles
+            )
+            if not bundles:
+                log.info(f"No latest bundle image found for {image_nvr}")
```
mprahl, line 145:
```suggestion
                log.info(f"No latest bundle image with the related image of {image_nvr}")
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -96,63 +135,144 @@ def handle(self, event):
...
+            return []
+
+        # A set of unique bundle digests, an image can be in multiple bundles' related_images
         bundle_digests = set()
-        for bundle in bundles:
-            if not bundle.get('bundle_path_digest'):
-                log.warning("Bundle %s doesn't have 'bundle_path_digests' set",
-                            bundle['bundle_path'])
+        for digests in bundle_digests_by_related_nvr.values():
+            bundle_digests.update(digests)
```
mprahl, line 161: Optional: You could populate this set at the same time that you populate `bundle_digests_by_related_nvr`. It avoid an additional loop over this dictionary and it makes it easier to read.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -96,63 +135,144 @@ def handle(self, event):
...
+        bundles_by_digest = {}
+        default_bundle_data = {
+            'images': [],
+            'nvr': None,
+            'auto_rebuild': False,
+            'osbs_pinning': False,
+            'pullspecs': [],
+        }
+
+        # Get images for each bundle digest, a bundle digest can has multiple images
```
mprahl, line 182:
```suggestion
        # Get images for each bundle digest, a bundle digest can have multiple images
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -96,63 +135,144 @@ def handle(self, event):
...
+                old_pullspec = pullspec.get('new')
+                # Replace 'original' with the old 'new' pullspec
+                pullspec['original'] = old_pullspec
+
+                pullspec_elems = old_pullspec.split('@')
+                old_digest = pullspec_elems[1]
+
+                if old_digest not in original_nvrs_by_digest:
+                    # This related image is not one of the original images
+                    continue
```
mprahl, line 252: This check should happen before `pullspec['original'] = old_pullspec` is called since we want the original pullspec replacement that OSBS computed to remain for related images which weren't rebuilt.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -96,63 +135,144 @@ def handle(self, event):
...
-        bundles_nvrs = self._filter_bundles_by_pinned_related_images(
-            auto_rebuild_nvrs)
+                # If related image is not pinned by OSBS, skip
+                if not pullspec.get('pinned', False):
+                    continue
+
+                # A pullspec path is in format of "registry/repository@digest"
+                old_pullspec = pullspec.get('new')
+                # Replace 'original' with the old 'new' pullspec
+                pullspec['original'] = old_pullspec
```
mprahl, line 245: We don't want to modify this value since the original value remains in the bundle image metadata (e.g. the pullspec with the floating tag) and we want OSBS to be able to identify that original value and replace it with either the rebuilt image or what it was previously pinned to.

mprahl:
@qixiang this is looking really good. Just a few comments to address. Afterwards I'll review the code one more time and review the tests.

qixiang:
Thank you @mprahl , all comments have been addressed in updated PR.

mprahl (approved):

PR #46, opened 2021-01-19
Mark EUS base image as directly affected

apaplaus:
Now EUS base image will be marked as directly affected, despite it's unpublished, so botas could
file an advisory.

RESOLVE: CWFHEALTH-24

### DIFF
--- a/freshmaker/lightblue.py
+++ b/freshmaker/lightblue.py
@@ -486,6 +486,9 @@ class LightBlue(object):
 
         self.entity_versions = entity_versions or {}
 
+        # dict with mapping of EUS repositories to their 'auto_rebuild_tags'
+        self.repo_to_auto_rebuild_tags = {}
+
     def _get_entity_version(self, entity_name):
         """Lookup configured entity's version
 
@@ -1576,8 +1579,10 @@ class LightBlue(object):
         to_rebuild = self._deduplicate_images_to_rebuild(to_rebuild)
         # Get all the directly affected images so that any parents that are not marked as
         # directly affected can be set in _images_to_rebuild_to_batches
+        # And in addition mark all latest EUS images as directly affected so
+        # botas could correctly file an advisory
         directly_affected_nvrs = {
-            image.nvr for image in images if image.get("directly_affected")
+            image.nvr for image in images if image.get("directly_affected") or self.is_latest_eus_image(image)
         }
         # Some images that aren't marked as directly affected may have already been fixed
         # in the latest published version of the image. Use those images instead.
@@ -1825,3 +1830,49 @@ class LightBlue(object):
         image = images[0]
         image.resolve(self)
         return image
+
+    def is_latest_eus_image(self, image):
+        """
+        Check if the image is EUS base image and its repository is marked with
+        auto_rebuild tag. EUS repository contains '-els' suffix.
+
+        :param ContainerImage image: image to check
+        :param dict repo_to_auto_rebuild_tags: mapping of processed repositories
+            names to set of their 'auto_rebuild_tags'
+        :return: True if image is EUS base image with 'auto_rebuild_tag', False otherwise
+        """
+        def get_repo(repository):
+            query = {
+                "objectType": "containerRepository",
+                "query": {
+                    "$and": [
+                        {
+                            "field": "repository",
+                            "op": "=",
+                            "rvalue": repository
+                        },
+                    ]
+                },
+                "projection": [
+                    {"field": "auto_rebuild_tags", "include": True, "recursive": False}
+                ]
+            }
+            repos = self.find_container_repositories(query, auto_rebuild=True)
+            return repos[0]
+
+        for repo in image["repositories"]:
+            repo_name = repo["repository"]
+            # Check tags only if it's EUS image
+            if re.match(r"rhel\d+-\d+-els\/rhel", repo_name):
+                auto_rebuild_tags = self.repo_to_auto_rebuild_tags.get(repo_name)
+                if auto_rebuild_tags is None:
+                    auto_rebuild_tags = set(get_repo(repo_name)["auto_rebuild_tags"])
+                    self.repo_to_auto_rebuild_tags[repo_name] = auto_rebuild_tags
+
+                if not auto_rebuild_tags:
+                    continue
+                for tag in repo["tags"]:
+                    if tag["name"] in auto_rebuild_tags:
+                        return True
+
+        return False
--- a/tests/test_lightblue.py
+++ b/tests/test_lightblue.py
@@ -1562,6 +1562,7 @@ class TestQueryEntityFromLightBlue(helpers.FreshmakerTestCase):
         self.assertEqual(set(ret[2]["content_sets"]),
                          set(['dummy-content-set-1', 'dummy-content-set-2']))
 
+    @patch('freshmaker.lightblue.LightBlue.find_all_container_repositories')
     @patch('freshmaker.lightblue.LightBlue.find_images_with_packages_from_content_set')
     @patch('freshmaker.lightblue.LightBlue.find_parent_images_with_package')
     @patch('freshmaker.lightblue.LightBlue._filter_out_already_fixed_published_images')
@@ -1570,8 +1571,10 @@ class TestQueryEntityFromLightBlue(helpers.FreshmakerTestCase):
                                exists,
                                _filter_out_already_fixed_published_images,
                                find_parent_images_with_package,
-                               find_images_with_packages_from_content_set):
+                               find_images_with_packages_from_content_set,
+                               find_repos):
         exists.return_value = True
+        find_repos.return_value = {}
 
         image_a = ContainerImage.create({
             'brew': {'package': 'image-a', 'build': 'image-a-v-r1'},
@@ -1783,7 +1786,8 @@ class TestQueryEntityFromLightBlue(helpers.FreshmakerTestCase):
             "Couldn't find parent image some-original-nvr-7.6-252.1561619826. "
             "Lightblue data is probably incomplete"))
 
-    @patch("freshmaker.lightblue.LightBlue.get_images_by_nvrs")
+    @patch('freshmaker.lightblue.LightBlue.find_all_container_repositories')
+    @patch('freshmaker.lightblue.LightBlue.get_images_by_nvrs')
     @patch('freshmaker.lightblue.LightBlue.find_images_with_packages_from_content_set')
     @patch('freshmaker.lightblue.LightBlue.find_parent_images_with_package')
     @patch('freshmaker.lightblue.LightBlue._filter_out_already_fixed_published_images')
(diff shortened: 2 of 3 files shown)


freshmaker/lightblue.py
```diff
@@ -1576,8 +1576,10 @@ def _get_images_to_rebuild(image):
         to_rebuild = self._deduplicate_images_to_rebuild(to_rebuild)
         # Get all the directly affected images so that any parents that are not marked as
         # directly affected can be set in _images_to_rebuild_to_batches
+        # And in addition mark all EUS images as directly affected so botas could
+        # correctly file an advisory
         directly_affected_nvrs = {
-            image.nvr for image in images if image.get("directly_affected")
+            image.nvr for image in images if image.get("directly_affected") or is_eus_image(image)
```
mprahl, line 1582: You need to ensure that the EUS image being rebuilt is the latest image tagged with an autorebuild tag.

For example, if I have child image A built on `rhel-els-container-8.2-4` but the latest is actually `rhel-els-container-8.2-5`, then `rhel-els-container-8.2-4` shouldn't be marked as directly affected.

mprahl:
Additionally, it'd be nice if `_filter_out_already_fixed_published_images` was adjusted to handle the case of ELS images. Right now it only works with published images.

mprahl:
Optional: It'd be nice if this was made more generic so that it was a configuration option (regex strings perhaps) of the repositories that have unpublished images that must be handled specially rather than making the logic specific to ELS images.

freshmaker/lightblue.py
apaplaus: So when we have some EUS image for example `rhel-els-container-8.2-4` that is a parent of the image attached to the advisory, we should check if it's the latest image, and if it's not, rebuild latests image instead of actual parent? Does this make sense?

apaplaus:
> Additionally, it'd be nice if `_filter_out_already_fixed_published_images` was adjusted to handle the case of ELS images. Right now it only works with published images.

But this method ignores 'directly_affected' images. So if EUS images will be marked as such, they will be ignored. If I understood its logic right.

apaplaus:
> Optional: It'd be nice if this was made more generic so that it was a configuration option (regex strings perhaps) of the repositories that have unpublished images that must be handled specially rather than making the logic specific to ELS images.

But for now we should have this special logic only for EUS images. I think when there will be other requests, we could think of something like filters for those cases.

apaplaus:
@mprahl I've made a little change, but not sure if it's enough just to check for 'latest' tag in image tags. What do you think, should I do it in some other way?

mprahl:
@apaplaus, you are on the right track. The idea here is if you see an EUS base image, check what the auto-rebuild tags are on the repository for that EUS base image. If the EUS base image is tagged with one of those auto-rebuild tags, then it should be marked as directly affected.

freshmaker/lightblue.py
qixiang: Probably you should update `_filter_out_already_fixed_published_images` to also check the latest EUS images, if the latest EUS image already has the CVE fix addressed, the latest EUS image should be used as parent, and in such cases, we don't need to mark the EUS image as affected.

freshmaker/utils.py
```diff
@@ -217,3 +218,21 @@ def get_ocp_release_date(ocp_version):
...
+def is_latest_eus_image(image):
+    """
+    Check if the image is EUS base image marked with 'latest' tag by checking it
+     repositories. EUS repository contains '-els' suffix.
+
+    :param ContainerImage image: image to check
+    :return: True if image is latest EUS base image, False otherwise
+    """
+    for repo in image["repositories"]:
+        if re.match(r".*-els($|\W)", repo["repository"]):
```
qixiang, line 232: By checking the repositories in Pyxis, I see there are 2 els repositories:
- rhel8-2-els/rhel
- rhel8-4-els/rhel
So you need to fix the regex pattern.

apaplaus: @qixiang  Do you mean I should match only those two repos? But what if in the future we will have other repos and not only for RHEL? And if I need to match only those two repos, why can't we just have a list in that case?

qixiang: The regex pattern `r".*-els($|\W)"` doesn't match the two repositories which have `/rhel` in repository name.

apaplaus: But if I execute `re.match(r".*-els($|\W)", "rhel8-2-els/rhel")` it matches, or do you mean something else?

qixiang: oops, you're right, I was checking with Pyxis's "repositories" endpoint with filter without quoting the regex pattern, sorry about that.

apaplaus:
@mprahl So now I am checking all the auto_rebuild tagged repos, is it fine now?

freshmaker/lightblue.py
```diff
@@ -1576,8 +1576,12 @@ def _get_images_to_rebuild(image):
         to_rebuild = self._deduplicate_images_to_rebuild(to_rebuild)
         # Get all the directly affected images so that any parents that are not marked as
         # directly affected can be set in _images_to_rebuild_to_batches
+        # And in addition mark all latest EUS images as directly affected so
+        # botas could correctly file an advisory
+        repos = self.find_all_container_repositories(published, release_categories)
```
mprahl, line 1581: `published` would need to be `False` to find the EUS image repositories. That being said, this is essentially duplicating the query in `find_images_with_packages_from_content_set` and it is quite expensive.

The `image` dictionary will have the `repositories` key, so you can search for the auto-rebuild tags of those repositories and see if the EUS image currently has that tag. I would also convert `is_latest_eus_image` to a method on this class and then have it cache the autorebuilds tags for the EUS repositories on the object.

apaplaus: So if I got you right, you suggest to make a request to LB for every repo in `repositories' of every image and then get `auto_rebuild_tags` from the response. And check if any of this tags corresponds to the tags that are in image's `repositories' and if so mark this image as **directly_affected**?

mprahl: Yes, but when you say "every image", it would be every EUS image.

apaplaus:
@mprahl so here is a new batch of changes

freshmaker/lightblue.py
```diff
@@ -1825,3 +1830,50 @@ def get_fixed_published_image(self, name, version, image_group, rpm_nvrs, conten
         image = images[0]
         image.resolve(self)
         return image
+
+    def is_latest_eus_image(self, image):
+        """
+        Check if the image is EUS base image and his repository marked with
```
mprahl, line 1836:
```suggestion
        Check if the image is EUS base image and its repository is marked with
```

freshmaker/lightblue.py
```diff
@@ -1825,3 +1830,50 @@ def get_fixed_published_image(self, name, version, image_group, rpm_nvrs, conten
...
+                    {"field": "auto_rebuild_tags", "include": True, "recursive": True}
+                ]
+            }
+            repos = self.find_container_repositories(query, auto_rebuild=True)
+            return repos[0]
+
+        for repo in image["repositories"]:
+            repo_name = repo["repository"]
+            # Check tags only if it's EUS image
+            if re.match(r".*-els($|\W)", repo_name):
```
mprahl, line 1866: How about a more precise reged?
```suggestion
            if re.match(r"rhel\d+-\d+-els\/rhel", repo_name):
```

freshmaker/lightblue.py
```diff
@@ -1825,3 +1830,50 @@ def get_fixed_published_image(self, name, version, image_group, rpm_nvrs, conten
...
+                auto_rebuild_tags = self.repo_to_auto_rebuild_tags.get(repo_name, None)
```
mprahl, line 1867: Optional: `None` is not needed. It's the default.
```suggestion
                auto_rebuild_tags = self.repo_to_auto_rebuild_tags.get(repo_name)
```

freshmaker/lightblue.py
```diff
@@ -1825,3 +1830,50 @@ def get_fixed_published_image(self, name, version, image_group, rpm_nvrs, conten
...
+                    "$and": [
+                        {
+                            "field": "repository",
+                            "op": "=",
+                            "rvalue": repository
+                        },
+                    ]
+                },
+                "projection": [
+                    {"field": "auto_rebuild_tags", "include": True, "recursive": True}
```
mprahl, line 1857: I don't believe `recursive` is required here.

freshmaker/lightblue.py
```diff
@@ -1825,3 +1830,50 @@ def get_fixed_published_image(self, name, version, image_group, rpm_nvrs, conten
...
+                auto_rebuild_tags = self.repo_to_auto_rebuild_tags.get(repo_name, None)
+                if auto_rebuild_tags is None:
+                    auto_rebuild_tags = get_repo(repo_name)["auto_rebuild_tags"]
+                    self.repo_to_auto_rebuild_tags[repo_name] = auto_rebuild_tags
+
+                if not auto_rebuild_tags:
+                    continue
+                repo_tags = (tag["name"] for tag in repo["tags"])
+                for tag in repo_tags:
+                    if tag in auto_rebuild_tags:
```
mprahl, line 1876: Optional: You can take this syntax if you prefer
```suggestion
                repo_tags = {tag["name"] for tag in repo["tags"]}
                if repo_tags & auto_rebuild_tags:
```

mprahl:
@apaplaus this looks really good. I just left a few comments.

freshmaker/lightblue.py
apaplaus: But if we will have not only RHEL EUS images? And in design doc there is this example of repository: `rh-osbs/rhel-els`. It will not match the regex.

freshmaker/lightblue.py
apaplaus: Wont it consume more time? Because in my option it will return when encounter a first tag. And in your option it needs to make intersection of two sets.

freshmaker/lightblue.py
mprahl: @apaplaus so it's only the RHEL EUS image that isn't published and needs rebuilding as far as I know. The `rh-osbs/rhel-els` repo is the "build repo", which is the repo that is used by OSBS to store built images in quay.io. Any image that is released then gets copied from the build repo to the Red Hat Container Registry. This means we specifically want to ignore the `rh-osbs` namespace since autorebuild tags will never be set on that.

freshmaker/lightblue.py
mprahl: You are right. I thought it was more readable what I suggested, but you can keep it as is. I also forgot that `auto_rebuild_tags` is not a set.

If you change `auto_rebuild_tags` to be a set, the most efficient way would be the following:
```suggestion
                for tag in repo["tags"]:
                    if tag["name"] in auto_rebuild_tags:
```

This avoids essentially looping through `repo["tags"]` and makes the `in` operation O(1) instead of O(n).

apaplaus: Done

freshmaker/lightblue.py
apaplaus: I've changed it

Issue #48, opened 2021-01-22
fix quay image build


ktdreyer:
The quay badge says the container has failed to build.

qixiang:
There are 2 reasons where the builds were failed:
1. We pull base image from docker registry and it reached to pull rate limit, probably we can pull base image (fedora) from quay.io instead.
2. Quay.io build trigger doesn't support build image on a particular branch, it has default branch set to "ALL", and we have a documentation branch "gh-pages" which should not perform image build against it.

ktdreyer:
Would you please re-open this? It's still failed

qixiang:
I was closed automatically. We fixed the issue of pulling image from docker hub by migrating to the base image in quay.io, so image can be built without issue now. However there is another issue, github doesn't support per-branch webhooks and quay.io doesn't support build trigger on a particular branch, so quay.io always try to build image on our documentation branch and always fails. I can keep it open at this moment.

qixiang:
Build trigger has been changed to github push trigger which supports specifying the branch to build image from, the quay badge now works, closing.

PR #51, opened 2021-01-25
Use fedora base image from quay.io

qixiang:
Pulling fedora base image from docker registry can reach to pull rate
limit, and image build job will fail due to that, change base image to
fedora from quay.io.

Resolves: #48


apaplaus (approved):

apaplaus:
Small comment about failing CI, is it ok?

qixiang:
CI fixed in #52

PR #53, opened 2021-01-25
[WIP] Add handler for EUS base images

apaplaus:
Since EUS base images are unpublished there should be handler that will
take care of those images that don't have children.

RESOLVE: CWFHEALTH-26


apaplaus:
@mprahl Could you take a look please? It's only part of the code without tests and probably some other things, but I wonder if I am doing everything fine?

freshmaker/handlers/koji/rebuild_eus_base_images.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+
+    name = 'RebuildEUSBaseImages'
+
+    def can_handle(self, event):
+        if not super().can_handle(event):
+            return False
+
+        # If 'content_sets' of the event doesn't exist yet, create and fill it
+        if not hasattr(self.event, 'content_sets'):
```
mprahl, line 51: When would `self.event.content_sets` ever be set since it is not stored in the database and the base class' `can_handle` method does not set it.

freshmaker/handlers/koji/rebuild_eus_base_images.py
```diff
@@ -0,0 +1,125 @@
...
+            errata = Errata()
+            errata_id = int(event.search_key)
+            # Use the errata_id to find out Pulp repository IDs from Errata Tool
+            # and furthermore get content_sets from Pulp where signed RPM will end
+            # up eventually when advisories are shipped.
+            pulp_repo_ids = list(set(errata.get_pulp_repository_ids(errata_id)))
```
mprahl, line 57: You can't call `_set_event_content_sets` instead of duplicating this code?

freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
```diff
@@ -423,8 +415,28 @@ def _find_images_to_rebuild(self, errata_id):
...
+        if not hasattr(self.event, 'content_sets'):
+            # Use the errata_id to find out Pulp repository IDs from Errata Tool
+            # and furthermore get content_sets from Pulp where signed RPM will end
+            # up eventually when advisories are shipped.
+            pulp_repo_ids = list(set(errata.get_pulp_repository_ids(errata_id)))
+
+            pulp = Pulp(server_url=conf.pulp_server_url,
+                        username=conf.pulp_username,
+                        password=conf.pulp_password)
+            self.event.content_sets = pulp.get_content_set_by_repo_ids(
```
mprahl, line 441: What is the benefit of setting it on the event?

mprahl:
One concern is that the `handle` method will skip the event if there are no builds, so that means the event will be marked as skipped if this new handler does not perform any builds. This is an issue if the `RebuildImagesOnRPMAdvisoryChange` handler actually performs a build.

mprahl:
@apaplaus could you also look to see what would happen if this handler performed some builds and the `RebuildImagesOnRPMAdvisoryChange` also performed some builds on the same event?

freshmaker/handlers/koji/rebuild_eus_base_images.py
apaplaus: It could be set in other handler during handling same event, for example in `_find_images_to_rebuild` method in `RebuildImagesOnRPMAdvisoryChange` handler, so to not request content_sets several times I've added this.

freshmaker/handlers/koji/rebuild_eus_base_images.py
apaplaus: Yeah you are right, probably forgot to replace it =)

freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
apaplaus: It's just as I've mentioned above, if one event will be handled by more handlers, it makes no sense to request same content_sets again. So just to make it more efficient I've decided to use Event to store this information.

apaplaus:
@mprahl so your last two comments were about that problem that you are discussing with @qixiang in mail. That handlers think that they 'own' the event. So I think I should make some changes to the code when we will decide what to do with this.

freshmaker/handlers/koji/rebuild_eus_base_images.py
mprahl: You are right. I thought `self.event` was the database event, but it is the "message" event. In that case, could you please set `self.content_sets` to `None` in the `__init__` method of `ErrataBaseEvent`? That way, this check can be `if self.event.content_sets is None` instead.

apaplaus:
@mprahl so your last two comments were about that problem that you are discussing with @qixiang in mail. That handlers think that they 'own' the event. So I think I should make some changes to the code when we will decide what to do with this.

freshmaker/handlers/koji/rebuild_eus_base_images.py
apaplaus: Sure

apaplaus:
@mprahl Please take a look.

freshmaker/handlers/koji/rebuild_eus_base_images.py
```diff
@@ -0,0 +1,107 @@
...
+            return False
+
+        # If 'content_sets' of the event doesn't exist yet, create and fill it
+        if self.event.content_sets is None:
+            errata = Errata()
+            self._set_event_content_sets(errata, int(event.search_key))
+
+        # Check if any of the content sets is EUS content set
+        for content_set in self.event.content_sets:
+            if re.match(r'.*-eus($|\W)', content_set):
```
mprahl, line 49:
```suggestion
            if re.match(r'.+-eus-rpms', content_set):
```

freshmaker/handlers/koji/rebuild_eus_base_images.py
```diff
@@ -0,0 +1,107 @@
...
+                return True
+
+        return False
```
mprahl, line 52: Optional: Just an alternative syntax that is shorter if you prefer this. You can disregard it if you prefer your syntax.
```suggestion
        return any(re.match(r'.+-eus-rpms', c) for c in self.event.content_sets)
```

freshmaker/handlers/koji/rebuild_eus_base_images.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def _find_images_to_rebuild(self, errata_id):
+        """
+        Finds docker rebuild images from each build added to specific Errata
+        advisory.
+
+        But only EUS base images that DON'T have children should be rebuilt.
+        So other bathes including EUS images with children will be handled by
```
mprahl, line 60:
```suggestion
        So other batches including EUS images with children will be handled by
```

freshmaker/handlers/koji/rebuild_eus_base_images.py
```diff
@@ -0,0 +1,107 @@
...
+        if not affected_nvrs:
+            affected_nvrs = errata.get_binary_rpm_nvrs(errata_id)
+
+        self.log_info(
+            "Going to find all the container images to rebuild as "
+            "result of %r update.", affected_nvrs)
+        batches = lb.find_images_to_rebuild(
+            affected_nvrs, self.event.content_sets,
+            filter_fnc=self._filter_out_not_allowed_builds,
+            published=None, release_categories=None,
```
mprahl, line 99: `published=False` should be set explicitly for this filter:
https://github.com/redhat-exd-rebuilds/freshmaker/blob/2f3bf7d8cf497ea9a216b391067d0ba4672f82f0/freshmaker/lightblue.py#L949-L956

I don't think it really matters, but it's probably best to be consistent here.

apaplaus:
Work scope was changed, there is no need in new handler anymore.

PR #54, opened 2021-02-01
Always set pinned to true when Freshmaker replaces a pullspec

mprahl:
This indicates that the pullspec was replaced by Freshmaker in
the bundle image.

This is a followup to #45.

### DIFF
--- a/freshmaker/handlers/botas/botas_shipped_advisory.py
+++ b/freshmaker/handlers/botas/botas_shipped_advisory.py
@@ -233,10 +233,6 @@ class HandleBotasAdvisory(ContainerBuildHandler):
                 #   'pinned': True
                 # }
 
-                # If related image is not pinned by OSBS, skip
-                if not pullspec.get('pinned', False):
-                    continue
-
                 # A pullspec path is in format of "registry/repository@digest"
                 pullspec_elems = pullspec.get('new').split('@')
                 old_digest = pullspec_elems[1]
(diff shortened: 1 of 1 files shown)


apaplaus (approved):

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -254,6 +254,10 @@ def handle(self, event):
                 pullspec_elems[1] = new_digest
                 new_pullspec = '@'.join(pullspec_elems)
                 pullspec['new'] = new_pullspec
+                # Always set pinned to True when it was replaced by Freshmaker
+                # since it indicates that the pullspec was modified from the
+                # original pullspec
+                pullspec['pinned'] = True
```
qixiang, line 260: We only replace the pullspec digest when `pinned` is True (https://github.com/redhat-exd-rebuilds/freshmaker/blob/4c0f1cab0e01d691665ce1841aa5738e74b47cfc/freshmaker/handlers/botas/botas_shipped_advisory.py#L237), so this is unnecessary.

mprahl:
@qixiang thanks for pointing that out. Okay, it's ready for review again.

qixiang:
> @qixiang thanks for pointing that out. Okay, it's ready for review again.

Sorry for making confusion, I mean we only replace the digest when `pinned` is true, when `pinned` is false, we just skip that pullspec, so when we do the replacement, the `pinned` is already `True` and it's unnecessary to set it `True` again.

mprahl:
> > @qixiang thanks for pointing that out. Okay, it's ready for review again.
> 
> Sorry for making confusion, I mean we only replace the digest when `pinned` is true, when `pinned` is false, we just skip that pullspec, so when we do the replacement, the `pinned` is already `True` and it's unnecessary to set it `True` again.

@qixiang I understand, but we want to fix CVEs even if the image was prepinned in the bundle metadata. This will be required to support ACM rebuilds. We just want to skip bundle images which completely bypassed OSBS metadata adjustments.

qixiang:
> @qixiang I understand, but we want to fix CVEs even if the image was prepinned in the bundle metadata. This will be required to support ACM rebuilds. We just want to skip bundle images which completely bypassed OSBS metadata adjustments.

ok, I see, then this looks good to me, free feel to merge.

PR #55, opened 2021-02-04
Eus images support

apaplaus:
Reverting commit that make EUS images marked as **directly_affected** + modify Lightblue module to query unpublished repositories specified in configuration file.

### DIFF
--- a/conf/config.py
+++ b/conf/config.py
@@ -204,6 +204,9 @@ class BaseConfiguration(object):
         }
     }
 
+    # repositories that should be searched for unpublished images, specifically because of EUS base images
+    UNPUBLISHED_EXCEPTIONS = []
+
 
 class DevConfiguration(BaseConfiguration):
     DEBUG = True
--- a/freshmaker/config.py
+++ b/freshmaker/config.py
@@ -414,7 +414,13 @@ class Config(object):
             'type': str,
             'default': '',
             'desc': 'The API URL of the Product Pages service'
-        }
+        },
+        'unpublished_exceptions': {
+            'type': list,
+            'default': [],
+            'desc': 'List of dictionaries with unpublished repos, containing '
+                    '"registry" and "repository" keys that should not be ignored '
+                    'when searching for images to rebuild.'}
     }
 
     def __init__(self, conf_section_obj):
@@ -562,3 +568,14 @@ class Config(object):
         ccache_file = ccache_file.replace(
             "$pid", str(os.getpid()))
         return ccache_file
+
+    def _setifok_unpublished_exceptions(self, exceptions):
+        for exception in exceptions:
+            if not exception.get("registry", ""):
+                raise ValueError("There is no 'registry' or it's empty in one"
+                                 " of the UNPUBLISHED_EXCEPTIONS")
+
+            if not exception.get("repository", ""):
+                raise ValueError("There is no 'repository' or it's empty in one"
+                                 " of the UNPUBLISHED_EXCEPTIONS")
+        self._unpublished_exceptions = exceptions
--- a/freshmaker/lightblue.py
+++ b/freshmaker/lightblue.py
@@ -643,14 +643,27 @@ class LightBlue(object):
         :type vendors: tuple[str]
         """
         if published is not None:
-            request["query"]["$and"].append({
+            request["query"]["$or"][0]["$and"].append({
                 "field": "published",
                 "op": "=",
                 "rvalue": published
             })
+            # If the query is for published images, add configurable repos  for
+            # unpublished images(like EUS) too because they shouldn't be ignored
+            if published is True and conf.unpublished_exceptions:
+                for repo in conf.unpublished_exceptions:
+                    request["query"]["$or"].append(
+                        {
+                            "$and": [
+                                {"field": "published", "op": "=", "rvalue": False},
+                                {"field": "registry", "op": "=", "rvalue": repo["registry"]},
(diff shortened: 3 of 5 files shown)


conf/config.py
```diff
@@ -204,6 +204,12 @@ class BaseConfiguration(object):
         }
     }
 
+    # repositories that should be searched for unpublished images, specifically because of EUS base images
+    UNPUBLISHED_EXCEPTIONS = [
+        {"registry": "registry.access.redhat.com", "repository": "rhel8-2-els/rhel"},
```
mprahl, line 209: Could you please make this default to an empty list but configure this in our internal configuration repo of `freshmaker-playbooks`? This is because this is specific to the internal deployment of Freshmaker and I don't want to update this value upstream and perform a deployment just to update this configuration value.

freshmaker/config.py
```diff
@@ -414,7 +414,13 @@ class Config(object):
             'type': str,
             'default': '',
             'desc': 'The API URL of the Product Pages service'
-        }
+        },
+        'unpublished_exceptions': {
```
mprahl, line 418: Could you please add a method called `_setifok_unpublished_exceptions` that validates that the registry and repository keys are both set for any dictionary in this list?

freshmaker/lightblue.py
```diff
@@ -873,6 +893,67 @@ def filter_out_images_based_on_content_set(self, images, content_sets):
...
+                {
+                    "field": "repositories.*.published",
+                    "op": "=",
+                    "rvalue": published
+                })
+            # If the query is for published images, add configurable repos  for
+            # unpublished images(like EUS) too because they shouldn't be ignored
+            if published is True and conf.unpublished_exceptions:
+                for repo in conf.unpublished_exceptions:
+                    request["query"]["$and"][1]["$or"].append(
```
mprahl, line 939: You'll also need the other filters that were applied to `request["query"]["$and"][1]["$or"][0]["$and"]`. My suggestion is that you move the `"repositories.*.published"` filter from above to below the `if published is True and conf.unpublished_exceptions:` code block. Then for each repo in `conf.unpublished_exceptions`, you would first deep copy `request["query"]["$and"][1]["$or"][0]["$and"]` and then you would also add the filters for `unpublished_exceptions` to the deep copy. Then you would append the result to `request["query"]["$and"][1]["$or"]`.

conf/config.py
apaplaus: https://github.com/redhat-exd-rebuilds/freshmaker/pull/55 Here is the PR for this.

apaplaus: Here is the MR for this change: https://gitlab.cee.redhat.com/ansible-playbooks/freshmaker-playbooks/-/merge_requests/84

freshmaker/lightblue.py
apaplaus: Are you sure that it will work? Because as I understood if I do as you suggested, I will get at the end wrong filters like:
```
request["query"]["$and"][1]["$or"]:
   [
      {"$or": [....some_rpm_manifests...]},
      {"$or": [....some_content_sets...]},
      {"$and": [...unpublished_repos...]}
   ]
```
but I think I might need to do something like:
```
request["query"]["$and"][1]["$or"]:
   [
      {"$and":
         [
           {"published": False},
           {"registry": "...registry_from_unpublished_exceptions..."}
           {"repository": "...repository_from_unpublished_exceptions..."}
           {"$or": [....some_rpm_manifests...]},
           {"$or": [....some_content_sets...]}
         ]

      },
      {"$and":
         [
           {"published": False},
           {"registry": "...registry#2_from_unpublished_exceptions..."}
           {"repository": "...repository#2_from_unpublished_exceptions..."}
           {"$or": [....some_rpm_manifests...]},
           {"$or": [....some_content_sets...]}
         ]
      },
     ...
 ]
```

apaplaus: So I've made a change in the code as I described, you can check if it's ok.

mprahl (approved):
Looks great. Could you remove `[WIP]` from the commit please?

@qixiang could you also take a quick look please?

qixiang:
Hi @apaplaus , the multi-level nested image query filter is a little hard to maintain and not good for code readability. Can we refactor it to some like this:
```
repo_filters = []
if published is not None:
    repo_filters.append({"field": "repositories.*.published", "op": "=", "rvalue": published})

    if published and conf.unpublished_exceptions:
        for repo in conf.unpublished_exceptions:
            repo_filters.append(
                {
                    "$and": [
                        {"field": "repositories.*.published", "op": "=", "rvalue": False},
                        {"field": "repositories.*.registry", "op": "=", "rvalue": repo["registry"]},
                        {"field": "repositories.*.repository", "op": "=", "rvalue": repo["repository"]},
                    ]
                }
            )

query = {"$and": []}
if len(repo_filters) == 1:
    query["$and"].append(repo_filters[0])
if len(repo_filters) > 1:
    query["$and"].append({"$or": [r for r in repo_filters]})

if auto_rebuild_tags:
    query["$and"].append({"field": "repositories.*.tags.*.name", "op": "$in", "values": auto_rebuild_tags})

if content_sets:
    query["$and"].append({"field": "content_sets.*", "op": "$in", "values": content_sets})

if rpm_names:
    query["$and"].append({"field": "rpm_manifest.*.rpms.*.name", "op": "$in", "values": rpm_names})

image_request = {
    "objectType": "containerImage",
    "query": query,
    "projection": projection,
}
```

qixiang:
Good job, :+1:

PR #56, opened 2021-02-04
`get_latest_bundle` now uses `latest_in_channel` field

Lukkra211:
@apaplaus PTAL

For now it's just a WIP since I have to fix the tests yet - had a bit of a problem to run them since I don't have rpm based distro.

### DIFF
--- a/tests/test_pyxis.py
+++ b/tests/test_pyxis.py
@@ -405,55 +405,28 @@ class TestQueryPyxis(helpers.FreshmakerTestCase):
         assert "4.8" not in [i["ocp_version"] for i in indices]
 
     @patch('freshmaker.pyxis.Pyxis._pagination')
-    def test_get_bundles_per_index_image(self, page):
+    def test_get_latest_bundles(self, page):
         page_copy = self.copy_call_args(page)
-        page_copy.side_effect = [self.bundles[0:2], self.bundles[2:]]
-        out = self.px._get_bundles_per_index_image(self.indices)
-
-        expected_out = {"path/to/registry:v4.5": self.bundles[0:2],
-                        "path/to/registry:v4.6": self.bundles[2:]}
-
-        self.assertEqual(out, expected_out)
-        page.assert_has_calls([
-            call('operators/bundles',
-                 {'include': 'data.channel_name,data.version,'
-                             'data.related_images,data.bundle_path_digest,'
-                             'data.bundle_path', 'filter':
-                     'source_index_container_path==path/to/registry:v4.6'}),
-            call('operators/bundles',
-                 {'include': 'data.channel_name,data.version,'
-                             'data.related_images,data.bundle_path_digest,'
-                             'data.bundle_path', 'filter':
-                     'source_index_container_path==path/to/registry:v4.6'})
-        ])
-
-    @patch('freshmaker.pyxis.Pyxis._get_bundles_per_index_image')
-    def test_get_latest_bundles(self, get_bundles_per_indices):
-        get_bundles_per_indices.return_value = {
-            "path/to/registry:v4.5": self.bundles[0:2],
-            "path/to/registry:v4.6": self.bundles[2:]
-        }
+        page_copy.side_effect = [[self.bundles[:3]], []]
 
         out = self.px.get_latest_bundles(self.indices)
-        # we expect 0 and 3 bundles to be filtered out, because of older versions
-        expected_out = self.bundles[1:3] + self.bundles[4:]
+        expected_out = [self.bundles[:3]]
+
         self.assertEqual(out, expected_out)
-        get_bundles_per_indices.assert_called_once()
-
-    @patch('freshmaker.pyxis.Pyxis._get_bundles_per_index_image')
-    def test_get_latest_bundles_invalid_version(self, get_bundles_per_indices):
-        # set invalid version
-        for bundle in self.bundles:
-            bundle['version'] = "InvalidVersion"
-        get_bundles_per_indices.return_value = {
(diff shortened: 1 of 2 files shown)


tests/test_pyxis.py
```diff
@@ -405,56 +405,30 @@ def test_get_operator_indices_with_unreleased_filtered_out(self, page):
...
-        out = self.px._get_bundles_per_index_image(self.indices)
+        page_copy.side_effect = [[self.bundles[2]], []]
 
-        expected_out = {"path/to/registry:v4.5": self.bundles[0:2],
-                        "path/to/registry:v4.6": self.bundles[2:]}
+        out = self.px.get_latest_bundles(self.indices)
+        expected_out = [self.bundles[2]]
 
         self.assertEqual(out, expected_out)
         page.assert_has_calls([
```
apaplaus, line 417: There was a mistake in original code, so it should call `assert_has_call` method on `page_copy` not `page`. So then you might want to adjust call object below too.

tests/test_pyxis.py
```diff
@@ -405,56 +405,30 @@ def test_get_operator_indices_with_unreleased_filtered_out(self, page):
         assert "4.8" not in [i["ocp_version"] for i in indices]
 
     @patch('freshmaker.pyxis.Pyxis._pagination')
-    def test_get_bundles_per_index_image(self, page):
+    def test_get_latest_bundles(self, page):
+        self.maxDiff = None
         page_copy = self.copy_call_args(page)
-        page_copy.side_effect = [self.bundles[0:2], self.bundles[2:]]
-        out = self.px._get_bundles_per_index_image(self.indices)
+        page_copy.side_effect = [[self.bundles[2]], []]
```
apaplaus, line 411: I would expect to have more bundles in output when calling `get_latest_bundles` for indices with different ocp_versions, so It's better to have at least two bundles in output of two calls of `pagination`.

tests/test_pyxis.py
```diff
@@ -405,56 +405,30 @@ def test_get_operator_indices_with_unreleased_filtered_out(self, page):
         assert "4.8" not in [i["ocp_version"] for i in indices]
 
     @patch('freshmaker.pyxis.Pyxis._pagination')
-    def test_get_bundles_per_index_image(self, page):
+    def test_get_latest_bundles(self, page):
+        self.maxDiff = None
```
apaplaus, line 409: I don't think you need it after you done with debugging.

Lukkra211: Removed.

tests/test_pyxis.py
Lukkra211: Should be ok now.

tests/test_pyxis.py
Lukkra211: Fixed.

tests/test_pyxis.py
apaplaus: No, I meant that you are making **2** requests to pyxis with `pagination`, because there are **2** ocp versions, so I am as a requester would expect after two calls end up with exactly two(or more?) latest bundle images, but since you have ` page_copy.side_effect = [[self.bundles[...]], []]` I will get one response with one(few?) bundles and seconds with none, right? I think it's not how it should work in reality.
 Second think is as you've made this change:
```
bundle_list = self._pagination('operators/bundles', request_params)
if bundle_list:
   latest_bundles.extend(bundle_list)
```
I have a question. Is it possible that there will be more than one **latest bundle per channel per index image**. I don't know, but probably no. But if there is more than one, in addition to existing code I would log some warning.

Lukkra211: As for the return values, sure, I can change it so each call returns something, but to fully test the function I still need a call which returns en empty list. It's not a common response but it's a possible one, so I thought that it should be tested as well.

Yes, in the index image, there are multiple channels which causes that there might be multiple bundles returned. Example of such query is `latest_in_channel==true and source_index_container_path==registry.stage.redhat.io/redhat/redhat-operator-index:v4.6`. I get 7 bundles. That's why I made that change.

Lukkra211: It should be a common thing to have more such bundles. For example
```
$ podman run -d -p 50051:50051 registry.redhat.io/redhat/redhat-operator-index:v4.6
$ grpcurl -plaintext -format json localhost:50051 describe api.Registry
$ grpcurl -plaintext -format json localhost:50051 api.Registry.ListPackages
$ grpcurl -plaintext -format json localhost:50051 describe api.Registry.GetPackage
api.Registry.GetPackage is a method:
rpc GetPackage ( .api.GetPackageRequest ) returns ( .api.Package );

$ grpcurl -plaintext -format json localhost:50051 describe .api.Package
api.Package is a message:
message Package {
  string name = 1;
  repeated .api.Channel channels = 2;
  string defaultChannelName = 3;
}

$ grpcurl -plaintext -format json -d '{"name": "servicemeshoperator"}' localhost:50051 api.Registry.GetPackage
{
  "name": "servicemeshoperator",
  "channels": [
    {
      "name": "1.0",
      "csvName": "servicemeshoperator.v2.0.1.1"
    },
    {
      "name": "stable",
      "csvName": "servicemeshoperator.v2.0.1.1"
    }
  ],
  "defaultChannelName": "stable"
}
```
So in the index image I listed the packages and choose one. You can see it has two channels + there are almost always more packages in one index image.

I think we shouldn't log more bundles were returned.

mprahl (approved):

apaplaus (approved):

apaplaus:
Maybe it would be better to squash all those commits into one, to have a cleaner history?

PR #57, opened 2021-02-12
Freshmaker continue building images after exception

apaplaus:
Now Freshmaker will continue building images even when some exception is
raised during one of the builds. Previously other builds were skipped
and build event was stuck in 'BUILDING' state.

JIRA: CWFHEALTH-35

### DIFF
--- a/freshmaker/handlers/__init__.py
+++ b/freshmaker/handlers/__init__.py
@@ -196,6 +196,12 @@ class BaseHandler(object):
         """
         return self._log(log.error, msg, *args, **kwargs)
 
+    def log_except(self, msg, *args, **kwargs):
+        """
+        Wraps log.exception, prefixes the message with a context of this handler.
+        """
+        return self._log(log.exception, msg, *args, **kwargs)
+
     def force_dry_run(self):
         """
         Forces the handling of the current even in DRY_RUN mode.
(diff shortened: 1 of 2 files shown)


freshmaker/handlers/__init__.py
```diff
@@ -611,6 +611,9 @@ def build_image(build):
                 # We skip this image for now. It will be built once the ODCS
                 # compose is finished.
                 return
+            except Exception as ex:
+                self.log_warn("While processing the event with id {}, exception"
```
mprahl, line 615: Could you make this use `log_exception` so that we get the whole traceback?

freshmaker/handlers/__init__.py
```diff
@@ -611,6 +611,9 @@ def build_image(build):
                 # We skip this image for now. It will be built once the ODCS
                 # compose is finished.
                 return
+            except Exception as ex:
```
mprahl, line 614: I'm thinking you should fail the build here with a state reason of `An unknown error occurred`. If you don't, then it'll say `Error while building container image in Koji.` which is misleading since the error may have happened outside of Koji.

freshmaker/handlers/__init__.py
apaplaus: Are you sure that it's ok to use flask's `log_exception`? I think it exists for more internal things, and there should be an active HTTP request. Maybe it's better to just use `traceback.print_exc` to print exception info with the traceback?

mprahl: @apaplaus I meant just creating a wrapper like `self.log_warn` but for exceptions. It wouldn't be using Flask in this case since this is the backend code.

apaplaus: Ohhh, ok) I wondered what is that mysterious method that I don't know about, so I found one in Flask)

mprahl (approved):

PR #58, opened 2021-02-15
Add repo ids to content sets

apaplaus:
Now content_sets list will consist of pulp
repository ids and content sets assiciated with those repositories

JIRA: CWFHEALTH-22


freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
```diff
@@ -387,6 +387,7 @@ def _find_images_to_rebuild(self, errata_id):
                     username=conf.pulp_username,
                     password=conf.pulp_password)
         content_sets = pulp.get_content_set_by_repo_ids(pulp_repo_ids)
+        content_sets.extend(pulp_repo_ids)
```
mprahl, line 390: Could you please add a comment here such as the following?
```
# Some container builds declare Pulp repos directly instead of content sets, but they are
# stored in the same location as content sets so they can be treated the same
```

mprahl (approved):
:+1: after the comment is added

PR #60, opened 2021-02-19
Don't generate compose for EUS images

apaplaus:
Freshmaker will not generate ODCS compose for unpublished images that
suppose to be rebuilt.

JIRA: CWFHEALTH-208

### DIFF
--- a/freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
+++ b/freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
@@ -35,6 +35,7 @@ from freshmaker.errata import Errata
 from freshmaker.types import (
     ArtifactType, ArtifactBuildState, EventState, RebuildReason)
 from freshmaker.models import Event, Compose
+from freshmaker.utils import is_in_unpublished_exceptions
 
 
 class RebuildImagesOnRPMAdvisoryChange(ContainerBuildHandler):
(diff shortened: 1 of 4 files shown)


freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
```diff
@@ -295,7 +296,8 @@ def _record_batches(self, batches, event, builds=None):
 
                 db.session.commit()
 
-                if state != ArtifactBuildState.FAILED.value:
+                is_unpublished_exception = is_in_unpublished_exceptions(image)
+                if state != ArtifactBuildState.FAILED.value and not is_unpublished_exception:
```
mprahl, line 300: To keep the same behavior, could you change the if statement below instead to something like this?
```python
if image["generate_pulp_repos"] or not (image["published"] or is_in_unpublished_exceptions(image)):
```

freshmaker/utils.py
```diff
@@ -217,3 +217,20 @@ def get_ocp_release_date(ocp_version):
...
+
+    :param ContainerImage image: image that should be checked
+    :return: True if image is published in 'exceptions' repository, False otherwise
+    """
+    for exception in conf.unpublished_exceptions:
+        exc_repo = exception["repository"]
+        exc_reg = exception["registry"]
+        for image_repo in image["repositories"]:
+            if image_repo["repository"] == exc_repo and image_repo["registry"] == exc_reg:
+                return True
```
hluk, line 235: A bit more easier to read IMO:

```python
    repo_reg_pairs = {
        (image_repo["repository"], image_repo["registry"])
        for image_repo in image["repositories"]
    }
    return any(
        (exception["repository"], exception["registry"]) in repo_reg_pairs
        for exception in conf.unpublished_exceptions
    )
```

tests/test_utils.py
```diff
@@ -40,6 +40,18 @@ def test_get_rebuilt_nvr(mock_time, rebuilt_nvr_release_suffix):
...
+@patch("freshmaker.config.Config.unpublished_exceptions",
+       new_callable=PropertyMock,
+       return_value=[{"repository": "some-repo", "registry": "some-registry"}])
+def test_is_in_unpublished_exceptions(exception):
+    images = [{"repositories": [{"registry": "some-registry", "repository": "repo"}]},
+              {"repositories": [{"registry": "registry", "repository": "some-repo"}]},
+              {"repositories": [{"registry": "some-registry", "repository": "some-repo"}]}]
+
+    results = list(map(is_in_unpublished_exceptions, images))
+    assert results == [False, False, True]
```
hluk, line 52:
```python
    assert is_in_unpublished_exceptions({"repositories": [{"registry": "some-registry", "repository": "repo"}]}) is False
    assert is_in_unpublished_exceptions({"repositories": [{"registry": "registry", "repository": "some-repo"}]}) is False
    assert is_in_unpublished_exceptions({"repositories": [{"registry": "some-registry", "repository": "some-repo"}]}) is True
```

But it's a good candidate to use `@pytest.mark.parametrize`.

apaplaus: When you use `@pytest.mark.parametrize` it generates a new test for every parameter, so with 3 parameters you will get 3 tests executed and addressed in the output. I prefer to have one output for one tested functionality so I would not use this fixture here. And secon

mprahl:
Based on #61, I believe this will not be necessary. Let's keep this open until that one is merged.

apaplaus:
Sure

mprahl:
Closing this in favor of #61

PR #61, opened 2021-02-19
Make unpublished_exceptions only apply to container repos

mprahl:
Unpublished exceptions for images such as EUS only matter at the
repository level. The repositories[].published value on a containerImage
entity is different according to Release Engineers.

This is based on the following statement from a Release Engineer:
>  I think that there is a published flag at the image level,
    which is true to indicate that the image is published in the repo.
    And there is another flag at the repo level to indicate whether the
    repo is visible in RHEC.

### DIFF
--- a/freshmaker/lightblue.py
+++ b/freshmaker/lightblue.py
@@ -904,36 +904,13 @@ class LightBlue(object):
             from.
         :param list rpm_nvrs_names: list names of the binary RPM NVRs to look for
         :param set auto_rebuild_tags: set of auto rebuild tags to add to query
-        :param bool published: whether to limit queries to published
-            repositories(some unpublished repos still will be queried)
+        :param bool published: whether to limit queries to images that are published
+            in a repository
         """
-        repo_filters = []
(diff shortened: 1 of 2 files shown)


apaplaus:
So if the image has `repositories[].published` set to False, that means that this image is not published in that repo, but why then this entry even exists? Just to tell where this image is not published? It's kind of strange.

mprahl:
@apaplaus from my understanding, `published` on the `containerRepository` entity means whether the repo is visible in the container cataolog. `published` on the `containerImage` entity means if the image has been pushed to the repo.

apaplaus:
So `published` field on the `containerImage` can't be False at all?

mprahl:
> So `published` field on the `containerImage` can't be False at all?

`published` on the `containerImage` entity would be `False` when the build is going to be shipped (e.g. attached to an advisory) but hasn't shipped yet.

apaplaus:
Oh, ok got it, thx.

apaplaus (approved):

PR #63, opened 2021-03-02
Logging ODCS compose state change

qixiang:
When see a message of ODCS compose state change, logging it so we can
check logs to know whether Freshmaker saw this message. And only logging
when this ODCS compose is connected with some builds.

### DIFF
--- a/freshmaker/handlers/internal/update_db_on_odcs_compose_fail.py
+++ b/freshmaker/handlers/internal/update_db_on_odcs_compose_fail.py
@@ -50,6 +50,9 @@ class UpdateDBOnODCSComposeFail(BaseHandler):
             Compose.odcs_compose_id == event.compose["id"],
             ArtifactBuildCompose.compose_id == Compose.id)
 
+        if builds_with_compose:
+            self.log_error("ODCS compose %s failed", event.compose["id"])
+
         for build in builds_with_compose:
             build.transition(
                 ArtifactBuildState.FAILED.value,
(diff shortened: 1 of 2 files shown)


freshmaker/handlers/koji/rebuild_images_on_odcs_compose_done.py
```diff
@@ -53,6 +53,10 @@ def handle(self, event):
             ArtifactBuild.state == ArtifactBuildState.PLANNED.value,
             Compose.odcs_compose_id == event.compose['id'],
             ArtifactBuildCompose.compose_id == Compose.id)
+
+        if builds_ready_to_rebuild:
+            self.log_info('ODCS compose %s finished', event.compose['id'])
```
apaplaus, line 58: Shouldn't this logging be without `if`? So we see a message about done compose every time handler is triggered and not only when there are builds waiting for this compose?

freshmaker/handlers/internal/update_db_on_odcs_compose_fail.py
```diff
@@ -50,6 +50,9 @@ def handle(self, event):
             Compose.odcs_compose_id == event.compose["id"],
             ArtifactBuildCompose.compose_id == Compose.id)
 
+        if builds_with_compose:
+            self.log_error("ODCS compose %s failed", event.compose["id"])
+
```
apaplaus, line 55: And here it makes sense to log error only when there are builds waiting for this compose.

freshmaker/handlers/koji/rebuild_images_on_odcs_compose_done.py
qixiang: We only cares about composes that requested by Freshmaker (there are builds in db connect to this compose when it was requested by Freshmaker), composes requested by other clients (like osbs) can simply be ignored.

apaplaus: Thanks for clarification

apaplaus (approved):

PR #64, opened 2021-03-02
Check blocking advisories for overlapping images

apaplaus:
Freshmaker while handling advisory with operator bundle will check not
only for original images, but also for images from blocking/previous
advisories and mark them for rebuild too. Because sometimes previously
rebuilt bundle image could have overriten original NVR of the
operator/operand images, so it might not be marked for rebuild event if
it should be.

JIRA: CWFHEALTH-60

### DIFF
--- a/freshmaker/errata.py
+++ b/freshmaker/errata.py
@@ -201,6 +201,12 @@ class Errata(object):
     def _get_bugs(self, errata_id):
         return self._errata_http_get("advisory/%s/bugs.json" % str(errata_id))
 
+    def _get_blocking_advisories(self, errata_id):
+        return self._errata_http_get(f"errata/blocking_errata_for/{errata_id}.json")
+
+    def _get_attached_builds(self, errata_id):
+        return self._errata_http_get(f"advisory/{errata_id}/builds.json")
+
     @region.cache_on_arguments()
     def _advisories_from_nvr(self, nvr):
         """
@@ -285,8 +291,7 @@ class Errata(object):
         :return: True if all builds in advisory are signed.
         :rtype: bool
         """
-        builds_per_product = self._errata_http_get(
-            "advisory/%s/builds.json" % str(errata_id))
+        builds_per_product = self._get_attached_builds(errata_id)
 
         # Store NVRs of all builds in advisory to nvrs set.
         nvrs = set()
@@ -365,8 +370,7 @@ class Errata(object):
         if rhel_release_prefix is None:
             rhel_release_prefix = conf.errata_rhel_release_prefix
 
-        builds_per_product = self._errata_http_get(
-            "advisory/%s/builds.json" % str(errata_id))
+        builds_per_product = self._get_attached_builds(errata_id)
 
         # Store NVRs of all builds in advisory to nvrs set.
         source_rpms = set()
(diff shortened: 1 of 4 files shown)


freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -82,6 +88,12 @@ def handle(self, event):
                 if original_nvr is None:
                     continue
                 nvrs_mapping[original_nvr] = build['nvr']
+                build_name = parse_nvr(build['nvr'])
+                # Check builds from blocking advisories and add to the mapping
+                # all of them, that have overlapping package names
+                for block_build in blocking_advisories_builds:
+                    if parse_nvr(block_build)['name'] == build_name:
```
mprahl, line 95: @apaplaus very close. Could you also please check that the version matches?

mprahl:
@apaplaus this looks good. Just a small change and then adding unit tests should do it.

qixiang:
@mprahl @apaplaus just want to confirm the purpose of this PR, is it for covering the case of blocking advisories were shipped but the rebuilt bundles (for the blocking advisories) have not been shipped yet?

apaplaus:
> @mprahl @apaplaus just want to confirm the purpose of this PR, is it for covering the case of blocking advisories were shipped but the rebuilt bundles (for the blocking advisories) have not been shipped yet?

Sorry forgot to link the issue. It's CWFHEALTH-60 you can check the description there.

freshmaker/errata.py
```diff
@@ -470,3 +474,18 @@ def get_cve_affected_rpm_nvrs(self, errata_id):
                                     nvrs.update(just_nvrs)
 
         return list(nvrs)
+
+    def get_blocking_advisories_builds(self, errata_id):
+        """ Get all advisories that block given advisory id, and fetch all builds from it
+
+        :param number errata_id: ID of advisory
```
mprahl, line 481:
```suggestion
        :param int errata_id: ID of advisory
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -322,3 +314,39 @@ def image_has_auto_rebuild_tag(self, image):
             if set(auto_rebuild_tags) & set(tags):
                 return True
         return False
+
+    def _create_original_to_rebuilt_nvrs_map(self, event):
+        """
+        Creates mapping of original build NVRs to rebuilt NVRs in advisory.
+        Including NVRs of the builds from the blocking advisories
+
+        :param BotasErrataShippedEvent event: Freshmaker's event that is
```
mprahl, line 323: Could you just use `self.event` instead?

freshmaker/errata.py
apaplaus: Well actually I am not sure about `int` because errata endpoint returns list of strings like this "2020:3095" and it's not an int but probably it could be, not sure about the return format.

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: Yes, you are right

freshmaker/errata.py
mprahl: Oh yeah, good point. Maybe omit the type?

apaplaus: Is it legal? =))

mprahl (approved):

PR #65, opened 2021-03-04
Add new API endpoint for pullspecs overrides

apaplaus:
Add API endpoint to get pullsecs overrides for bundle images builds

### DIFF
--- a/freshmaker/views.py
+++ b/freshmaker/views.py
@@ -160,6 +160,14 @@ api_v1 = {
                 'methods': ['GET'],
             }
         },
+    },
+    'pullspec_overrides': {
+        'pullspec_overrides': {
+            'url': '/api/1/pullspec_overrides/<int:id>',
+            'options': {
+                'methods': ['GET'],
+            }
+        },
     }
 }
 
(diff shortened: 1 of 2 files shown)


freshmaker/views.py
```diff
@@ -744,6 +752,20 @@ def get(self, project, repo):
         return jsonify(ret), 200
 
 
+class PullsecOverrideAPI(MethodView):
+    @freshmaker_build_api_latency.time()
+    def get(self, id):
+        if id is None:
```
mprahl, line 758: This should not be possible since the API route in Flask will not match if there is no ID provided.

freshmaker/views.py
```diff
@@ -744,6 +752,20 @@ def get(self, project, repo):
         return jsonify(ret), 200
 
 
+class PullsecOverrideAPI(MethodView):
```
qixiang, line 755: typo of `Pullspec`.

freshmaker/views.py
```diff
@@ -754,6 +776,7 @@ def get(self, project, repo):
     'about': AboutAPI,
     'verify_image': VerifyImageAPI,
     'verify_image_repository': VerifyImageRepositoryAPI,
+    'pullspec_overrides': PullsecOverrideAPI,
```
qixiang, line 779: Typo of `Pullspec`.

freshmaker/views.py
apaplaus: Yes, it works. But it doesn't show pullsec_overritdes in the common `builds` endpoint. And it returns just `null` if there are no pullspecs overrides. But since it wasn't mentioned in the tickets I didn't do anything with it.

mprahl (approved):

qixiang (approved):

PR #67, opened 2021-03-10
Override advisory state with state in state change message

qixiang:
When an ErrataAdvisoryStateChangedEvent is generated from an advisory
state change message, override the advisory state with new state in
message. This is because when there is message delay, a message of
advisory state change to other states (QE, REL_PREP, PUSH_READY) can 
arrived after an advisory has already been shipped, and rebuilds can 
be triggered for multiple times.

For example, there is an advisory 101, and its state was changed as
following:

1. 10:10:58: NEW_FILES -> QE, message A sent
2. 10:39:52: QE -> REL_PREP, message B sent
3. 10:45:07: REL_PREP -> PUSH_READY, message C sent
4. 10:46:22: PUSH_READY -> IN_PUSH, message D sent
5. 11:15:11: IN_PUSH -> SHIPPED_LIVE, message E sent

But there is something wrong on messaging, Freshmaker received the 
first message A at 12:38:02, and it checked advisory state in Errata,
because at this moment, advisory 101 has been moved to "SHIPPED_LIVE",
message A will trigger image rebuilds, the other 4 messages received
after message A will also trigger rebuilds.

### DIFF
--- a/freshmaker/parsers/errata/state_change.py
+++ b/freshmaker/parsers/errata/state_change.py
@@ -40,13 +40,18 @@ class ErrataAdvisoryStateChangedParser(BaseParser):
         msg_id = msg.get('msg_id')
         inner_msg = msg.get('msg')
         errata_id = int(inner_msg.get('errata_id'))
+        new_state = inner_msg.get('to')
 
         errata = Errata()
         advisory = ErrataAdvisory.from_advisory_id(errata, errata_id)
+        # When there is message delay, state change messages can arrive after
+        # advisory has been changed to a different state other than the one
+        # in message, so we override advisory state with the state in message
+        advisory.state = new_state
+
         # If advisory created by BOTAS and it's shipped,
         # then return BotasErrataShippedEvent event
-        if advisory.state == "SHIPPED_LIVE" and \
-           advisory.reporter.startswith('botas'):
+        if advisory.state == "SHIPPED_LIVE" and advisory.reporter.startswith('botas'):
             event = BotasErrataShippedEvent(msg_id, advisory)
         else:
             event = ErrataAdvisoryStateChangedEvent(msg_id, advisory)


freshmaker/parsers/errata/state_change.py
```diff
@@ -40,13 +40,17 @@ def parse(self, topic, msg):
         msg_id = msg.get('msg_id')
         inner_msg = msg.get('msg')
         errata_id = int(inner_msg.get('errata_id'))
+        new_state = inner_msg.get('to')
+
+        # Ingore other state change messages which are not for SHIPPED_LIVE
+        if new_state != "SHIPPED_LIVE":
```
apaplaus, line 46: Just asking, on *rebuild_images_on_rpm_advisory_change.py:107* we have condition `if event.advisory.state != 'SHIPPED_LIVE'` and based on it we will generate composes. If I understood right now it will not make sense, because we are reacting only on advisories that are already in 'SHIPPED_LIVE' state?

qixiang: `rebuild_images_on_rpm_advisory_change.py` handles `ErrataAdvisoryRPMsSignedEvent` events, but such type events can come from:

1. all rpms are signed in an advisory, then an `ErrataAdvisoryRPMsSignedEvent` is generated.
2. advisory state is moved to `SHIPPED_LIVE`, then a faked `ErrataAdvisoryRPMsSignedEvent` is generated.

Case 2 is what we're changing in this PR, and we don't need to check advisory state again. But in case 1, an advisory can be in other states other than `SHIPPED_LIVE`, so we still need to check advisory state.

Seems this `ErrataAdvisoryRPMsSignedEvent` exists for some historical reason and probably we can clean up and simply the code to have one event (e.g. `RPMAdvisoryShippedEvent`) generated directly from advisory state change message and remove the code of handling rpm signed messages. I'd recommend having a look on the code base and check whether this is true.

mprahl:
@qixiang I think this would conflict with the `HANDLER_BUILD_ALLOWLIST` configuration where you can set `advisory_state` as a filter.  Perhaps the `is_allowed` method could use the state from the message rather than checking the advisory state directly.

qixiang:
> @qixiang I think this would conflict with the `HANDLER_BUILD_ALLOWLIST` configuration where you can set `advisory_state` as a filter. Perhaps the `is_allowed` method could use the state from the message rather than checking the advisory state directly.

You're right, I updated patch to override advisory state with the state in message, this only applies to `freshmaker/parsers/errata/state_change.py`, so other state change messages (which are not to "SHIPPED_LVIE") will not generate the faked `ErrataAdvisoryRPMsSignedEvent`. I didn't change `is_allowed` because events can be generated from other messages which doesn't include advisory state.

mprahl (approved):

PR #68, opened 2021-03-16
Add support for unpublished exceptions on the verify-image API endpoint

mprahl:
The first two commits fix the unit tests.

### DIFF
--- a/freshmaker/image_verifier.py
+++ b/freshmaker/image_verifier.py
@@ -52,9 +52,13 @@ class ImageVerifier(object):
                 "rebuilt, but found %r." % (categories, repo_categories))
 
         if not repo["published"]:
-            raise ValueError(
-                "Only published repositories can be rebuilt, but this repository is not "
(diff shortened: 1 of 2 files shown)


mprahl:
It seems there is an additional test failure due to SQLAlchemy 1.4 being used for the tests.

apaplaus (approved):

apaplaus:
As I see there are still some pipeline fails.

apaplaus:
So should something be done with those fails?

apaplaus (approved):

mprahl:
@apaplaus for some reason, TravisCI is not running on my last two pushes.

mprahl:
@apaplaus it seems that we ran out of TravisCI tokens so I am trying to migrate us to use GitHub actions

mprahl:
@apaplaus, I moved the commits that fixed the unit tests to #70

PR #72, opened 2021-04-07
Add manual bundle rebuild functionality

apaplaus:
Release driver will send manual request for bundle rebuild to Freshmaker
in format:
{
  "container_images": ["app-bundle-1.0-1.22222"],
  "bundle_images": ["app-bundle-1.0-1.11111"],
  "metadata": release_driver_metadata,
}

Freshmaker will check pullspecs overrides of 'bundle_images'(the
original overwritten values) and if some of them are the same as the
current pullspec overrides in 'container_images' bundles. Then those
bundles(from 'container_images') will be rebuilt with the new pullspec overrides.

JIRA: CWFHEALTH-332

### DIFF
--- a/freshmaker/handlers/botas/botas_shipped_advisory.py
+++ b/freshmaker/handlers/botas/botas_shipped_advisory.py
@@ -29,7 +29,7 @@ import semver
 
 from freshmaker import db, conf, log
 from freshmaker.handlers import ContainerBuildHandler
-from freshmaker.events import BotasErrataShippedEvent
+from freshmaker.events import BotasErrataShippedEvent, ManualBundleRebuild
 from freshmaker.lightblue import ContainerImage
 from freshmaker.models import ArtifactBuild, ArtifactType, Event
 from freshmaker.types import EventState, ArtifactBuildState, RebuildReason
@@ -44,6 +44,10 @@ class HandleBotasAdvisory(ContainerBuildHandler):
     BOTAS to SHIPPED_LIVE state
     """
     name = "HandleBotasAdvisory"
+    # This prefix should be added to event reason, when skipping the event.
+    # Because Release Driver checks event's reason for certain prefixes,
+    # to determine if there is an error in bundles processing.
+    _no_bundle_prefix = "No bundles to rebuild: "
 
     def __init__(self, pyxis=None):
         super().__init__()
@@ -57,10 +61,15 @@ class HandleBotasAdvisory(ContainerBuildHandler):
         if not conf.freshmaker_root_url or "://" not in conf.freshmaker_root_url:
             raise ValueError("'FRESHMAKER_ROOT_URL' parameter should be set to "
                              "a valid URL")
+        # Currently processed event
+        self.event = None
 
     def can_handle(self, event):
         if (isinstance(event, BotasErrataShippedEvent) and
                 'docker' in event.advisory.content_types):
+            return True
+        # This handler can handle manual bundle rebuilds too
+        if isinstance(event, ManualBundleRebuild):
             return True
 
         return False
@@ -82,13 +91,41 @@ class HandleBotasAdvisory(ContainerBuildHandler):
             self.log_info(msg)
             return []
 
+        if isinstance(event, ManualBundleRebuild):
+            bundles_to_rebuild = self._handle_manual_rebuild(db_event)
+        else:
+            bundles_to_rebuild = self._handle_auto_rebuild(db_event)
+
+        if not bundles_to_rebuild:
+            return []
+
+        builds = self._prepare_builds(db_event, bundles_to_rebuild)
+
+        # Reset context to db_event.
+        self.set_context(db_event)
+
+        self.start_to_build_images(builds)
+        msg = f"Advisory {db_event.search_key}: Rebuilding " \
+              f"{len(db_event.builds.all())} bundle images."
+        db_event.transition(EventState.BUILDING, msg)
+
+        return []
+
+    def _handle_auto_rebuild(self, db_event):
+        """
+        Handle auto rebuild for an advisory created by Botas
+
+        :param db_event: database event that represent rebuild event
+        :rtype: list
+        :return: list of advisories that should be rebuilt
+        """
         # Mapping of original build nvrs to rebuilt nvrs in advisory
         nvrs_mapping = self._create_original_to_rebuilt_nvrs_map()
 
         original_nvrs = nvrs_mapping.keys()
         self.log_info(
             "Orignial nvrs of build in the advisory #{0} are: {1}".format(
-                event.advisory.errata_id, " ".join(original_nvrs)))
+                self.event.advisory.errata_id, " ".join(original_nvrs)))
 
         # Get image manifest_list_digest for all original images, manifest_list_digest is used
         # in pullspecs in bundle's related images
@@ -237,9 +274,11 @@ class HandleBotasAdvisory(ContainerBuildHandler):
             for pullspec in bundle_data['pullspecs']:
                 # A pullspec item example:
                 # {
-                #   'new': 'registry.exampe.io/repo/example-operator@sha256:<sha256-value>'
+                #   'new': 'registry.exampe.io/repo/example-operator@sha256:<sha256-value>',
                 #   'original': 'registry.example.io/repo/example-operator:v2.2.0',
-                #   'pinned': True
+                #   'pinned': True,
+                #   # value used for internal purpose during manual rebuilds, it's an old pullspec that was replaced
+                #   '_old': 'registry.exampe.io/repo/example-operator@sha256:<previous-sha256-value>,
                 # }
 
                 # A pullspec path is in format of "registry/repository@digest"
@@ -255,6 +294,9 @@ class HandleBotasAdvisory(ContainerBuildHandler):
                 new_nvr = nvrs_mapping[old_nvr]
                 new_digest = rebuilt_digests_by_nvr[new_nvr]
 
+                # save pullspec that image had before rebuild
+                pullspec['_old'] = pullspec.get('new')
+
                 # Replace the old digest with new digest
                 pullspec_elems[1] = new_digest
                 new_pullspec = '@'.join(pullspec_elems)
@@ -269,24 +311,127 @@ class HandleBotasAdvisory(ContainerBuildHandler):
                 to_rebuild_digests.add(digest)
 
         if not to_rebuild_digests:
-            msg = f"No bundle images to rebuild for advisory {event.advisory.name}"
+            msg = self._no_bundle_prefix + "No bundle images to rebuild for " \
+                                           f"advisory {self.event.advisory.name}"
             self.log_info(msg)
             db_event.transition(EventState.SKIPPED, msg)
             db.session.commit()
             return []
 
-        builds = self._prepare_builds(db_event, bundles_by_digest,
-                                      to_rebuild_digests)
+        bundles_to_rebuild = list(map(lambda x: bundles_by_digest[x],
+                                      to_rebuild_digests))
+        return bundles_to_rebuild
 
-        # Reset context to db_event.
-        self.set_context(db_event)
+    def _handle_manual_rebuild(self, db_event):
+        """
+        Handle manual rebuild submitted by Release Driver for an advisory created by Botas
 
-        self.start_to_build_images(builds)
-        msg = f"Advisory {db_event.search_key}: Rebuilding " \
-              f"{len(db_event.builds.all())} bundle images."
-        db_event.transition(EventState.BUILDING, msg)
+        :param db_event: database event that represents a rebuild event
+        :rtype: list
+        :return: list of advisories that should be rebuilt
+        """
+        old_to_new_pullspec_map = self._get_pullspecs_mapping()
 
-        return []
+        if not old_to_new_pullspec_map:
+            msg = self._no_bundle_prefix + 'None of the bundle images have ' \
+                                           'applicable pullspecs to replace'
+            log.warning(msg)
+            db_event.transition(EventState.SKIPPED, msg)
+            return []
+
+        # Unauthenticated koji session to fetch build info of bundles
+        koji_api = KojiService(conf.koji_profile)
+        rebuild_nvr_to_pullspecs_map = dict()
+        # compare replaced pullspecs with pullspecs in 'container_images' and
+        # create map for bundles that should be rebuilt with their nvrs
+        for container_image_nvr in self.event.container_images:
+            artifact_build = db.session.query(ArtifactBuild).filter(
+                ArtifactBuild.rebuilt_nvr == container_image_nvr,
+                ArtifactBuild.type == ArtifactType.IMAGE.value,
+            ).one_or_none()
+            pullspecs = []
+            # Try to find build in FM database, if it's not there check in Brew
+            if artifact_build:
+                pullspecs = artifact_build.bundle_pullspec_overrides["pullspecs"]
+            else:
+                # Fetch buildinfo from Koji
+                buildinfo = koji_api.get_build(container_image_nvr)
+                # Get the original pullspecs
+                pullspecs = (
+                    buildinfo.get('extra', {})
+                             .get('image', {})
+                             .get('operator_manifests', {})
+                             .get('related_images', {})
+                             .get('pullspecs', [])
+                )
+
+            for pullspec in pullspecs:
+                if pullspec.get('new') not in old_to_new_pullspec_map:
+                    continue
+                # use newer pullspecs in the image
+                pullspec['new'] = old_to_new_pullspec_map[pullspec['new']]
+                rebuild_nvr_to_pullspecs_map[container_image_nvr] = pullspecs
+
+        if not rebuild_nvr_to_pullspecs_map:
+            msg = self._no_bundle_prefix + 'None of the container images have ' \
+                                           'applicable pullspecs from the input bundle images'
+            log.info(msg)
+            db_event.transition(EventState.SKIPPED, msg)
+            return []
+
+        # list with metadata about every bundle to do rebuild
+        to_rebuild_bundles = []
+        # fill 'append' and 'update' fields for bundles to rebuild
+        for nvr, pullspecs in rebuild_nvr_to_pullspecs_map.items():
+            bundle_digest = self._pyxis.get_manifest_list_digest_by_nvr(nvr)
+            if bundle_digest is not None:
+                bundles = self._pyxis.get_bundles_by_digest(bundle_digest)
+                temp_bundle = bundles[0]
+                csv_updates = (self._get_csv_updates(temp_bundle['csv_name'],
+                                                     temp_bundle['version']))
+                to_rebuild_bundles.append({
+                    'nvr': nvr,
+                    'update': csv_updates['update'],
+                    'pullspecs': pullspecs,
+                })
+            else:
+                log.warning('Can\'t find manifest_list_digest for bundle '
+                            f'"{nvr}" in Pyxis')
+
+        if not to_rebuild_bundles:
+            msg = 'Can\'t find digests for any of the bundles to rebuild'
+            log.warning(msg)
+            db_event.transition(EventState.FAILED, msg)
+            return []
+
+        return to_rebuild_bundles
+
+    def _get_pullspecs_mapping(self):
+        """
+        Get map of all replaced pullspecs from 'bundle_images' provided in an event.
+
+        :rtype: dict
+        :return: map of all '_old' pullspecs that was replaced by 'new'
+            pullspecs in previous Freshmaker rebuilds
+        """
+        old_to_new_pullspec_map = dict()
+        for bundle_nvr in self.event.bundle_images:
+            artifact_build = db.session.query(ArtifactBuild).filter(
+                ArtifactBuild.rebuilt_nvr == bundle_nvr,
+                ArtifactBuild.type == ArtifactType.IMAGE.value,
+            ).one_or_none()
+            if artifact_build is None:
+                log.warning(
+                    f'Can\'t find build for a bundle image "{bundle_nvr}"')
+                continue
+            pullspec_overrides = artifact_build.bundle_pullspec_overrides
+            for pullspec in pullspec_overrides['pullspecs']:
+                old_pullspec = pullspec.get('_old', None)
+                if old_pullspec is None:
+                    continue
+                old_to_new_pullspec_map[old_pullspec] = pullspec['new']
+
+        return old_to_new_pullspec_map
 
     @classmethod
     def _get_csv_updates(cls, csv_name, version):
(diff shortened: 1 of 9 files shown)


freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -82,13 +85,41 @@ def handle(self, event):
             self.log_info(msg)
             return []
 
+        if isinstance(event, ManualBundleRebuild):
+            bundles_to_rebuild = self._handle_manual_rebuild(db_event)
+        else:
+            bundles_to_rebuild = self._handle_auto_rebuild(db_event)
+
+        if not bundles_to_rebuild:
+            return []
```
mprahl, line 94: Could you please set the event to skipped in this case?

mprahl: Nevermind, you handle that in the helper methods.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -270,24 +306,107 @@ def handle(self, event):
...
+        :return: list of advisories that should be rebuilt
+        """
+        default_bundle_data = {
+            'nvr': '',
+            'update': {},
+            'append': {},
+            'pullspecs': [],
+        }
+        # map with old pullspec and bundle info
+        old_pullspec_to_bundle_map = dict()
```
mprahl, line 334: Optional:
```suggestion
        old_pullspec_to_bundle_map = {}
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -270,24 +306,107 @@ def handle(self, event):
...
+        # get all replaced pullspecs from provided 'bundle_images'
+        for bundle_nvr in self.event.bundle_images:
+            artifact_build = db.session.query(ArtifactBuild).filter(
+                ArtifactBuild.rebuilt_nvr == bundle_nvr,
+                ArtifactBuild.type == ArtifactType.IMAGE.value,
+            ).one_or_none()
+            pullspec_overrides = artifact_build.bundle_pullspec_overrides
```
mprahl, line 341: You need to handle if `artifact_build` is `None` before doing this.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -270,24 +306,107 @@ def handle(self, event):
...
+            for pullspec in pullspec_overrides:
+                old_pullspec = pullspec.get('_old', None)
+                if old_pullspec is not None:
```
mprahl, line 344: Optional: to reduce the indentation below which makes it easier to read.
```suggestion
                if old_pullspec is None:
                    continue
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -270,24 +306,107 @@ def handle(self, event):
...
+                    old_pullspec_to_bundle_map.setdefault(
+                        old_pullspec, copy.deepcopy(default_bundle_data))
+                    # bundle will now contain new pullspec overrides
+                    old_pullspec_to_bundle_map[old_pullspec]['pullspecs'] = pullspec_overrides
+                    break
+
+        if not old_pullspec_to_bundle_map:
+            msg = 'None of the bundle images have old pullspecs overrides in FM database'
```
mprahl, line 352:
```suggestion
            msg = 'None of the bundle images have applicable pullspecs to replace'
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -270,24 +306,107 @@ def handle(self, event):
...
+            log.warning(msg)
+            db_event.transition(EventState.SKIPPED, msg)
+            return []
 
-        return []
+        # Unauthenticated koji session to fetch build info of bundles
+        koji_api = KojiService(conf.koji_profile)
+        rebuild_nvr_to_bundle_map = dict()
```
mprahl, line 359: Optional:
```suggestion
        rebuild_nvr_to_bundle_map = {}
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -270,24 +306,107 @@ def handle(self, event):
...
+            # Get the original pullspecs
+            pullspecs = (
+                buildinfo.get('extra', {})
+                         .get('image', {})
+                         .get('operator_manifests', {})
+                         .get('related_images', {})
+                         .get('pullspecs', [])
+            )
+            for pullspec in pullspecs:
+                if pullspec.get('new') in old_pullspec_to_bundle_map:
```
mprahl, line 374: Optional: This will make it easier to read.
```suggestion
                if pullspec.get('new') not in old_pullspec_to_bundle_map:
                    continue
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -270,24 +306,107 @@ def handle(self, event):
...
+        for bundle_nvr in self.event.bundle_images:
+            artifact_build = db.session.query(ArtifactBuild).filter(
+                ArtifactBuild.rebuilt_nvr == bundle_nvr,
+                ArtifactBuild.type == ArtifactType.IMAGE.value,
+            ).one_or_none()
+            pullspec_overrides = artifact_build.bundle_pullspec_overrides
+            for pullspec in pullspec_overrides:
+                old_pullspec = pullspec.get('_old', None)
+                if old_pullspec is not None:
+                    old_pullspec_to_bundle_map.setdefault(
```
mprahl, line 345: If I'm reading this PR correctly, I think the logic is reversed. It should be that for all `container_images` (these are actually bundle images but just reusing the key that the API already accepts), check if any of the pullspecs that Freshmaker replaced in each bundle image in `bundle_images` has an "_old" (as you call it in this PR) pullspec that matches a pinned pullspec in the container image (from the `container_images` list).

Does that make sense?

apaplaus: Yes, so at first I am collecting all the '_old' pullspecs from the submitted `bundle_images`. And then comparing them with the pullspecs of images from `container_images` list. I have it there  
```
        for container_image_nvr in self.container_images:
            # Fetch buildinfo
            buildinfo = koji_api.get_build(container_image_nvr)
            # Get the original pullspecs
            pullspecs = (
                buildinfo.get('extra', {})
                         .get('image', {})
                         .get('operator_manifests', {})
                         .get('related_images', {})
                         .get('pullspecs', [])
            )
            for pullspec in pullspecs:
                if pullspec.get('new') in old_pullspec_to_bundle_map:
                ....
``` 
So at the end I am comparing `container_images->pullspecs->new` with the `bundle_images->pullspecs->_old` for every **pullspec** in every **container_image**. Is it right?

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: I prefer using `dict()` because for me it's kind of more readable. When you what to know what is `old_pullspec_to_bundle_map`, it straight away says that it's a `dict`. But with `{}`, I know that sets can be created with that kind of parenthesis, and if I forgot that empty `{}` means `dict` It could take me some time to remember or I can just overlook it.

mprahl: No problem. `{}` is faster since the Python bytecode doesn't need to find the global of `dict`. I find it nicer to read but it's up to you. There's no style guide for Freshmaker so it's a free for all.  :smile:

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -270,24 +306,112 @@ def handle(self, event):
...
+        # Unauthenticated koji session to fetch build info of bundles
+        koji_api = KojiService(conf.koji_profile)
+        rebuild_nvr_to_bundle_map = dict()
+        # compare replaced pullspecs with pullspecs in 'container_images' and
+        # create map for bundles that should be rebuilt with their nvrs
+        for container_image_nvr in self.container_images:
+            # Fetch buildinfo
+            buildinfo = koji_api.get_build(container_image_nvr)
+            # Get the original pullspecs
+            pullspecs = (
```
mprahl, line 370: Could you please check if `container_image` was a Freshmaker rebuild? If so, use the pullspec replacements stored in the DB of it so that way we inherit the `_old` values and know in total which pullspecs Freshmaker replaced.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -270,24 +306,112 @@ def handle(self, event):
...
+        default_bundle_data = {
+            'nvr': '',
+            'update': {},
+            'append': {},
+            'pullspecs': [],
+        }
+        # map with old pullspec and bundle info
+        old_pullspec_to_bundle_map = dict()
+        # get all replaced pullspecs from provided 'bundle_images'
+        for bundle_nvr in self.event.bundle_images:
```
mprahl, line 336: Could you move this for loop of code to a helper method to make it easier to read?

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -270,24 +306,112 @@ def handle(self, event):
...
+                buildinfo.get('extra', {})
+                         .get('image', {})
+                         .get('operator_manifests', {})
+                         .get('related_images', {})
+                         .get('pullspecs', [])
+            )
+            for pullspec in pullspecs:
+                if pullspec.get('new') not in old_pullspec_to_bundle_map:
+                    continue
+                bundle = old_pullspec_to_bundle_map[old_pullspec]
```
mprahl, line 380: So we don't want to replace all the pullspecs, we just want to replace the pullspecs in `container_images` that still reference `_old` in any of `bundle_images`.

So the flow would be as follows:
* Create a dictionary of `_old` to `new` pullspec entries
* Loop through each entry in `bundle_images` and for each entry, loop through their pullspecs stored in FM if available
* For each pullspec that has `_old`, create an entry in the dictionary of `_old` to `new`. This gives you the all possible replacements that could apply to any of the entries in `container_images`.
* For each entry in `container_images`, get the pullspecs from Freshmaker if present, if not, fallback to Brew for this metadata
* For each pullspec in the pullspecs of an entry in `container_images`, check if the `new` pullspec has an entry in the `_old` to `new` dictionary, if yes, replace `new` with this entry in the dictionary.

Does this make sense?

qixiang:
Beside of these changes, you need add the new event types in https://github.com/redhat-exd-rebuilds/freshmaker/blob/ba1fe9adb9d970542eba72b134eff908efee3ade/freshmaker/models.py#L51.

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: Yes, I will do it this way.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -270,24 +306,125 @@ def handle(self, event):
...
+            db_event.transition(EventState.SKIPPED, msg)
+            return []
+
+        # list with metadata about every bundle to do rebuild
+        to_rebuild_bundles = []
+        # fill 'append' and 'update' fields for bundles to rebuild
+        for nvr, pullspecs in rebuild_nvr_to_pullspecs_map.items():
+            bundle_digest = self._pyxis.get_manifest_list_digest_by_nvr(nvr)
+            if bundle_digest is not None:
+                # TODO I am not sure if there could be more than one bundle for particular digest, I assume not
```
mprahl, line 380: You are correct. You can remove this comment.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -270,24 +306,125 @@ def handle(self, event):
...
+                    'nvr': nvr,
+                    'update': csv_updates['update'],
+                    'append': csv_updates['append'],
+                    'pullspecs': pullspecs,
+                })
+            else:
+                log.warning(f'Can\'t find manifest_list_digest for bundle "{nvr}" in Pyxis')
+
+        if not to_rebuild_bundles:
+            msg = 'Can\'t find digests for any of bundles to rebuild'
```
mprahl, line 395:
```suggestion
            msg = 'Can\'t find digests for any of the bundles to rebuild'
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -270,24 +306,125 @@ def handle(self, event):
...
+
+        return to_rebuild_bundles
+
+    def _get_pullspecs_mapping(self):
+        """
+        Get map of all replaced pullspecs from 'bundle_images' provided in an event.
+
+        :rtype: dict
+        :return: map of all '_old' pullspecs that was replaced by 'new'
+            pullspecs in some of previous Freshmaker rebuilds
```
mprahl, line 408:
```suggestion
            pullspecs in previous Freshmaker rebuilds
```

freshmaker/views.py
```diff
@@ -756,9 +773,12 @@ class PullspecOverrideAPI(MethodView):
     @freshmaker_build_api_latency.time()
     def get(self, id):
         build = models.ArtifactBuild.query.filter_by(id=id).first()
-        # 'bundle_pullspec_overrides' should already contain valid JSON
         if build:
-            return jsonify(build.bundle_pullspec_overrides), 200
+            pullspec_overrides = build.bundle_pullspec_overrides
+            # remove old pullspec from output, since it used only internally for manual rebuilds
+            if pullspec_overrides and "_old" in pullspec_overrides:
```
mprahl, line 779: Isn't `pullspec_overrides` a list of dictionaries? I think you'll need to loop through each pullspec and delete the `_old` key if present.

mprahl (changes requested):
This looks good! Just a few comments and then go ahead and add/fix the tests.

freshmaker/views.py
apaplaus: Yes, you are right, I will change it.

apaplaus: @mprahl Wait, I just realized that pullspec_overrides is a disc that looks like:
```
{
  "update": csv_modification,
  "pullspecs": [{pullspec_1}, {pullspec_2}]
}
```
Right? So I need to check `if pullspec_overrides and "pullspecs" in pullspec_overrides:` instead. And then go through all pullspecs.

apaplaus: And API endpoint should not return a list of dicts, but dict with `pullspecs` key that should contain all the pullpecs dicts, right?

mprahl: @apaplaus ah right. So you want to return `bundle_pullspec_overrides` as is except for you want to remove any `_old` key in the dictionaries in the list in `pullspecs`.

apaplaus: ok, I will adjust tests then

apaplaus:
@mprahl So the changes, as well as tests, are done. And this error with docs, I assume it should be fixed by your other MR.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -269,24 +307,125 @@ def handle(self, event):
                 to_rebuild_digests.add(digest)
 
         if not to_rebuild_digests:
-            msg = f"No bundle images to rebuild for advisory {event.advisory.name}"
+            msg = f"No bundle images to rebuild for advisory {self.event.advisory.name}"
             self.log_info(msg)
             db_event.transition(EventState.SKIPPED, msg)
```
qixiang, line 312: Hi @apaplaus , could you update the event with a consistent string at the beginning of reason string? Release-Driver needs to check this reason string to know whether there is something wrong or just there is no bundle image need to be rebuilt. For example, something like "No bundle images to rebuild for advisory ...: <exact reason>".

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -269,24 +307,125 @@ def handle(self, event):
...
+        :rtype: list
+        :return: list of advisories that should be rebuilt
+        """
+        old_to_new_pullspec_map = self._get_pullspecs_mapping()
 
-        return []
+        if not old_to_new_pullspec_map:
+            msg = 'None of the bundle images have applicable pullspecs to replace'
+            log.warning(msg)
+            db_event.transition(EventState.SKIPPED, msg)
```
qixiang, line 333: As above, the state reason string need to be updated.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -269,24 +307,125 @@ def handle(self, event):
...
+                    continue
+                # use newer pullspecs in the image
+                pullspec['new'] = old_to_new_pullspec_map[pullspec['new']]
+                rebuild_nvr_to_pullspecs_map[container_image_nvr] = pullspecs
+
+        if not rebuild_nvr_to_pullspecs_map:
+            msg = 'None of the container images have applicable pullspecs ' \
+                  'from the input bundle images'
+            log.info(msg)
+            db_event.transition(EventState.SKIPPED, msg)
```
qixiang, line 373: As above, the state reason string need to be updated.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -269,24 +307,125 @@ def handle(self, event):
...
+                    'pullspecs': pullspecs,
+                })
+            else:
+                log.warning('Can\'t find manifest_list_digest for bundle '
+                            f'"{nvr}" in Pyxis')
+
+        if not to_rebuild_bundles:
+            msg = 'Can\'t find digests for any of the bundles to rebuild'
+            log.warning(msg)
+            db_event.transition(EventState.SKIPPED, msg)
```
qixiang, line 398: As above, the state reason string need to be updated.

apaplaus: @qixiang  But this situation is an error state. Shouldn't be there some other prefix? That the situation when none of the submitted `container_images` have digest in Pyxis, so it's not just 'No bundle to rebuild' situation.

qixiang: Is this an error? If yes, can we move it to `FAILED` instead. If freshmaker always move events to `FAILED` when there is error happens, you don't need to update these messages, release-driver can just ignore `state_reason` and only check the state.

apaplaus: I am not sure. @mprahl what do you think? Is this an error, or just an unordinary situation?

apaplaus:
@qixiang So I've changed reason strings, can you check it?

qixiang:
> @qixiang So I've changed reason strings, can you check it?

thanks, LGTM.

freshmaker/parsers/internal/manual_rebuild.py
```diff
@@ -44,16 +44,21 @@ def parse_post_data(self, data):
             from the UMB message sent from Frontend to Backend.
         """
         msg_id = data.get('msg_id', "manual_rebuild_%s" % (str(time.time())))
-        errata_id = data.get('errata_id')
         dry_run = data.get('dry_run', False)
 
-        errata = Errata()
-        advisory = ErrataAdvisory.from_advisory_id(errata, errata_id)
+        if data.get('errata_id', None):
```
mprahl, line 49: Is `errata_id` set if `freshmaker_event_id` is supplied instead? If it's not, then this conditional will not work as intended.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -82,13 +91,41 @@ def handle(self, event):
...
+        if not bundles_to_rebuild:
+            return []
+
+        builds = self._prepare_builds(db_event, bundles_to_rebuild)
+
+        # Reset context to db_event.
+        self.set_context(db_event)
+
+        self.start_to_build_images(builds)
+        msg = f"Advisory {db_event.search_key}: Rebuilding " \
```
mprahl, line 108: I don't see where `search_key` is set for `ManualBundleRebuild` and it's a required column in the database.

@qixiang what do you think about having `search_key` by a comma separated list of the values in `container_images`?

freshmaker/handlers/botas/botas_shipped_advisory.py
mprahl: In my opinion, this is an error.

mprahl:
Could you please add a test for the API endpoint when requesting a manual rebuild for bundle images?

mprahl (changes requested):

freshmaker/parsers/internal/manual_rebuild.py
apaplaus: I think it's set in `post()` method on *views.py:521*. If `freshmaker_event_id` is supplied then it sets `errata_id` to the search_key of the dependent event.

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: Well, `search_key` for `ManualBundleRebuild` is inherited from `BaseEvent`. And there it's actually `msg_id`. And `msg_id` created in `FreshmakerManualRebuildParser.parse_post_data()` and it's generated by FM.

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: Ok, so I will transition the event to the FAILED state, and remove prefix from the message

apaplaus:
@mprahl *new changes incoming*

tests/test_views.py
```diff
@@ -779,6 +794,52 @@ def test_manual_rebuild(self, time, from_advisory_id, publish):
             {'msg_id': 'manual_rebuild_123', u'errata_id': 1,
              'requester': 'root'})
 
+    @patch('freshmaker.messaging.publish')
+    @patch('freshmaker.parsers.internal.manual_rebuild.ErrataAdvisory.'
```
mprahl, line 798: Is this needed?

tests/test_views.py
```diff
@@ -779,6 +794,52 @@ def test_manual_rebuild(self, time, from_advisory_id, publish):
...
+           'from_advisory_id')
+    @patch('freshmaker.parsers.internal.manual_rebuild.time.time')
+    @patch('freshmaker.views._validate_rebuild_request', return_value=None)
```
mprahl, line 801: Why is this mocked?

apaplaus: Because we are testing other method, not `_validate_rebuild_request`.

tests/test_views.py
apaplaus: Yes, you are right, it's not needed

tests/test_views.py
mprahl: Could you please create a test then that runs that function for a manual bundle rebuid?

mprahl:
@apaplaus could you please fix the CI failure?

tests/test_views.py
apaplaus: Added

mprahl:
@apaplaus this looks good. Could you please fix the CI failure?

apaplaus:
> @apaplaus could you please fix the CI failure?

Yeah, I am trying to do it, but I can't reproduce this behavior locally.

apaplaus:
Ok, I was able to reproduce it locally, I needed to delete already generated docs directory.

mprahl (approved):

PR #74, opened 2021-04-20
Check extra image repo before start

mvalik:
JIRA: RHELWF-3065

### DIFF
--- a/freshmaker/lightblue.py
+++ b/freshmaker/lightblue.py
@@ -117,6 +117,11 @@ class KojiLookupError(ValueError):
     pass
 
 
+class ExtraRepoNotConfiguredError(ValueError):
+    """ Extra repo required but missing in config """
+    pass
+
+
 class ContainerRepository(dict):
     """Represent a container repository"""
 
@@ -228,6 +233,15 @@ class ContainerImage(dict):
                         "Cannot find task_id or container_koji_task_id "
                         "in the Koji build %r" % build)
 
+            fs_koji_task_id = build.get('extra', {}).get('filesystem_koji_task_id')
+            if fs_koji_task_id:
+                parsed_nvr = koji.parse_NVR(nvr)
+                name_version = f'{parsed_nvr["name"]}-{parsed_nvr["version"]}'
+                if name_version not in conf.image_extra_repo:
+                    msg = (f'{name_version} is a base image, but extra image repo for it '
+                           f'is not specified in the Freshmaker configuration.')
+                    raise ExtraRepoNotConfiguredError(msg)
+
             # Get the list of ODCS composes used to build the image.
             extra_image = build.get("extra", {}).get("image", {})
             if extra_image.get("odcs", {}).get("compose_ids"):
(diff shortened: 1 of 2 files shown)


mvalik:
@mprahl @qixiang please review

freshmaker/lightblue.py
```diff
@@ -228,6 +229,15 @@ def get_additional_data_from_koji(cls, nvr):
...
                         "in the Koji build %r" % build)
 
+            fs_koji_task_id = build.get('extra', {}).get('filesystem_koji_task_id')
+            if fs_koji_task_id:
+                parsed_nvr = koji.parse_NVR(nvr)
+                name_version = f'{parsed_nvr["name"]}-{parsed_nvr["version"]}'
+                if name_version not in conf.image_extra_repo:
+                    msg = (f'Build for {name_version} contains filesystem_koji_task_id, '
+                           f'but extra image repo is not specified.')
+                    raise ValidationError(msg)
```
qixiang, line 239: We need to catch this exception and update the image with default additional data when it happens, just like what we're doing for the `KojiLookupError`: https://github.com/redhat-exd-rebuilds/freshmaker/blob/6d0c059439ec2f7ad8235c391e94350c5c0eb437/freshmaker/lightblue.py#L302

freshmaker/lightblue.py
```diff
@@ -228,6 +229,15 @@ def get_additional_data_from_koji(cls, nvr):
                         "Cannot find task_id or container_koji_task_id "
                         "in the Koji build %r" % build)
 
+            fs_koji_task_id = build.get('extra', {}).get('filesystem_koji_task_id')
+            if fs_koji_task_id:
+                parsed_nvr = koji.parse_NVR(nvr)
+                name_version = f'{parsed_nvr["name"]}-{parsed_nvr["version"]}'
+                if name_version not in conf.image_extra_repo:
+                    msg = (f'Build for {name_version} contains filesystem_koji_task_id, '
```
qixiang, line 237: I'd suggest an error message like this:
{name_version} is a base image, extra repo in Freshmaker configuration is missing for building this base image.

mvalik:
Comments resolved

freshmaker/lightblue.py
```diff
@@ -299,7 +309,7 @@ def resolve_commit(self):
         # Find the additional data for Container build in Koji.
         try:
             data = self.get_additional_data_from_koji(self.nvr)
-        except KojiLookupError as e:
+        except (KojiLookupError, ValidationError) as e:
```
qixiang, line 312: The `data["error"]` in this except clause is not correct for this type of missing extra repo error.
Can you have a separate except clause to add the corresponding error to `data`? The error message will be present as the failed reason for this build, so we want more meaningful error message, and better to define a new custom Exception
```
    except MissingInstallMediaRepoError as e:
        log.error(e)
        data = self._get_default_additional_data()
        data["error"] = str(e)
```

mvalik:
Comments resolved

qixiang (approved):
LGTM

mvalik:
> LGTM

Thank you. Could you please merge it, cause I don't have rights for it?

PR #75, opened 2021-05-10
Fix get build's NVR from errata advisory builds data

qixiang:


tests/handlers/botas/test_botas_shipped_advisory.py
```diff
@@ -431,8 +431,8 @@ def test_get_original_nvrs(self, get_build):
         event = BotasErrataShippedEvent("test_msg_id", self.botas_advisory)
         self.botas_advisory._builds = {
             "product_name": {
-                "builds": [{"nvr": "some_name-2-2"},
-                           {"nvr": "some_name_two-2-2"}]
+                "builds": [{"some_name-2-2": {"nvr": "some_name-2-2"}},
+                           {"some_name_two-2-2": {"nvr": "some_name_two-2-2"}}]
```
apaplaus, line 435: If nvr is a key, then there is no 'nvr' : '(some_nvr_1_1)' kind of pairs as values? Shouldn't it be changed then? I mean shouldn't it be then 
```
  "builds": [{"some_name-2-2": {...}},
             {"some_name_two-2-2": {...}}]
```
Or something like that?

qixiang: nvr is still one of the key-paris in the values, so we don't need to change this fake data format.

apaplaus: ok then

apaplaus:
LGFM

PR #76, opened 2021-05-10
Fix Pyxis filter request param in get_latest_bundles

qixiang:
change

    f'latest_in_channel==true%20and%20source_index_container_path=={path}'
to
    f'latest_in_channel==true and source_index_container_path=={path}'

requests.get will encode the params, no necessary to encode it before
sending the request, and it will result in bad request.


qixiang:
Rebased on top of latest master.

PR #77, opened 2021-05-11
Delete the default fedmsg logging configuration in the container

mprahl:
This will allow logging to be configured in /etc/fedmsg.d.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,8 @@ RUN \
     pip3 install . --no-deps
 
 RUN mkdir -p /usr/share/freshmaker && cp contrib/freshmaker.wsgi /usr/share/freshmaker/
+# Delete the default logging configuration
+RUN rm -f fedmsg.d/freshmaker-logging.py
 
 RUN \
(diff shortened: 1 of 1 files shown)


mprahl:
@qixiang and @apaplaus, I noticed that there weren't debug logs for Freshmaker in our prod deployment even though it seems to be configured that way. I think this may be the culprit.

qixiang (approved):
LGTM

apaplaus (approved):

Dockerfile
```diff
@@ -26,6 +26,8 @@ RUN \
     pip3 install . --no-deps
 
 RUN mkdir -p /usr/share/freshmaker && cp contrib/freshmaker.wsgi /usr/share/freshmaker/
+# Delete the default logging configuration
+RUN rm -f fedmsg.d/freshmaker-logging.py
```
hluk, line 30: It might be better not use `-f` flag here which would indicate error in case the current directory is different or the file no longer exists in the repo.

PR #82, opened 2021-06-01
Get image digest from non-osbs registry repository

qixiang:
Internal testing images will not be published, and the repository will be
unpublished, getting image digests from published repository will block
the testing with these images, so just skip osbs registry repositories
to allow getting digest from unpublished repositories.

### DIFF
--- a/freshmaker/pyxis.py
+++ b/freshmaker/pyxis.py
@@ -191,7 +191,10 @@ class Pyxis(object):
         # get manifest_list_digest of ContainerImage from Pyxis
         for image in self._pagination(f'images/nvr/{nvr}', request_params):
             for repo in image.get('repositories'):
-                if repo['published'] and 'manifest_list_digest' in repo:
+                # skip internal build registry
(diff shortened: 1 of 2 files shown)


mprahl:
@qixiang after looking at the code, your change is exactly what we need. The only thing is the commit message is not quite accurate.

Could you please change it to the following?
```
There may be a delay from the moment when a rebuilt operator/operand image is shipped and it getting marked as published in Pyxis. To workaround this, use the manifest list digest of a non-build registry regardless of if the image is marked as published yet. This is safe since manifest list digests will not change.
```

Once the message is changed, please go ahead and merge it.

qixiang:
Updated commit message, however we need to check beside of updating repository to published: True, does the delay also happen to the digests availability in these non-build registry, if there is also delay of the digests availability, this PR won't solve the problem and can be dropped.

qixiang:
Confirmed with Mark, digests are populated at the same time with setting published to True, so this actually doesn't work, closing.

PR #83, opened 2021-06-02
Allow manifest list digests from unpublished repositories

mprahl:
The issue is that there is a delay from when a BOTAS advisory
is shipped and when Pyxis is updated with the published
repositories entry on the image. The manifest list digest
is the same in both cases, so it's safe to fallback to the
unpublished manifest list digest value for this case.

### DIFF
--- a/freshmaker/handlers/botas/botas_shipped_advisory.py
+++ b/freshmaker/handlers/botas/botas_shipped_advisory.py
@@ -153,7 +153,10 @@ class HandleBotasAdvisory(ContainerBuildHandler):
         rebuilt_digests_by_nvr = {}
         rebuilt_nvrs = nvrs_mapping.values()
         for nvr in rebuilt_nvrs:
-            digest = self._pyxis.get_manifest_list_digest_by_nvr(nvr)
+            # Don't require that the manifest list digest be published in this case because
+            # there's a delay from after an advisory is shipped and when the published repositories
+            # entry is populated
+            digest = self._pyxis.get_manifest_list_digest_by_nvr(nvr, must_be_published=False)
             if digest:
                 rebuilt_digests_by_nvr[nvr] = digest
             else:
--- a/freshmaker/pyxis.py
+++ b/freshmaker/pyxis.py
@@ -178,11 +178,13 @@ class Pyxis(object):
 
         return latest_bundles
 
-    def get_manifest_list_digest_by_nvr(self, nvr):
+    def get_manifest_list_digest_by_nvr(self, nvr, must_be_published=True):
         """
         Get image's digest(manifest_list_digest field) by its NVR
 
         :param str nvr: NVR of ContainerImage to query Pyxis
+        :param bool must_be_published: determines if the image must be published to the repository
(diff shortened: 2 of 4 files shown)


freshmaker/pyxis.py
```diff
@@ -178,21 +178,30 @@ def get_latest_bundles(self, index_images):
...
-            for repo in image.get('repositories'):
+            for repo in image['repositories']:
                 if repo['published'] and 'manifest_list_digest' in repo:
                     return repo['manifest_list_digest']
+            # If no published manifest list digest exists, check if there is an unpublished one
+            # if must_be_published is False
+            if not must_be_published:
+                for repo in image['repositories']:
+                    if 'manifest_list_digest' in repo:
+                        return repo['manifest_list_digest']
```
qixiang, line 204: If the `manifest_list_digest` are same, I think we are safe to do this:
```suggestion
                if must_be_published and not repo['published']:
                    continue
                if 'manifest_list_digest' in repo:
                    return repo['manifest_list_digest']
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -132,7 +132,10 @@ def _handle_auto_rebuild(self, db_event):
         original_digests_by_nvr = {}
         original_nvrs_by_digest = {}
         for nvr in original_nvrs:
-            digest = self._pyxis.get_manifest_list_digest_by_nvr(nvr)
+            # Don't require that the manifest list digest be published in this case because
+            # there's a delay from after an advisory is shipped and when the published repositories
+            # entry is populated
+            digest = self._pyxis.get_manifest_list_digest_by_nvr(nvr, must_be_published=False)
```
qixiang, line 138: I believe we should `must_be_published=False` for rebuilt nvrs, not original nvrs, right?

freshmaker/pyxis.py
```diff
@@ -178,21 +178,30 @@ def get_latest_bundles(self, index_images):
 
         return latest_bundles
 
-    def get_manifest_list_digest_by_nvr(self, nvr):
+    def get_manifest_list_digest_by_nvr(self, nvr, must_be_published=True):
```
qixiang, line 181: maybe `published=True` is clear enough?

freshmaker/handlers/botas/botas_shipped_advisory.py
mprahl: You are right. I misread the code.

freshmaker/pyxis.py
mprahl: I was being cautious so I went the other route, but you're right, there's no reason to be so cautious.

freshmaker/pyxis.py
mprahl: I decided against it because `published=False` insinuates that we don't want those from published repositories, but really, we just want to make the safety checks less strict.

qixiang (approved):
LGTM

mulaievaRH (approved):
+1

PR #95, opened 2021-06-17
WIP: Support manifest v2 digests for manual rebuilds

mprahl:
I ran into an issue when trying the manual bundle rebuilds and the digest in `bundle_path_digest` was a v2 manifest and not a manifest list.

I am marking this as WIP since I don't know if we want to accept this since I'm not realizing that the manual rebuild code relies on the operators/bundles Pyxis API endpoint be and many of these images will be not yet published and thus won't have such an entry.

### DIFF
--- a/freshmaker/pyxis.py
+++ b/freshmaker/pyxis.py
@@ -188,15 +188,17 @@ class Pyxis(object):
 
         return latest_bundles
 
-    def get_manifest_list_digest_by_nvr(self, nvr, must_be_published=True):
+    def get_manifestv2_digests_by_nvr(self, nvr, must_be_published=True):
         """
-        Get image's digest(manifest_list_digest field) by its NVR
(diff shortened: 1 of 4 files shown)


freshmaker/pyxis.py
```diff
@@ -205,9 +207,16 @@ def get_manifest_list_digest_by_nvr(self, nvr, must_be_published=True):
...
+                if 'manifest_list_digest' not in repo:
+                    continue
+
+                mld = repo['manifest_list_digest']
+                msd = None
+                if 'manifest_schema2_digest' in repo:
+                    msd = repo['manifest_schema2_digest']
+
+                return mld, msd
+        return None, None
```
apaplaus, line 219: And what if `manifest_list_digest` is present in the image and `manifest_schema2_digest` isn't and vice versa? We still should return (None None)? As for now, I see that the code(`get_bundles_by_digest`) can work with just `manifest_list_digest` and without schema2.

PR #96, opened 2021-06-21
Set olm.substitutesFor to the CSV name

mprahl:
This was previously set to the version which is incorrect.


mprahl:
FYI, I made a hotfix release to include this fix.

apaplaus (approved):

PR #97, opened 2021-07-02
Add a log message when an invalid semver is encountered

mprahl:


qixiang:
:+1:

freshmaker/pyxis.py
```diff
@@ -169,19 +169,24 @@ def get_latest_bundles(self, index_images):
             request_params['filter'] = \
                 f'latest_in_channel==true and source_index_container_path=={path}'
 
-            def _isvalid(version):
+            def _isvalid(version, csv_name):
```
apaplaus, line 172: so the `csv_name` parameter needed only for printing an error message in case of exception. Seems kind of weird.

mprahl: @apaplaus it's a small helper function, so I wouldn't be too considered if it seems a little weird. That being said, I could pass in the whole bundle instead.

qixiang:
I don't have strong opinion with adding `csv_name` to `_is_valid`, so I just give my +1 and leave the decision to Andrei :)

freshmaker/pyxis.py
apaplaus: Ok, if it isn't a problem for you we can leave it as it is.

mprahl:
Can this be merged now?

apaplaus:
Sure, I will merge it.

PR #98, opened 2021-07-02
Add manual retriggering of bundle rebuild

apaplaus:
Now when there is need in retriggering bundle build due to fail or
something else, there is functionality for it.

Note: it's implemented only to retrigger automatic rebuilds, it's not
for manual rebuild event triggered by BOTAS.

JIRA: CWFHEALTH-603

### DIFF
--- a/freshmaker/events.py
+++ b/freshmaker/events.py
@@ -462,13 +462,33 @@ class BotasErrataShippedEvent(ErrataBaseEvent):
 
 
 class ManualBundleRebuild(BaseEvent):
-    def __init__(self, msg_id, container_images, bundle_images, metadata=None,
-                 dry_run=False, requester=None):
+    """
+    Event triggered when Release Driver requests manual rebuild
+    OR when manual rebuild of bundles requested by person
+    """
+    def __init__(self, msg_id, dry_run=False):
         super().__init__(msg_id, manual=True, dry_run=dry_run)
-        self.container_images = container_images
-        self.bundle_images = bundle_images
-        self.metadata = metadata
-        self.requester = requester
 
     def is_allowed(self, handler, **kwargs):
         return super().is_allowed(handler, ArtifactType.IMAGE, **kwargs)
+
+    @classmethod
+    def from_manual_rebuild_request(cls, msg_id, advisory,
+                                    freshmaker_event_id=None,
+                                    container_images=[], requester=None, **kwargs):
+        event = cls(msg_id, **kwargs)
+        event.advisory = advisory
+        event.freshmaker_event_id = freshmaker_event_id
+        event.container_images = container_images
+        event.requester = requester
+        return event
+
+    @classmethod
+    def from_release_driver_request(cls, msg_id, container_images, bundle_images,
+                                    metadata=None, requester=None, **kwargs):
+        event = cls(msg_id, **kwargs)
+        event.container_images = container_images
+        event.bundle_images = bundle_images
+        event.metadata = metadata
+        event.requester = requester
+        return event
--- a/freshmaker/handlers/botas/botas_shipped_advisory.py
+++ b/freshmaker/handlers/botas/botas_shipped_advisory.py
@@ -91,10 +91,12 @@ class HandleBotasAdvisory(ContainerBuildHandler):
             self.log_info(msg)
             return []
 
-        if isinstance(event, ManualBundleRebuild):
-            bundles_to_rebuild = self._handle_manual_rebuild(db_event)
+        if isinstance(event, ManualBundleRebuild) and \
+                hasattr(event, 'bundle_images'):
+            bundles_to_rebuild = self._handle_release_driver_rebuild(db_event)
+        # automatic rebuild and manual bundle rebuild(triggered by post request)
         else:
-            bundles_to_rebuild = self._handle_auto_rebuild(db_event)
+            bundles_to_rebuild = self._handle_bundle_rebuild(db_event)
 
         if not bundles_to_rebuild:
             return []
@@ -111,15 +113,16 @@ class HandleBotasAdvisory(ContainerBuildHandler):
 
         return []
 
-    def _handle_auto_rebuild(self, db_event):
+    def _handle_bundle_rebuild(self, db_event):
         """
         Handle auto rebuild for an advisory created by Botas
+        OR manually triggered rebuild
 
         :param db_event: database event that represent rebuild event
         :rtype: list
         :return: list of advisories that should be rebuilt
         """
-        # Mapping of original build nvrs to rebuilt nvrs in advisory
+        # Mapping of operators' original build nvrs to rebuilt nvrs in advisory
         nvrs_mapping = self._create_original_to_rebuilt_nvrs_map()
 
         original_nvrs = nvrs_mapping.keys()
@@ -237,9 +240,26 @@ class HandleBotasAdvisory(ContainerBuildHandler):
             if not bundles:
                 self.log_warn('The bundle digest %r was not found in Pyxis. Skipping.', digest)
                 continue
+            bundle_nvr = bundles[0]['brew']['build']
+
+            # If specific container images where requested to rebuild, process only them
+            if (isinstance(self.event, ManualBundleRebuild)
+                    and self.event.container_images                         # noqa: W503
+                    and bundle_nvr not in self.event.container_images):     # noqa: W503
+                self.log_debug("Ignoring '%s', because it's not in requested rebuilds"
+                               " (container_images in request)", bundle_nvr)
+                continue
+
+            # Filter out builds from dependent event that were rebuilt recently
+            done_build = db_event.get_artifact_build_from_event_dependencies(
+                bundle_nvr)
+            if done_build:
+                self.log_debug("Ignoring '%s' bundle, because it was already rebuilt"
+                               " in dependent event", bundle_nvr)
+                continue
 
             bundles_by_digest.setdefault(digest, copy.deepcopy(default_bundle_data))
-            bundles_by_digest[digest]['nvr'] = bundles[0]['brew']['build']
+            bundles_by_digest[digest]['nvr'] = bundle_nvr
             bundles_by_digest[digest]['images'] = bundles
 
         # Unauthenticated koji session to fetch build info of bundles
@@ -345,7 +365,7 @@ class HandleBotasAdvisory(ContainerBuildHandler):
                                       to_rebuild_digests))
         return bundles_to_rebuild
 
-    def _handle_manual_rebuild(self, db_event):
+    def _handle_release_driver_rebuild(self, db_event):
         """
         Handle manual rebuild submitted by Release Driver for an advisory created by Botas
 
@@ -647,7 +667,7 @@ class HandleBotasAdvisory(ContainerBuildHandler):
 
     def _create_original_to_rebuilt_nvrs_map(self):
         """
-        Creates mapping of original build NVRs to rebuilt NVRs in advisory.
+        Creates mapping of original operator build NVRs to rebuilt NVRs in advisory.
         Including NVRs of the builds from the blocking advisories
 
         :rtype: dict
@@ -677,8 +697,8 @@ class HandleBotasAdvisory(ContainerBuildHandler):
                 # all of them, that have overlapping package names
                 for block_build in blocking_advisories_builds:
                     block_build_nvr = parse_nvr(block_build)
-                    if block_build_nvr['name'] == parsed_build_nvr['name'] and \
-                            block_build_nvr['version'] == parsed_build_nvr['version']:
+                    if (block_build_nvr['name'] == parsed_build_nvr['name']
+                            and block_build_nvr['version'] == parsed_build_nvr['version']):     # noqa: W503
                         nvrs_mapping[block_build] = build_nvr
         return nvrs_mapping
 
--- a/freshmaker/parsers/internal/manual_rebuild.py
+++ b/freshmaker/parsers/internal/manual_rebuild.py
@@ -51,14 +51,31 @@ class FreshmakerManualRebuildParser(BaseParser):
             errata = Errata()
             advisory = ErrataAdvisory.from_advisory_id(errata, errata_id)
 
-            event = ManualRebuildWithAdvisoryEvent(
-                msg_id, advisory, data.get("container_images", []), data.get("metadata", None),
-                freshmaker_event_id=data.get('freshmaker_event_id'), manual=True,
-                dry_run=dry_run, requester=data.get('requester', None))
+            # Manually triggered rebuild of bundles, executed by POST request
+            if advisory.state == "SHIPPED_LIVE" and \
+                    advisory.reporter.startswith('botas'):
+                event = ManualBundleRebuild.from_manual_rebuild_request(
+                    msg_id,
+                    advisory,
+                    data.get('freshmaker_event_id', None),
+                    data.get('container_images', []),
+                    requester=data.get('requester', None),
+                    dry_run=dry_run)
+            else:
+                event = ManualRebuildWithAdvisoryEvent(
+                    msg_id,
+                    advisory,
+                    data.get("container_images", []),
+                    data.get("metadata", None),
+                    freshmaker_event_id=data.get('freshmaker_event_id'),
+                    dry_run=dry_run,
+                    requester=data.get('requester', None))
+        # Retriggered rebuild of bundles by Release Driver
         else:
-            event = ManualBundleRebuild(msg_id, data.get('container_images', []),
-                                        data.get('bundle_images'), data.get('metadata', None),
-                                        dry_run=dry_run, requester=data.get('requester', None))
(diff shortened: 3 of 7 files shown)


freshmaker/parsers/internal/manual_rebuild.py
```diff
@@ -51,10 +52,16 @@ def parse_post_data(self, data):
...
 
-            event = ManualRebuildWithAdvisoryEvent(
-                msg_id, advisory, data.get("container_images", []), data.get("metadata", None),
-                freshmaker_event_id=data.get('freshmaker_event_id'), manual=True,
-                dry_run=dry_run, requester=data.get('requester', None))
+            # If we want to manually retrigger bundle rebuild, create
+            # BotasErrataShippedEvent event
+            if advisory.state == "SHIPPED_LIVE" and \
+                    advisory.reporter.startswith('botas'):
+                event = BotasErrataShippedEvent(msg_id, advisory)
```
qixiang, line 59: I believe we need different event to distinguish the original `BotasErrataShippedEvent` and the re-triggered one, so reusing `BotasErrataShippedEvent` is not what we want.

qixiang (commented):
The first thing need to be addressed is Freshmaker should skip the succeed builds in original event.

apaplaus:
> The first thing need to be addressed is Freshmaker should skip the succeed builds in original event.

Yes, probably it would be better to skip successful builds from the previous event. But I am trying to find parts of the code where we are doing this for the manual rebuilds for advisories. And I can't find any neither for ManualRebuildWithAdvisoryEvent nor for FreshmakerAsyncManualBuildEvent. Can you please help me with finding it?

qixiang:
> > The first thing need to be addressed is Freshmaker should skip the succeed builds in original event.
> 
> Yes, probably it would be better to skip successful builds from the previous event. But I am trying to find parts of the code where we are doing this for the manual rebuilds for advisories. And I can't find any neither for ManualRebuildWithAdvisoryEvent nor for FreshmakerAsyncManualBuildEvent. Can you please help me with finding it?

iirc, it's handled by:
https://github.com/redhat-exd-rebuilds/freshmaker/blob/master/freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py#L231-L247
https://github.com/redhat-exd-rebuilds/freshmaker/blob/28723e72c59c026996c03b4019c9489f3d8d90a2/freshmaker/models.py#L497-L510

apaplaus:
@qixiang well I've added the new event, and added filtering out already done builds from the dependent event. I think the new event can be handled exactly the same way as the automatic rebuild, since we are processing the advisory, and the difference will be in ignoring builds from the dependent event. But I am not sure =) So can you take a look again pls?

freshmaker/parsers/internal/manual_rebuild.py
```diff
@@ -51,10 +52,17 @@ def parse_post_data(self, data):
...
+                    advisory.reporter.startswith('botas'):
+                event = RetriggeredBundleRebuild(
+                    msg_id, advisory, data.get('freshmaker_event_id'), dry_run)
+            else:
+                event = ManualRebuildWithAdvisoryEvent(
+                    msg_id, advisory, data.get("container_images", []), data.get("metadata", None),
+                    freshmaker_event_id=data.get('freshmaker_event_id'), manual=True,
+                    dry_run=dry_run, requester=data.get('requester', None))
         else:
             event = ManualBundleRebuild(msg_id, data.get('container_images', []),
```
qixiang, line 67: `ManualBundleRebuild` now doesn't support advisory id, this should be adjusted, release-driver will also need the support of advisory id in `ManaualBundleRebuild` event.

freshmaker/events.py
```diff
@@ -472,3 +473,11 @@ def __init__(self, msg_id, container_images, bundle_images, metadata=None,
 
     def is_allowed(self, handler, **kwargs):
         return super().is_allowed(handler, ArtifactType.IMAGE, **kwargs)
+
+
+class RetriggeredBundleRebuild(ErrataBaseEvent):
```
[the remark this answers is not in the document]
apaplaus: Well when Matt was describing this task he said that this 'retriggering' functionality should work in almost the same way as it works now for manually retriggering failed event of rhsa advisory rebuild. And it works in that way that you can submit post request with the `freshmaker_event_id` of failed event. FM will search for it -> create new db_event with the dependency set to the failed event  + set errata_id from the failed event so we have an advisory -> rebuild images + while rebuilding since it has dependent event it will not rebuild successfully built images from it. Right? So I thought if we want the same functionality(just submit post request with eighter errata_id OR freshmaker_event_id) I should do it like I did. So now if you submit POST request with the `errata_id` or with the `freshmaker_event_id`, it will trigger the rebuild like it handles advisory, and the difference will be that if `freshmaker_event_id` is submitted it will add dependent event and will ignore done builds. Maybe I missed something in implementation, but that's the other question. And about `ManualBundleRebuild`, it's an event triggered exclusively by BOTAS sending the request(maybe it's not the best name for the event then but still), and BOTAS always submits `bundle_images` as well as `container_images` and there is no advisory at all in the processing process.

freshmaker/parsers/internal/manual_rebuild.py
apaplaus: Yes, and it was the part of the implementation, I was pretty sure during the implementation of this event, that there will be no advisory submitted/passed from BOTAS, so BOTAS will submit ONLY bundles + container_images, so I should work with only submitted information.

freshmaker/events.py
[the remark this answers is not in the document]
qixiang: > Well when Matt was describing this task he said that this 'retriggering' functionality should work in almost the same way as it works now for manually retriggering failed event of rhsa advisory rebuild. And it works in that way that you can submit post request with the `freshmaker_event_id` of failed event. FM will search for it -> create new db_event with the dependency set to the failed event + set errata_id from the failed event so we have an advisory -> rebuild images + while rebuilding since it has dependent event it will not rebuild successfully built images from it. Right? So I thought if we want the same functionality(just submit post request with eighter errata_id OR freshmaker_event_id) I should do it like I did. So now if you submit POST request with the `errata_id` or with the `freshmaker_event_id`, it will trigger the rebuild like it handles advisory, and the difference will be that if `freshmaker_event_id` is submitted it will add dependent event and will ignore done builds.

All right.

> Maybe I missed something in implementation, but that's the other question. And about `ManualBundleRebuild`, it's an event triggered exclusively by BOTAS sending the request(maybe it's not the best name for the event then but still), and BOTAS always submits `bundle_images` as well as `container_images` and there is no advisory at all in the processing process.

How about just updating `ManualBundleRebuild` to be triggered by either `errata_id` or `freshmaker_event_id` or `bundle_images`?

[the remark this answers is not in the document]
apaplaus: So you want to have one event to represent two different situation. One- is when botas submits `bundle images` and it handled in one way, and Second - when event submitted manually with the `errata_id/freshmaker_event_id` and then it handled in another way? Seem a little bit like overcomplicating, you will not know what was the trigger of the event just by it type, you will need to check other stuff. For me it's much easier just to maybe rename current event to kind of 'BotasTriggeredRebuild' or something else and use 'ManualBundleRebuild' for the purpose of manually triggered reuilds(with `errata_id/freshmaker_event_id`) What do you think?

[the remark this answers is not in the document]
qixiang: It's release-driver (not botas) triggers the manual rebuild.

They're both manual rebuilds, though one is triggered by a service (release-driver), another one is triggered by human beings, this is same as the existing manual rebuild of rpm advisories, where release-driver requests manual rebuilds with the same event used by us to trigger by manual, though release-driver's request usually contains some additional metadata.

Actually, release-driver can always provide the `errata_id` in request, so you just need to check the advisory (from `errata_id` or `freshmaker_event_id`) to determine whether this should be handled by the botas handler, and in the handler, you need to handle a special request that has `bundle_images` included, what do you think?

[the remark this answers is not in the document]
apaplaus: But I don't need `errata_id` at all when RD triggers rebuild, and now the flow is very different when an event has bundle_images(+container_images) and when an event has errata_id(+ freshamker_event_id)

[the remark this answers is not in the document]
qixiang: I'm thinking of dropping the `bundle_images` in manual rebuild request to simplify the workflow, seems it's possible to achieve the same goal of `bundle_images` + `container_images` with `errata_id` + `container_images`.

With `bundle_images`, freshmaker can get the pulspec overrides from its db, and with `errata_id`, freshmaker can get a mapping of old digests to new digests, we can repeat the construction of pullspec overrides, what do you think?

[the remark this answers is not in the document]
apaplaus: I think that I don't want to do the additional work of reimplementing manual rebuilds triggered by RD in the scope of the current ticket, because it will prolong this ticket more(it's already pretty long-lasting as for me). If we want to change it, I would prefer another ticket so the work is tracked.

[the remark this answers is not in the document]
qixiang: > I think that I don't want to do the additional work of reimplementing manual rebuilds triggered by RD in the scope of the current ticket, because it will prolong this ticket more(it's already pretty long-lasting as for me). If we want to change it, I would prefer another ticket so the work is tracked.

ok, then how about at least reusing the existing `ManualBundleRebuild` and handle it differently when `bundle_images` is provided or not, I can investigate on simplifying the workflow later, but we'll need to do db schema migration and alter the records in db if we have a new event now and then remove it later.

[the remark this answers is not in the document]
apaplaus: Ok, I will try to make it like that. I will ping you again when it will be ready.

freshmaker/events.py
```diff
@@ -462,13 +462,30 @@ def __init__(self, msg_id, advisory, dry_run=False):
...
     def is_allowed(self, handler, **kwargs):
         return super().is_allowed(handler, ArtifactType.IMAGE, **kwargs)
+
+    @classmethod
+    def from_manual_rebuild_request(cls, msg_id, advisory,
+                                    freshmaker_event_id=None, **kwargs):
+        event = cls(msg_id, **kwargs)
+        event.advisory = advisory
+        event.freshmaker_event_id = freshmaker_event_id
+        return event
```
qixiang, line 481: Like the case of manual rebuild of rpm advisory, `event.container_images` should also be supported, in some cases, we just need to rebuild some of the impacted bundle images.

apaplaus:
@qixiang Well I've made some changes, so can you check pls?

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -127,6 +129,14 @@ def _handle_auto_rebuild(self, db_event):
             "Orignial nvrs of build in the advisory #{0} are: {1}".format(
                 self.event.advisory.errata_id, " ".join(original_nvrs)))
 
+        # Ignore builds from dependent events that were rebuilt recently
+        for orig_nvr, rebuilt_nvr in nvrs_mapping.copy().items():
+            done_build = db_event.get_artifact_build_from_event_dependencies(rebuilt_nvr)
```
qixiang, line 134: This is a little confusing, the `nvrs_mapping` is a mapping of images in advisory and their original image NVRs, they're operator or operand images, and the images to skip (`event.container_images`) is the bundle images in the result of images to be rebuilt in this handler, could you correct me if I'm wrong? I believe the skipping feature is only possible to be done after the rebuilt list is figured out and before recording them in db.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -655,33 +665,43 @@ def _create_original_to_rebuilt_nvrs_map(self):
...
-
-                # Check builds from blocking advisories and add to the mapping
-                # all of them, that have overlapping package names
-                for block_build in blocking_advisories_builds:
-                    block_build_nvr = parse_nvr(block_build)
-                    if block_build_nvr['name'] == parsed_build_nvr['name'] and \
-                            block_build_nvr['version'] == parsed_build_nvr['version']:
-                        nvrs_mapping[block_build] = build_nvr
+        # If specific container images where requested to rebuild, process only them
+        if self.event.container_images:
```
qixiang, line 669: Same as above, the `event.container_images` are actually bundle images, and I believe here we're adding these bundle images to the nvr mapping of operator/operand images.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -237,9 +239,24 @@ def _handle_auto_rebuild(self, db_event):
             if not bundles:
                 self.log_warn('The bundle digest %r was not found in Pyxis. Skipping.', digest)
                 continue
+            bundle_nvr = bundles[0]['brew']['build']
+
+            # If specific container images where requested to rebuild, process only them
+            if bundle_nvr in self.event.container_images:
```
qixiang, line 245: it should be `not in`, right?

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -237,9 +239,24 @@ def _handle_auto_rebuild(self, db_event):
...
+                self.log_debug("Ignoring '%s' bundle, because it's not in 'container_images'",
```
qixiang, line 246:
```suggestion
                self.log_debug("Ignoring '%s', because it's not in requested rebuilds (container_images in request)",
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -237,9 +239,24 @@ def _handle_auto_rebuild(self, db_event):
...
+                               bundle_nvr)
+                continue
+
+            # Filter out builds from dependent event that were rebuilt recently
+            done_build = db_event.get_artifact_build_from_event_dependencies(
+                bundle_nvr)
+            if done_build:
+                self.log_debug("Ignoring '%s' bundle, because it was already rebuilt",
```
qixiang, line 254:
```suggestion
                self.log_debug("Ignoring '%s', because it was already rebuilt in dependent event",
```

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -677,8 +694,8 @@ def _create_original_to_rebuilt_nvrs_map(self):
                 # all of them, that have overlapping package names
                 for block_build in blocking_advisories_builds:
                     block_build_nvr = parse_nvr(block_build)
-                    if block_build_nvr['name'] == parsed_build_nvr['name'] and \
-                            block_build_nvr['version'] == parsed_build_nvr['version']:
+                    if block_build_nvr['name'] == parsed_build_nvr['name']\
+                            and block_build_nvr['version'] == parsed_build_nvr['version']:
```
qixiang, line 698: Per PEP8, wrapping in parentheses should be used in preference to use a backslash for line continuation.

```suggestion
                    if (
                        block_build_nvr["name"] == parsed_build_nvr["name"]
                        and block_build_nvr["version"] == parsed_build_nvr["version"]
                    ):
```

freshmaker/models.py
```diff
@@ -66,7 +66,7 @@
     ManualRebuildWithAdvisoryEvent: 13,
     FreshmakerAsyncManualBuildEvent: 14,
     BotasErrataShippedEvent: 15,
-    ManualBundleRebuild: 16,
+    ManualBundleRebuild: 16
```
qixiang, line 69: This change is unnecessary, and per PEP8, a trailing comma is preferred.

freshmaker/parsers/internal/manual_rebuild.py
```diff
@@ -21,7 +21,8 @@
 
 import time
 from freshmaker.parsers import BaseParser
-from freshmaker.events import ManualRebuildWithAdvisoryEvent, ManualBundleRebuild
+from freshmaker.events import ManualRebuildWithAdvisoryEvent, \
+    ManualBundleRebuild
```
qixiang, line 25: This is a unnecessary change.

freshmaker/handlers/botas/botas_shipped_advisory.py
apaplaus: Yes)

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -237,9 +240,25 @@ def _handle_auto_rebuild(self, db_event):
             if not bundles:
                 self.log_warn('The bundle digest %r was not found in Pyxis. Skipping.', digest)
                 continue
+            bundle_nvr = bundles[0]['brew']['build']
+
+            # If specific container images where requested to rebuild, process only them
+            if (isinstance(self.event, ManualBundleRebuild)
+                    and bundle_nvr not in self.event.container_images):     # noqa: W503
```
qixiang, line 247: You need to check if `container_images` is empty or not, otherwise all bundles will be ignored in case of `container_images` is not specified.

apaplaus: fixed

qixiang:
LGTM

PR #101, opened 2021-07-21
further updates to build_id docs

ktdreyer:
Explain that the Koji task that we track in `build_id` is a buildContainer task.

Publicly document that a null `build_id` implies that Freshmaker did not submit the task *or* that Koji did not give a task ID back to Freshmaker.


qixiang:
:+1: thank you

PR #102, opened 2021-07-26
Fix bundle manual rebuild

qixiang:
1. return advisory id as search_key of manual bundle rebuild when it exists
2. store requester_metadata_json data in database


mulaievaRH:
LGTM

mulaievaRH (approved):

PR #103, opened 2021-07-26
Do not generate composes which have already been in original build

qixiang:
When freshmaker rebuids an image, it will generate composes to enable
all content sets in image and submit the build task with the composes.

For example there is an image (with 4 arches) is having content sets:

    aarch64: "rhel-8-for-aarch64-baseos-rpms"
    ppc64le: "rhel-8-for-ppc64le-baseos-rpms"
    s390x:   "rhel-8-for-s390x-baseos-rpms"
    x86_64:  "rhel-8-for-x86_64-baseos-rpms"

Freshmaker will generate a signle ODCS compose with source value:

    "rhel-8-for-aarch64-baseos-rpms rhel-8-for-ppc64le-baseos-rpms
     rhel-8-for-s390x-baseos-rpms rhel-8-for-x86_64-baseos-rpms"

then submits the image build task with this compose included in
"compose_ids", usually this is not a problem for layered images.
However when this image is a base image, the original build will
also has these composes enabled in "compose_ids", freshmaker will
enable the original composes, and also the composes freshmaker
generated to enable all content sets, so there is one additional
compose enabled for each arch and with same content.

This is not a problem for layered images, but it will result in
issue for base images, because there will be base filesystem container
created during the build process, anaconda will download all content
from all enabled repos, for content from each repo, there is memory
occupied, when there are several duplicated repos are enabled, the
image build task can run into lack of memory issue.

The even worse case is when the original image was rebuilt by
Freshmaker, there were already duplicated composes in original build,
there will be duplicated composes added one more time.

In this change, Freshmaker is changed to generate a compose for
each arch, and before that, Freshmaker will check if this compose has
already been in original build's "compose_ids" by comparing the
compose "source" value.

JIRA: CWFHEALTH-646


mulaievaRH (approved):
+1

PR #106, opened 2021-08-05
Remove odcs composes added by freshmaker in original image

qixiang:
When an original image was built by freshmaker, there could be some odcs
composes added by freshmaker in build task, so if there is an image was
rebuilt by freshmaker for several times, the latest original image could
have some duplicated odcs composes in build task. For example:

    1. image-2.1 was built and it has compose 101.
    2. image-2-1 was rebuilt by freshmaker, new compose 102 added, new
       NVR is image-2-1.0001, composes 101 and 102 are enabled in build
       task.
    3. image-2-1.0001 was rebuilt by freshmaker again, new compose 103
       added, new NVR is image-2-1.0002, its build task has composes:
       101, 102 and 103.

Then if freshmaker needs to rebuild image-2-1.0002, it will submit the
build task with composes: 101, 102 and 103. After 4694dda, freshmaker
doesn't add new composes which already exist, but it can't remove the
duplicated composes from original image (image-2-1.0002 in this case).

With this change, freshmaker will only keep the composes from most
original image (in this case it's compose 101 from image-2.1), thus
composes added by freshmaker in original image (image-2-1.0002) are
removed.

JIRA: CWFHEALTH-689

### DIFF
--- a/freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
+++ b/freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
@@ -290,7 +290,7 @@ class RebuildImagesOnRPMAdvisoryChange(ContainerBuildHandler):
                     "target": image["target"],
                     "branch": image["git_branch"],
                     "arches": image["arches"],
-                    "renewed_odcs_compose_ids": image["odcs_compose_ids"],
+                    "renewed_odcs_compose_ids": image["original_odcs_compose_ids"],
                 })
 
                 db.session.commit()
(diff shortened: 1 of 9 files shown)


freshmaker/models.py
```diff
@@ -626,6 +626,23 @@ def get_lowest_build_id(cls, session):
             return 0
         return build.build_id
 
+    @classmethod
+    def get_most_original_nvr(cls, nvr):
```
apaplaus, line 630: Would be nice to have unit test for this method too.

freshmaker/models.py
```diff
@@ -626,6 +626,23 @@ def get_lowest_build_id(cls, session):
...
+        Return the NVR of most original image
+        """
+        original_nvr = None
+        build = db.session.query(ArtifactBuild).filter(cls.rebuilt_nvr == nvr).first()
+        while build:
+            original_nvr = build.original_nvr
+            build = db.session.query(ArtifactBuild).filter(cls.rebuilt_nvr == original_nvr).first()
+        # This can be called for many times while resolving images, close the session
+        # to avoid possible flask-SqlAlchemy QueuePool limit size overflow
+        db.session.close()
```
apaplaus, line 643: I am not, but is it safe to close the session here? If it's used further, will it be opened again automatically or how this works?

freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
```diff
@@ -303,24 +303,33 @@ def _record_batches(self, batches, event, builds=None):
...
+                        if new_pulp_sources:
+                            # Check if the compose for these new pulp sources is
                             # already cached and use it in this case.
+                            cache_key = " ".join(sorted(new_pulp_sources))
                             if cache_key in odcs_cache:
                                 db_compose = odcs_cache[cache_key]
                             else:
                                 compose = self.odcs.prepare_pulp_repo(
-                                    build, arch_content_sets)
+                                    build, list(new_pulp_sources))
```
apaplaus, line 332: I don't understand composes completely, but if we are creating new compose with only NEW pulp sources, so the old ones will stay or they will be replaced by new ones or how this works?

freshmaker/models.py
qixiang: Added.

freshmaker/models.py
qixiang: SQLAlchemy session is a factory, a new session will be created automatically when the old one is closed. But yes, there is a risk if someone call this API after some instance is attached to the session but before commit, this can result in issues, so probably it's more safe to increase the SQLAlchemy pool size or even disabling pooling.
So I've removed this and opened an internal MR to change the pool size for our deployment. I didn't add that in default config because sqlite is used by default, which is using `NullPool`, it doesn't have pooling and doesn't accept `pool_size` as engine option.

freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
qixiang: We only generate new pulp composes for the contest sets which don't exist in original odcs composes, the original odcs composes will be renewed, so in the image build task, the composes are old ones and the new ones added here, the purpose of this PR is to remove the duplicated repos in old composes, we already has 4694dda to not adding duplicated repos, but we can't remove the existing duplicated repos without this PR. Actually this is a temporary solution and probably can be removed after a while when there is no latest published images are having duplicated repos enabled.

Check:
https://github.com/redhat-exd-rebuilds/freshmaker/blob/040cd4b8f4feb4d7607e47d20bd106ef6c7c7e01/freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py#L293
Note: the `odcs_compose_ids` has been changed to `original_odcs_compose_ids` in this PR.

apaplaus: Thx for explanation

apaplaus (approved):

PR #109, opened 2021-08-25
Retry LightBlue query when searching for images

apaplaus:
This quickfix affects error when connection to LightBlue is closed
during the request for some reason. Now request will be repeated

JIRA: CWFHEALTH-716


qixiang (approved):
LGTM

PR #112, opened 2021-08-31
Set alembic version to <1.7

qixiang:
alembic >= 1.7 is not compatible with the migrations, this fixes
the mypy test failure:

    freshmaker/migrations/env.py:9: error: Module has no attribute "config"; maybe "configure"?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+alembic<1.7
 fedmsg
 moksha.hub
 psutil


apaplaus:
Maybe we should update the code to support a newer version of alembic?

qixiang:
> Maybe we should update the code to support a newer version of alembic?

The latest version of  alembic in fedora 34 is 1.5.5, and we're even using fedora 32. Tox uses pip to install the requirements, so it has much newer versions.

But maybe we should update our image to use fedora 34, fedora 32 is a little old.

apaplaus:
> > Maybe we should update the code to support a newer version of alembic?
> 
> The latest version of alembic in fedora 34 is 1.5.5, and we're even using fedora 32. Tox uses pip to install the requirements, so it has much newer versions.
> 
> But maybe we should update our image to use fedora 34, fedora 32 is a little old.

Yes, I agree on updating to a newer Fedora image, I think it might be beneficial for us in the future.

qixiang:
> Yes, I agree on updating to a newer Fedora image, I think it might be beneficial for us in the future.

An internal card filed CWFHEALTH-775 for it.

apaplaus:
> An internal card filed CWFHEALTH-775 for it.
Thanks, so I think we can drop this one then? Or if you want to merge it it's fine.

apaplaus (approved):

qixiang:
> > Thanks, so I think we can drop this one then? Or if you want to merge it it's fine.

We need this because currently the mypy tests fails: https://github.com/redhat-exd-rebuilds/freshmaker/pull/111/checks
I'm going to merge it.

PR #113, opened 2021-09-01
Disallow multiple events

ElenaKarolinaSemanova:
Check if there is already another event running for the same advisory.

### DIFF
--- a/freshmaker/views.py
+++ b/freshmaker/views.py
@@ -399,6 +399,9 @@ def _validate_rebuild_request(request):
     if not isinstance(data.get('dry_run', False), bool):
         return json_error(400, 'Bad Request', '"dry_run" must be a boolean.')
 
+    if not isinstance(data.get('force', False), bool):
+        return json_error(400, 'Bad Request', '"force" must be a boolean.')
+
     return None
 
 
@@ -553,6 +556,20 @@ class BuildAPI(MethodView):
                     'The provided "errata_id" doesn\'t match the Advisory ID associated with the '
                     'input "freshmaker_event_id".',
                 )
+
+        if data.get("errata_id") or data.get("freshmaker_event_id"):
+            running_events = (
+                models.Event.query.filter(models.Event.search_key == str(data['errata_id'])).filter(
+                    models.Event.state.in_(
+                        [EventState.INITIALIZED.value, EventState.BUILDING.value])).all())
+
+            if running_events and not data.get('force', False):
+                event_ids = [e.id for e in running_events]
+                return json_error(
+                    400,
+                    'Bad Request',
+                    f'Events triggered by advisory {data["errata_id"]} are running: {event_ids}. '
+                    f'If you want to rebuild it anyway, use "force": true option.')
 
         # Use the shared code to parse the POST data and generate right
         # event based on the data.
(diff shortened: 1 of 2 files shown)


ElenaKarolinaSemanova:
@qixiang Can you please take a look if this is what you meant in the ticket?
https://issues.redhat.com/browse/CWFHEALTH-688

freshmaker/views.py
```diff
@@ -386,6 +385,24 @@ def _validate_rebuild_request(request):
...
+        event = models.Event.get_by_event_id(db.session, freshmaker_event_id)
+        event_state = event.state
+        if event_state in (0, 1):
+            return json_error(400, 'Bad Request', f'"{event.id}" is in Initialized or Building state. If you want to continue, use "force": True option.')
+
+    if not data.get("force", None):
+        if 'freshmaker_event_id' in data:
+            _event_state_check(data.get('freshmaker_event_id'))
+        if 'errata_id' in data:
+            events = []  # Todo
```
ElenaKarolinaSemanova, line 398: @qixiang Is there any method or function which would return a list of events by errata_id.

qixiang: `search_key` in `Event` model is the errata id.

We don't need to check against `freshmaker_event_id` and `errata_id` separately since errata id is available in event, a better place to perform such a check is before:
https://github.com/redhat-exd-rebuilds/freshmaker/blob/c8807d784990b058347ca22e56d839a4718e6d8f/freshmaker/views.py#L556-L559

with something like:
```
running_events = (
    models.Event.query.filter(models.Event.search_key == str(data['errata_id']))
    .filter(models.Event.state.in_([EventState.INITIALIZED.value, EventState.BUILDING.value]))
    .all()
)
```

freshmaker/views.py
```diff
@@ -386,6 +385,24 @@ def _validate_rebuild_request(request):
...
+            for event in events:
+                _event_state_check(event)
+        if 'bundle_images' in data:
+            events = []  # Todo
```
ElenaKarolinaSemanova, line 402: @qixiang Is there any method or function which would return a list of events by bundle_images?

qixiang: hm, you can ignore the bundle case at this moment, the feature actually doesn't work as expected, we're working on fixing this, sorry for didn't clarify this before.

ElenaKarolinaSemanova:
@qixiang Please review.

freshmaker/views.py
```diff
@@ -554,6 +557,19 @@ def post(self):
...
+                models.Event.query.filter(models.Event.search_key == str(data['errata_id'])).filter(
+                    models.Event.state.in_(
+                        [EventState.INITIALIZED.value, EventState.BUILDING.value])).all())
+
+            if running_events and not data.get('force', False):
+                return json_error(
+                    400,
+                    'Bad Request',
+                    f'Some freshmaker event with errata_id {data["errata_id"]} is in INITIALIZED or '
+                    f'BUILDING state. If you want to rebuild it anyway, use "force": True option.')
```
qixiang, line 571:
```suggestion
                event_ids = [e.id for e in running_events]
                return json_error(
                    400,
                    'Bad Request',
                    f'Events triggered by advisory {data["errata_id"]} are running: {event_ids}. '
                    f'If you want to rebuild it anyway, use "force": true option.')
```

qixiang (commented):
Looks good to me, just minor comments.

ElenaKarolinaSemanova:
@qixiang I fixed everything. Please review again.

qixiang:
LGTM

PR #114, opened 2021-09-10
Don't run the handlers for unrelated messages

ElenaKarolinaSemanova:
Don't run the handlers for messages of unrelated odcs composes or brew tasks
(CWFHEALTH-706)

### DIFF
--- a/freshmaker/handlers/internal/update_db_on_odcs_compose_fail.py
+++ b/freshmaker/handlers/internal/update_db_on_odcs_compose_fail.py
@@ -39,14 +39,20 @@ class UpdateDBOnODCSComposeFail(BaseHandler):
     def can_handle(self, event):
         if not isinstance(event, ODCSComposeStateChangeEvent):
             return False
+
+        compose_id = event.compose["id"]
+
+        # check db to see whether this compose exists in db
+        found_compose = Compose.query.filter_by(odcs_compose_id=compose_id).first()
+
+        if not found_compose:
+            return False
         return event.compose["state"] == COMPOSE_STATES["failed"]
 
     @fail_event_on_handler_exception
     def handle(self, event):
 
         compose_id = event.compose["id"]
-        if Compose.query.filter_by(odcs_compose_id=compose_id).first() is None:
-            return
 
         self.log_error("ODCS compose %s failed", compose_id)
 
--- a/freshmaker/handlers/koji/rebuild_images_on_odcs_compose_done.py
+++ b/freshmaker/handlers/koji/rebuild_images_on_odcs_compose_done.py
@@ -39,6 +39,14 @@ class RebuildImagesOnODCSComposeDone(ContainerBuildHandler):
     def can_handle(self, event):
         if not isinstance(event, ODCSComposeStateChangeEvent):
             return False
+
+        compose_id = event.compose['id']
+
+        # check db to see whether this compose exists in db
+        found_compose = Compose.query.filter_by(odcs_compose_id=compose_id).first()  # db.session.query(ArtifactBuild).filter_by(odcs_compose_id=compose_id).first()
+
+        if not found_compose:
+            return False
         return event.compose['state'] == COMPOSE_STATES['done']
 
     @fail_event_on_handler_exception
@@ -47,8 +55,6 @@ class RebuildImagesOnODCSComposeDone(ContainerBuildHandler):
             self.force_dry_run()
 
         compose_id = event.compose['id']
-        if Compose.query.filter_by(odcs_compose_id=compose_id).first() is None:
-            return
 
         self.log_info('ODCS compose %s finished', compose_id)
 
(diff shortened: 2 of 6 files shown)


freshmaker/handlers/koji/rebuild_images_on_parent_image_build.py
```diff
@@ -42,6 +42,17 @@ class RebuildImagesOnParentImageBuild(ContainerBuildHandler):
...
+
+        # check db to see whether this build exists in db
+        found_build = db.session.query(ArtifactBuild).filter_by(
+            type=ArtifactType.IMAGE.value,
+            build_id=build_id
+        ).first()
+
+        if not found_build:
+            return False
+
```
ElenaKarolinaSemanova, line 55: @qixiang I have a question about this part of code. I am not sure about this solution, because it is duplicated below in method handle. We can not delete this part of code in handle method, because the variable found_builds is used there as well.

qixiang: we can remove this check in `handle` method since the message will always be related after the `can_handle` check.

freshmaker/handlers/koji/rebuild_images_on_parent_image_build.py
```diff
@@ -61,13 +72,12 @@ def handle(self, event):
...
-                                               EventState.BUILDING.value]:
-                return
-            self.update_db_build_state(build_id, found_build, event)
-            self.rebuild_dependent_containers(found_build)
+        self.set_context(found_build)
+        if found_build.event.state not in [EventState.INITIALIZED.value,
+                                           EventState.BUILDING.value]:
+            return
+        self.update_db_build_state(build_id, found_build, event)
+        self.rebuild_dependent_containers(found_build)
```
mulaievaRH, line 80: @qixiang , in the above comment @ElenaKarolinaSemanova meant we use the `found_build` variable here. So we can't remove from the `handler` method this variable.

qixiang: ah, I see, we can keep it duplicated or set `self.found_build` in `can_handle`, and then in `handle` do something like:
```
found_build = getattr(self, 'found_build', None)
if found_build is None:
    ...
```
personally I think keep this duplicated as now is simple and good enough.

freshmaker/handlers/internal/update_db_on_odcs_compose_fail.py
```diff
@@ -37,16 +37,19 @@ class UpdateDBOnODCSComposeFail(BaseHandler):
     order = 0
 
     def can_handle(self, event):
-        if not isinstance(event, ODCSComposeStateChangeEvent):
```
qixiang, line 40: You need to keep this check before checking compose id, other events may not have compose in event at all.

freshmaker/handlers/koji/rebuild_images_on_odcs_compose_done.py
```diff
@@ -37,7 +37,12 @@ class RebuildImagesOnODCSComposeDone(ContainerBuildHandler):
     """Start image rebuild with this compose containing included packages"""
 
     def can_handle(self, event):
-        if not isinstance(event, ODCSComposeStateChangeEvent):
```
qixiang, line 40: As above, this check need to performed first.

freshmaker/handlers/koji/rebuild_images_on_parent_image_build.py
```diff
@@ -42,6 +42,17 @@ class RebuildImagesOnParentImageBuild(ContainerBuildHandler):
     name = 'RebuildImagesOnParentImageBuild'
 
     def can_handle(self, event):
+        build_id = event.task_id
```
qixiang, line 45: As above, need to check the event type first.

ElenaKarolinaSemanova:
@qixiang @mulaieva When you have time, please review again.

qixiang (approved):
forgot to commit my review comments. lgtm, added a minor suggestion.

freshmaker/handlers/koji/rebuild_images_on_odcs_compose_done.py
```diff
@@ -39,6 +39,14 @@ class RebuildImagesOnODCSComposeDone(ContainerBuildHandler):
     def can_handle(self, event):
         if not isinstance(event, ODCSComposeStateChangeEvent):
             return False
+
+        compose_id = event.compose['id']
+
+        # check db to see whether this compose exists in db
+        found_compose = Compose.query.filter_by(odcs_compose_id=compose_id).first()  # db.session.query(ArtifactBuild).filter_by(odcs_compose_id=compose_id).first()
```
mulaievaRH, line 46:
```suggestion
        found_compose = Compose.query.filter_by(odcs_compose_id=compose_id).first() 
```

PR #116, opened 2021-09-17
Check bundle image's OpenShift versions range

qixiang:
If original bundle image has invalid OpenShift versions range, build
system still can rebuild it, but the rebuilt image will be invalid and
can not be shipped. Before submit the build task, freshmaker can check
the OpenShift versions range, in case it has an invalid value,
freshmaker can just move the build to FAILED state and not submit the
build task.

JIRA: CWFHEALTH-701

### DIFF
--- a/freshmaker/handlers/__init__.py
+++ b/freshmaker/handlers/__init__.py
@@ -27,12 +27,12 @@ import re
 import copy
 from functools import wraps
 
-from freshmaker import conf, log, db, models
+from freshmaker import conf, log, db, models, events
 from freshmaker.kojiservice import koji_service, parse_NVR
 from freshmaker.models import ArtifactBuildState
 from freshmaker.types import EventState
 from freshmaker.models import ArtifactBuild, Event
-from freshmaker.utils import get_rebuilt_nvr
+from freshmaker.utils import get_rebuilt_nvr, is_valid_ocp_versions_range
 from freshmaker.errors import UnprocessableEntity, ProgrammingError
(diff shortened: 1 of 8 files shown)


freshmaker/handlers/__init__.py
```diff
@@ -510,6 +510,24 @@ def build_image_artifact_build(self, build, repo_urls=None):
                 "Container image does not have original_nvr set.")
             return
 
+        # If this is a bundle rebuild, check original build's OpenShift versions
+        # range, if its value is invalid, the bulid ystem can still rebuild it,
```
mulaievaRH, line 514:
```suggestion
        # range, if its value is invalid, the build system can still rebuild it,
```

mulaievaRH (approved):

freshmaker/utils.py
```diff
@@ -217,3 +217,23 @@ def get_ocp_release_date(ocp_version):
...
+    """ Check if an ocp_versions_range string is valid
+
+    :param str ocp_versions_range: the OpenShift versions range value
+    :return: True if the OpenShift versions range is valid, otherwise False
+    :rtype: bool
+    """
+    # Commas are generally not allowed in ocp versions range, and should not be used.
+    # For historical reasons, there are two special values that are currently allowed
+    # to contain commas, "v4.5,v4.6" and "v4.6,v4.5".
+    valid_commas_ranges = ["v4.5,v4.6", "v4.6,v4.5"]
```
apaplaus, line 232: Maybe it should not be hardcoded and we want to add config for this? what if it will change in the future?

apaplaus (approved):

PR #117, opened 2021-09-22
Add support for Flatpak runtimes

hluk:
JIRA: RHELWF-3011


freshmaker/handlers/__init__.py
```diff
@@ -582,6 +586,7 @@ def build_image_artifact_build(self, build, repo_urls=None):
         return self.build_container(
             scm_url, branch, target,
             repo_urls=repo_urls,
+            flatpak=flatpak,
             isolated=True,
```
qixiang, line 590: By checking some old flatpak runtime images, seems these flatpak images need to be built with "isolated=False".

hluk: OK, I added support for fetching the "isolated" parameter from build.

freshmaker/handlers/__init__.py
```diff
@@ -544,6 +546,9 @@ def build_image_artifact_build(self, build, repo_urls=None):
         else:
             parent = args["original_parent"]
 
+        flatpak = args.get("flatpak", False)
+        isolated = args.get("isolated", True)
```
qixiang, line 550: While submitting the build task, freshmaker gets the build args from `build_args` in database, so we need to store `flatpak` and `isolated` in database first, it was not stored before because freshmaker doesn't need to use different `flatpak` or `isolated` params before enabling this.

The `build_args` is saved into db in https://github.com/redhat-exd-rebuilds/freshmaker/blob/809b435d7cab1907eb74ecd898693835a92db9d8/freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py#L297-L307

qixiang (approved):
LGTM

PR #121, opened 2021-10-12
Add event to rebuild Flatpak applications

hluk:
JIRA: RHELWF-3777

### DIFF
--- a/freshmaker/errata.py
+++ b/freshmaker/errata.py
@@ -124,6 +124,26 @@ class ErrataAdvisory(object):
             product_data["product"]["short_name"], cve_list,
             has_hightouch_bug)
 
+    def is_flatpak_module_advisory_ready(self):
+        """ Returns True only if a Flatpaks can be rebuilt from module advisory.
+
+        Flatpaks can be rebuilt only if all of the following are true:
+        - Advisory must contain modules.
+        - Advisory is in QE state.
+        - All attached builds are only shipped to hidden pulp repositories.
+        - All attached builds are signed.
+        """
+        errata = Errata()
+        return (
+            self.state == "QE" and
+            "module" in self.content_types and
+            all(
+                "-hidden-" in repo_id
+                for repo_id in errata.get_pulp_repository_ids(self.errata_id)
+            ) and
+            errata.builds_signed(self.errata_id)
+        )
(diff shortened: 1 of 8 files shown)


freshmaker/errata.py
```diff
@@ -124,6 +124,25 @@ def from_advisory_id(cls, errata, errata_id):
...
+        - Advisory is in QE state.
+        - All attached builds are only shipped to hidden pulp repositories.
+        - All attached builds are signed.
+        """
+        errata = Errata()
+        return (
+            self.state == "QE" and
+            "module" in self.content_types and
+            all(
+                "-hidden-" in repo_id for repo_id in errata.get_pulp_repository_ids()
```
qixiang, line 141: `errata.get_pulp_repository_ids` call is missing the advisory id.

freshmaker/errata.py
```diff
@@ -124,6 +124,25 @@ def from_advisory_id(cls, errata, errata_id):
...
+            ) and
+            errata.builds_signed(self.errata_id)
```
qixiang, line 143: Interesting, I was not aware there is already a `builds_signed` api, maybe we can use update it to the new errata API (RHELWF-4641) which will be released soon, it will be more efficient because it doesn't need to call errata API for each build. However this can be optional change in this PR, the old `builds_signed` check is actually unnecessary and the related event and handler can be refactored (mentioned in CLOUDWF-6150).

freshmaker/errata.py
```diff
@@ -124,6 +124,25 @@ def from_advisory_id(cls, errata, errata_id):
...
+        )
```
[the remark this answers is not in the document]
hluk: That would be nice, the problem is someone already selected the other "incorrect" style preferring operators at the end of the line. :)

freshmaker/errata.py
hluk: Ouch. Fixed. I've added `autospec=True` to the `patch()` calls in tests to verify the arguments.

freshmaker/errata.py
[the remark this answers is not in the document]
qixiang: You're right.

qixiang (approved):

PR #127, opened 2021-10-27
Freshmaker Flatpak Handler

mvalik:
JIRA: RHELWF-3778


qixiang:
> > @mvalik can you move the test file `tests/test_flatpaks.py` to `tests/handlers/koji/test_rebuild_flatpak_application_on_module_ready.py`?
> 
> Not sure if it's good idea, cause It actually tests both handler and event for flatpak

It's fine, at least we currently put the tests related to a specific handler under `tests/handlers`.

mvalik:
Updated, @qixiang please merge if you are ok

PR #129, opened 2021-11-03
Add method to get module builds nvr in advisory

chuazhan:
JIRA: RHELWF-3779

### DIFF
--- a/freshmaker/errata.py
+++ b/freshmaker/errata.py
@@ -495,6 +495,28 @@ class Errata(object):
 
         return list(nvrs)
 
+    def get_cve_affected_build_nvrs(self, errata_id, only_modules=False):
+        """ Get module build NVRs which are affected by the CVEs in errata
+
+        :param errata_id: Errata advisory ID, e.g. 25713.
+        :type errata_id: str or int
+        :param only_modules: only return module builds NVR if True.
+        :type only_modules: bool
+        :return: a list of strings each of them is a build NVR
+        :rtype: list
+        """
+        data = self._errata_rest_get(f"/erratum/{errata_id}/builds_by_cve")
+        nvrs = set()
+        for data_by_product in data.values():
+            for product_data in data_by_product.values():
+                for build in product_data.get("builds", []):
+                    for build_info in build.values():
+                        if only_modules and not build_info.get("is_module"):
+                            continue
+                        nvrs.add(build_info.get("nvr"))
+
+        return list(nvrs)
+
     def get_blocking_advisories_builds(self, errata_id):
         """ Get all advisories that block given advisory id, and fetch all builds from it
 
--- a/tests/test_errata.py
+++ b/tests/test_errata.py
@@ -422,6 +422,86 @@ class TestErrata(helpers.FreshmakerTestCase):
         ret = self.errata.get_cve_affected_rpm_nvrs(28484)
         self.assertEqual(ret, ['libntirpc-1.4.3-4.el6rhs'])
 
+    @patch.object(Errata, "_errata_rest_get")
+    @patch.object(Errata, "_errata_http_get")
+    def test_errata_get_cve_affected_build_nvrs(self, errata_http_get, errata_rest_get):
+        mocked_errata = MockedErrataAPI(errata_rest_get, errata_http_get)
+        ret = self.errata.get_cve_affected_build_nvrs(28484)
+        self.assertEqual(ret, ['libntirpc-1.4.3-4.el6rhs'])
+        ret = self.errata.get_cve_affected_build_nvrs(28484, True)
+        self.assertEqual(ret, [])
+        mocked_errata.builds_by_cve = {
+            "CVE-2021-42574": {
+                "PRODUCT1": {
+                    "builds": [
+                        {
+                            "rust-toolset-rhel8-8050020211027231136.5c15747c": {
+                                "id": 1775730,
+                                "is_module": True,
+                                "nevr": "rust-toolset-0:rhel8-8050020211027231136.5c15747c",
+                                "nvr": "rust-toolset-rhel8-8050020211027231136.5c15747c",
+                                "variant_arch": {
+                                    "PRODUCT1": {
+                                        "SRPMS": [
+                                            "rust-1.54.0-3.module+el8.5.0+13074+d655d86c.src.rpm",
+                                            "rust-toolset-1.54.0-1.module+el8.5.0+12195+effd8a03.src.rpm"
+                                        ],
+                                        "aarch64": [
+                                            "cargo-1.54.0-3.module+el8.5.0+13074+d655d86c.aarch64.rpm",
+                                            "cargo-debuginfo-1.54.0-3.module+el8.5.0+13074+d655d86c.aarch64.rpm",
+                                            "clippy-1.54.0-3.module+el8.5.0+13074+d655d86c.aarch64.rpm",
+                                            "clippy-debuginfo-1.54.0-3.module+el8.5.0+13074+d655d86c.aarch64.rpm",
+                                            "rls-1.54.0-3.module+el8.5.0+13074+d655d86c.aarch64.rpm",
+                                            "rls-debuginfo-1.54.0-3.module+el8.5.0+13074+d655d86c.aarch64.rpm",
+                                            "rust-1.54.0-3.module+el8.5.0+13074+d655d86c.aarch64.rpm",
+                                            "rust-analysis-1.54.0-3.module+el8.5.0+13074+d655d86c.aarch64.rpm",
+                                            "rust-debuginfo-1.54.0-3.module+el8.5.0+13074+d655d86c.aarch64.rpm",
+                                            "rust-debugsource-1.54.0-3.module+el8.5.0+13074+d655d86c.aarch64.rpm",
+                                            "rust-doc-1.54.0-3.module+el8.5.0+13074+d655d86c.aarch64.rpm",
+                                            "rust-std-static-1.54.0-3.module+el8.5.0+13074+d655d86c.aarch64.rpm",
+                                            "rust-std-static-wasm32-unknown-unknown-1.54.0-3.module+el8.5.0+13074+d655d86c.aarch64.rpm",
+                                            "rust-toolset-1.54.0-1.module+el8.5.0+12195+effd8a03.aarch64.rpm",
+                                            "rustfmt-1.54.0-3.module+el8.5.0+13074+d655d86c.aarch64.rpm",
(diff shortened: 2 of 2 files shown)


chuazhan:
@qixiang @hluk @mvalik Can you help to review? Thanks!

freshmaker/errata.py
```diff
@@ -495,6 +495,27 @@ def get_cve_affected_rpm_nvrs(self, errata_id):
...
+        data = self._errata_rest_get(f"/erratum/{errata_id}/builds_by_cve")
+        nvrs = set()
+        for data_by_product in data.values():
+            for product_data in data_by_product.values():
+                for build in product_data.get("builds", []):
+                    for build_info in build.values():
+                        if not build_info.get("is_module"):
+                            continue
+                        nvr = build_info.get("nvr")
+                        nvrs.update([nvr])
```
qixiang, line 515:
```suggestion
                        nvrs.add(nvr)
```

tests/test_errata.py
```diff
@@ -422,6 +422,82 @@ def test_errata_get_cve_affected_rpm_nvrs(self, errata_http_get, errata_rest_get
         ret = self.errata.get_cve_affected_rpm_nvrs(28484)
         self.assertEqual(ret, ['libntirpc-1.4.3-4.el6rhs'])
 
+    @patch.object(Errata, "_errata_rest_get")
+    @patch.object(Errata, "_errata_http_get")
+    def test_errata_get_cve_affected_build_nvrs(self, errata_http_get, errata_rest_get):
+        mocked_errata = MockedErrataAPI(errata_rest_get, errata_http_get)
+        mocked_errata.builds_by_cve = {
+            "CVE-2021-42574": {
```
qixiang, line 430: I'd suggest using some fake data instead of these real data which may have some security risks.

freshmaker/errata.py
```diff
@@ -495,6 +495,27 @@ def get_cve_affected_rpm_nvrs(self, errata_id):
...
+        :return: a list of strings each of them is a module build nvr
+        :rtype: list
+        """
+        data = self._errata_rest_get(f"/erratum/{errata_id}/builds_by_cve")
+        nvrs = set()
+        for data_by_product in data.values():
+            for product_data in data_by_product.values():
+                for build in product_data.get("builds", []):
+                    for build_info in build.values():
+                        if not build_info.get("is_module"):
```
qixiang, line 512: Not sure why my comment got lost in previous review comments, re-adding.

This API actually returns only module builds, it's misleading, you should change the API name to indicate that, or I'd suggest something like this:
```
def get_cve_affected_build_nvrs(self, errata_id, is_module=False):
    ...
    if is_module and not build_info.get("is_module"):
        continue
    ...
```

hluk: Wouldn't the condition be `is_module == build_info.get("is_module", False)` instead? Otherwise it's better to name the argument `only_modules` since `only_modules is False` would mean "both non-modules and module".

qixiang: yes, `only_modules` is a better arg name than `is_module`.

freshmaker/errata.py
chuazhan: Sure. I was imitating an existing method. add() is good enough.

freshmaker/errata.py
chuazhan: Actually I was thinking to add module in the method name, I must make some mistake. But changing in this way is better.

tests/test_errata.py
chuazhan: Good suggestion, fixing.

qixiang (approved):
:+1:

freshmaker/errata.py
```diff
@@ -495,6 +495,28 @@ def get_cve_affected_rpm_nvrs(self, errata_id):
...
+        :return: a list of strings each of them is a build NVR
+        :rtype: list
+        """
+        data = self._errata_rest_get(f"/erratum/{errata_id}/builds_by_cve")
+        nvrs = set()
+        for data_by_product in data.values():
+            for product_data in data_by_product.values():
+                for build in product_data.get("builds", []):
+                    for build_info in build.values():
+                        if only_modules and not build_info.get("is_module"):
```
hluk, line 514: Let me return to this. Is this even needed? Isn't it better to check all attached builds? Why would a module advisory have attached anything else than module builds (and vice versa for non-module advisories)?

qixiang: The reason is this is a generic API, others may use this API to get build NVRs from other type of advisories, another choice is renaming this to something like `get_cve_affected_module_nvrs` and the caller should make sure they only call this API against module advisories.
But I'd prefer the current generic API implementation, and there is only `is_module` in Errata's API result, no `is_rpm`, `is_docker` exist, that's why we only need to support `only_modules`

hluk: My question is: Why not always check all attached builds? I don't understand why some builds should be ignored in a module advisory.

qixiang: I see your concern now, I don't know whether there could be other content other than modules exist in a module advisory, so this is the safest way we can get all modules from a module advisory.

hluk (approved):
:+1:

freshmaker/errata.py
```diff
@@ -495,6 +495,28 @@ def get_cve_affected_rpm_nvrs(self, errata_id):
...
+                            continue
+                        nvrs.add(build_info.get("nvr"))
+
+        return list(nvrs)
```
mvalik, line 518: more python (and quick) way to do exactly the same thing...    
```
return list({build_info.get("nvr")
        for build_info in build.values()
        for build in product_data.get("builds", [])
        for product_data in data_by_product.values()
        for data_by_product in data.values()
        if not only_modules or build_info.get("is_module")})
```

freshmaker/errata.py
```diff
@@ -495,6 +495,28 @@ def get_cve_affected_rpm_nvrs(self, errata_id):
...
 
+    def get_cve_affected_build_nvrs(self, errata_id, only_modules=False):
+        """ Get module build NVRs which are affected by the CVEs in errata
+
+        :param errata_id: Errata advisory ID, e.g. 25713.
+        :type errata_id: str or int
+        :param only_modules: only return module builds NVR if True.
+        :type only_modules: bool
+        :return: a list of strings each of them is a build NVR
+        :rtype: list
```
mvalik, line 506: Does it really need to return a list? Please consider if `set` wouldn't be enough.

tests/test_errata.py
```diff
@@ -422,6 +422,82 @@ def test_errata_get_cve_affected_rpm_nvrs(self, errata_http_get, errata_rest_get
         ret = self.errata.get_cve_affected_rpm_nvrs(28484)
         self.assertEqual(ret, ['libntirpc-1.4.3-4.el6rhs'])
 
+    @patch.object(Errata, "_errata_rest_get")
+    @patch.object(Errata, "_errata_http_get")
+    def test_errata_get_cve_affected_build_nvrs(self, errata_http_get, errata_rest_get):
+        mocked_errata = MockedErrataAPI(errata_rest_get, errata_http_get)
+        mocked_errata.builds_by_cve = {
```
mvalik, line 429: I would rather make short excessive lists of rpms, but use more items in builds and the main dictionary. They can also have different values of 'is_module', so you can test both calls with the second argument `True` and `False`.

mvalik (changes requested):

freshmaker/errata.py
chuazhan: Thanks @mvalik for your review and suggestion!
There are two reasons to write this method in this way:
1. This method is imitating from previous one, we want to make them consistent. So no matter the looping style and  the return value, they are the same. If we change the return value from list to set, I am wondering if the user will be confused that similar method return different type of value.
2. Second,  I like one line style too, it is more pythonic. I think if the data structure is simple, I prefer this way, but this build data has several embedded layers, I think it is better to put them in loops because it is easy to debug and maintain.

tests/test_errata.py
chuazhan: This test data is tweaking from the real data, actually I am not sure about all the build data format, it will cost some time for me to find a build with both module and other content types(I don't even know if that really exists or not).  So I only focus on module build data and its testing. I think if we really meet this use case, we can add more test cases in future.

freshmaker/errata.py
chuazhan: Thanks @mvalik , please check it in above reply.

mvalik: If it's an overwritten method of the base class then it's OK

tests/test_errata.py
mvalik: I didn't mean that you should run all the data in one run. You can make a couple of runs with different samples of data if you want. In the unit tests, you should cover as many branches as possible. It doesn't really matter if your test data "really exists" or not.

mvalik (changes requested):

freshmaker/errata.py
mvalik: For an unknown reason I don't see the 'Resolve' button, please consider this particular thread as resolved.

freshmaker/errata.py
chuazhan: Performance is also a good point. Can we think about this potential performance issue in another patch, I'd like to deal it together with another method "get_cve_affected_rpm_nvrs", they are similar methods and  they all need to be improved.

chuazhan: @qixiang @hluk what about your suggestions? Welcome to comment.

tests/test_errata.py
chuazhan: Yes. I agree that test coverage is important, for this patch, it is simple to add and I am fine, but I am still wondering if that makes sense to add a non-existing case with lots of effort to mock it. Just want to align everyone on the same page @qixiang @hluk

freshmaker/errata.py
qixiang: @mvalik @chuazhan list comprehension is more pythonic, but when there are too many levels or the expression becomes too long, I'd prefer to avoid list comprehension for better code readability and maintainability.

tests/test_errata.py
chuazhan: Tests updated

chuazhan:
Test case updated

tests/test_errata.py
mvalik: Thank you @chuazhan

freshmaker/errata.py
mvalik: Since the performance impact is not that big in this case, we can keep it as loops.

mvalik (approved):

mvalik:
+1

freshmaker/errata.py
chuazhan: Cool. Thanks @mvalik

PR #119, opened 2021-09-27
Update the Freshmaker Docker Image to Fedora 34

ronnyhlim:
Updating the Docker image to use Fedora 34. Once merged in Dev, will run the Ansible tower job and see if there are any issues.

PR #131, opened 2021-11-10
Parse module name stream through modulemd_str

chuazhan:
Use libmodulemd to parse modue name and stream.
One thing to note is no package available in pypi.
Install it through python3-libmodulemd rpm package

JIRA: RHELWF-3779


chuazhan:
Haven't found a good solution as no libmodulemd package in pypi. Just add some doc in README for developer

chuazhan:
@hluk @qixiang @mvalik Please help to review, thanks!

freshmaker/kojiservice.py
```diff
@@ -407,6 +407,41 @@ def get_bundle_csv(self, build_nvr):
...
+        :param str build_nvr: NVR of module build.
+        :return: a dict for module name and its streams, e.g. {"key": [values]}
+        :rtype: dict
+        """
+        module_name_streams_mapping = dict()
+        build = self.get_build(build_nvr)
+        modulemd_str = build.get("extra", {}).get("typeinfo", {}).get("module", {}).get("modulemd_str")
+        if modulemd_str:
+            try:
+                import gi
```
qixiang, line 424: I think there is no strong reason of importing the modules in method instead of the beginning of this module, right?

freshmaker/kojiservice.py
```diff
@@ -407,6 +407,41 @@ def get_bundle_csv(self, build_nvr):
...
+                gi.require_version("Modulemd", "2.0")
+                from gi.repository import Modulemd
+                fedora_index = Modulemd.ModuleIndex.new()
```
qixiang, line 427: By checking the API doc, `Modulemd.ModuleIndex` [1] is mainly designed for dealing with repodata, we only need to deal with a modulemd string, so I think using `Modulemd.ModuleStream` [2] could be better.

[1] https://fedora-modularity.github.io/libmodulemd/latest/modulemd-2.0-Modulemd.ModuleIndexMerger.html
[2] https://fedora-modularity.github.io/libmodulemd/latest/ModulemdModuleStream.html

freshmaker/kojiservice.py
```diff
@@ -407,6 +407,41 @@ def get_bundle_csv(self, build_nvr):
...
+                    for module_name in module_names:
+                        module = merged_index.get_module(module_name)
+                        streams = module.get_stream_names()
+                        module_name_streams_mapping[module_name] = streams
+                else:
+                    log.error("Update ModuleIndex failed from %s: %s", modulemd_str, str(failures))
+            except Exception as e:
+                log.error("Parse module name and streams failed from NVR %s: %s", build_nvr, str(e))
+
+        return module_name_streams_mapping
```
qixiang, line 443: When error happens, an empty dict is returned, it makes no sense, I'd prefer return `None` if we hit any issue of fetching the module metadata.

qixiang: I believe a module build can only have one stream, not 100% sure, but according to the API names, it should be true, you can get only one stream from a modulemd, there is a `get_stream_name` API, as you see, it's `name` not `names`.

I'd suggest refactoring this API to just return a `Modulemd.ModuleStream` instance, like:
```
def get_modulemd(self, build_nvr):
    ...
    try:
        mmd = Modulemd.ModuleStream.read_string(modulemd_str, False, None, None)
    ...
    # any error happens
    # return None
    return mmd
```
then you can get module name and stream with:
```
mmd = session.get_modulemd("foobar-flatpak-xxxxx")
module_name = mmd.get_module_name()
module_stream = mmd.get_stream_name()
```

freshmaker/kojiservice.py
chuazhan: Ack. Probably this is a bad habit from previous experience. We were annoying by cyclic issue, so for an unusual package and it is only used in one place, normally it will be put in lines.

freshmaker/kojiservice.py
chuazhan: Ack. Fixing

freshmaker/kojiservice.py
chuazhan: Thanks @qixiang for you carefully review and constructive suggestions. I check the doc again, and there is description "modulemd_module_stream_read_string has been deprecated since version 2.11 and should not be used in newly-written code.
Use modulemd_read_packager_string() instead.". I change it with read_packager_string, please have a look.

.github/actions/tests/Dockerfile
```diff
@@ -7,7 +7,9 @@ LABEL \
 
 USER 0
 
-RUN dnf install -y python-devel fedora-packager gcc krb5-devel openldap-devel
+RUN dnf install -y python-devel fedora-packager gcc krb5-devel openldap-devel python3-libmodulemd \
+                   gobject-introspection-devel cairo-devel cairo-gobject-devel
```
hluk, line 11: Why are these needed? Why not only `python3-libmodulemd`?

test-requirements.txt
```diff
@@ -8,3 +8,4 @@ vcrpy
 requests_mock
 tox
 mypy
+pygobject
```
hluk, line 11: Why is this needed?

.github/actions/tests/Dockerfile
chuazhan: This is a good question. I was confused too when running the CI. I only install python3-libmodulemd before, but met error like "....Try sudo dnf install gobject-introspection-devel " when installing pygobject

test-requirements.txt
chuazhan: There is no pypi package available for libmodulemd . And actually we are only using gi package. After investigation,  It seems pygobject is good enough. So for testing, we include this package.

freshmaker/kojiservice.py
```diff
@@ -407,6 +409,25 @@ def get_bundle_csv(self, build_nvr):
...
+        """
+        Get Modulemd.ModuleStreamV2 instance through build NVR
+        which can be used to get module's name, stream, etc.
+
+        :param str build_nvr: NVR of module build.
+        :return: Modulemd.ModuleStreamV2 instance.
+        """
+        mmd = None
+        build = self.get_build(build_nvr)
+        modulemd_str = build.get("extra", {}).get("typeinfo", {}).get("module", {}).get("modulemd_str")
```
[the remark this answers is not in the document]
chuazhan: Ok. It is cleaner

.github/actions/tests/Dockerfile
chuazhan: So I am guessing it is a side-effect for installing pygobject

test-requirements.txt
hluk: I'm confused. Why is this needed only in tests? Is it because of `tox` fails to find the packages installed on the system?

Installing `python3-libmodulemd` should make Modulemd available for Python - I think it should install all necessary packages enable `from gi.repository import Modulemd`.

chuazhan: Yeah. I had the same idea with you before. I thought that would be ok to install  python3-libmodulemd only. Let me resubmit it again and double check it.

chuazhan:
Sorry guys for the notification spamming. I just want to save some test links for discussion.

test-requirements.txt
chuazhan: `I'm confused. Why is this needed only in tests? Is it because of tox fails to find the packages installed on the system?`
I am guessing the dependencies installed by python3-libmodulemd in system can not be used in tox environment.

Following is the failed cases with its codes:
https://github.com/redhat-exd-rebuilds/freshmaker/runs/4185263908?check_suite_focus=true
https://github.com/redhat-exd-rebuilds/freshmaker/commit/065dfdf2f1d4616d5eb054a17ecf6cead1767010

The error is: ModuleNotFoundError: No module named 'gi'

.github/actions/tests/Dockerfile
qixiang: Tests are running inside the tox virtualenvs (without using system site packages), so:

1. First you need `pygobject` to be installed in virtualenvs, that's why you need `pygobject` in `requirements.txt` (it's generated by pip-compile).
2. To allow pip install `pygobject` in virtualenvs, you need some `-devel` packages on system, because they're required to compile `pygobject` and its dependencies (`pycairo` and etc) while installing them within virtualenvs, thus you need to install `gobject-introspection-devel` and `cairo-gobject-devel` with yum/dnf before creating virtualenvs.
3. Fedora systems has `libmodulemd` installed by default, so once you have `pygobject` installed in virtualenvs, it's good enough to import modulemd modules within the virtualenvs. Thus you don't need to install `python3-libmodulemd` with dnf/yum.

chuazhan:
Rebase with the latest codes.

README.md
```diff
@@ -89,6 +89,12 @@ Install the dependencies with:
 python3 setup.py develop
 ```
 
+Install required rpm packages:
+
+```bash
+dnf install -y python-devel fedora-packager gcc krb5-devel openldap-devel gobject-introspection-devel cairo-gobject-devel
```
hluk, line 95: Would it make sense to drop the package list to `yum-packages-devel.txt` (similar to `yum-packages.txt`)?

```suggestion
sudo dnf install $(sed 's/#.*//' yum-packages-devel.txt)
```

```
❯ cat yum-packages-devel.txt
fedora-packager
gcc
krb5-devel
openldap-devel
python-devel

# pygobject dependencies
cairo-gobject-devel
gobject-introspection-devel
```

freshmaker/kojiservice.py
```diff
@@ -407,6 +412,29 @@ def get_bundle_csv(self, build_nvr):
...
+            modulemd_str = build["extra"]["typeinfo"]["module"]["modulemd_str"]
+        except KeyError:
+            return None
+
+        if modulemd_str:
+            try:
+                _, mmd = Modulemd.read_packager_string(modulemd_str, None, None)
+            except Exception as e:
+                log.error("Read packager string failed for NVR %s: %s", build_nvr, str(e))
+        return mmd
```
hluk, line 436:
```suggestion
        try:
            _, mmd = Modulemd.read_packager_string(modulemd_str, None, None)
            return mmd
        except Exception as e:
            log.error("Read packager string failed for NVR %s: %s", build_nvr, str(e))
            return None
```

chuazhan: I know some convention says return immediately, and some says one return point.  I don't have strong opinion to use one not the other.

README.md
chuazhan: Put it all in a file is ok for me. Just curious about why adding sudo and removing -y?

hluk: `sudo` is probably needed for most user but I'm not sure what's the convention for putting it into docs. It's OK to leave it out because it is clear when it is required.

I removed `-y` so that user/dev has chance to review the list of packages before they are installed. I would prefer to have the flag only in scripts.

requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with python 3.7
+# This file is autogenerated by pip-compile with python 3.8
```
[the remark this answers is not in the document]
chuazhan: yeah. I do refer the DockerFile's fedora version for generating this file, but I am also wondering if the requirements are really dependent on it. If the DockerFile updates the fedora version, we need to update this file as well? I am not sure.

README.md
chuazhan: > I removed -y so that user/dev has chance to review the list of packages before they are installed. I would prefer to have the flag only in scripts.
Ok. Sounds reasonable to me.

requirements.txt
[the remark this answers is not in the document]
chuazhan: I regenerate this file with fedora34's image and the only difference is the comment
"This file is autogenerated by pip-compile with python 3.9". I think we don't need to
wait for #119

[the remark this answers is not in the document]
chuazhan: Updated

README.md
chuazhan: Updated

freshmaker/kojiservice.py
chuazhan: Updated

freshmaker/kojiservice.py
```diff
@@ -407,6 +412,30 @@ def get_bundle_csv(self, build_nvr):
...
+        :return: Modulemd.ModuleStreamV2 instance.
+        """
+        mmd = None
+        build = self.get_build(build_nvr)
+        try:
+            modulemd_str = build["extra"]["typeinfo"]["module"]["modulemd_str"]
+        except KeyError:
+            return None
+
+        if modulemd_str:
```
hluk, line 431: This should not be needed now.

hluk (approved):

freshmaker/kojiservice.py
chuazhan: Ah. Yes, Empty string should be covered in following codes.

chuazhan:
@gnaponie Can you help to have a review, thanks!(Sorry for not involving you at the beginning of task. Just in the process of being familiar with the team and project)

tests/test_kojiservice.py
```diff
@@ -142,3 +142,19 @@ def test_get_bundle_csv_unavailable(mock_get, mock_koji, mock_log):
...
...\n      rpms:\n      - ghc\n    minimal:\n      description: just compiler and base\n      rpms:\n      - ghc-base-devel\n    small:\n      description: compiler with main core libs\n      rpms:\n      - ghc-devel\n  components:\n    rpms:\n      ghc:\n        rationale: compiler\n        repository: git+https://src.fedoraproject.org/rpms/ghc\n        cache: https://src.fedoraproject.org/repo/pkgs/ghc\n        ref: 9.2\n        buildorder: 1\n        arches: [aarch64, armv7hl, i686, ppc64le, s390x, x86_64]\n...\n', 'name': 'ghc', 'stream': '9.2', 'module_build_service_id': 13274, 'version': '3620211101111632', 'context': 'd099bf28', 'content_koji_tag': 'module-ghc-9.2-3620211101111632-d099bf28'}}}, 'id': 1850907, 'name': 'ghc', 'nvr': 'ghc-9.2-3620211101111632.d099bf28', 'owner_id': 56, 'owner_name': 'petersen', 'package_id': 1853, 'package_name': 'ghc', 'release': '3620211101111632.d099bf28', 'source': 'https://src.fedoraproject.org/modules/ghc.git?#5405c8a084c04476a956f1b3e49a733b888066a5', 'start_time': '2021-11-01 13:48:17+00:00', 'start_ts': 1635774497.0, 'state': 1, 'task_id': None, 'version': '9.2', 'volume_id': 0, 'volume_name': 'DEFAULT', 'cg_name': 'module-build-service'}
```
gnaponie, line 150: Could you put this on multiple lines for readability? Also you can remove fields that you don't need for the test, since this is a test.

freshmaker/kojiservice.py
```diff
@@ -407,6 +412,29 @@ def get_bundle_csv(self, build_nvr):
...
+        """
+        mmd = None
+        build = self.get_build(build_nvr)
+        try:
+            modulemd_str = build["extra"]["typeinfo"]["module"]["modulemd_str"]
+        except KeyError:
+            return None
+
+        try:
+            _, mmd = Modulemd.read_packager_string(modulemd_str, None, None)
```
gnaponie, line 432: I'm not familiar with Modulemd, could you explain it a bit to me? Thanks.

gnaponie:
> @gnaponie Can you help to have a review, thanks!(Sorry for not involving you at the beginning of task. Just in the process of being familiar with the team and project)

No worries! :)

gnaponie:
Seems like you have a conflict on the freshmaker/kojiservice.py file. Could you resolve it?

freshmaker/kojiservice.py
chuazhan: This is a library to manipulate module metadata files. More detail, you can check https://github.com/fedora-modularity/libmodulemd
I am using this api https://fedora-modularity.github.io/libmodulemd/latest/modulemd-2.0-Modulemd.html#modulemd-read-packager-string   to parse the module's name and stream from a yaml string.

tests/test_kojiservice.py
chuazhan: I was thinking to save the actual data and maybe in future we can reference different attributes other than name and stream.
But sure, let's make it more readable.

chuazhan:
> Seems like you have a conflict on the freshmaker/kojiservice.py file. Could you resolve it?

Sure. Rebased and upated

freshmaker/kojiservice.py
```diff
@@ -37,11 +37,16 @@
...
 import requests
 import yaml
-import freshmaker.utils
-from freshmaker import log, conf, db
-from freshmaker.consumer import work_queue_put
-from freshmaker.events import BrewContainerTaskStateChangeEvent
-from freshmaker.models import ArtifactBuild
+
+import gi
+gi.require_version('Modulemd', '2.0')  # noqa E402
```
hluk, line 42: Any `import` that goes under this, needs `# noqa E402`. :( Can we perhaps hide this under separate module (e.g. `freshmaker.modulemd`)?

tests/test_kojiservice.py
```diff
@@ -142,3 +142,37 @@ def test_get_bundle_csv_unavailable(mock_get, mock_koji, mock_log):
...
...tps://src.fedoraproject.org/modules/ghc.git?#5405c8a084c04476a956f1b3e49a733b888066a5\n      ursine_rpms:\n      - ghc-bytestring-devel-0:0.10.12.0-115.fc35.s390x\n      - ghc-exceptions-prof-0:0.10.4-115.fc35.armv7hl\n  dependencies:\n  - buildrequires:\n      ghc: [8.10]\n      platform: [f36]\n    requires:\n      platform: [f36]\n  references:\n    community: https://www.haskell.org/ghc\n    documentation: https://wiki.haskell.org/GHC\n    tracker: https://gitlab.haskell.org/ghc/ghc/issues\n  profiles:\n    all:\n      description: standard installation\n      rpms:\n      - ghc\n      - ghc-doc\n      - ghc-prof\n    default:\n      description: standard installation\n      rpms:\n      - ghc\n    minimal:\n      description: just compiler and base\n      rpms:\n      - ghc-base-devel\n    small:\n      description: compiler with main core libs\n      rpms:\n      - ghc-devel\n  components:\n    rpms:\n      ghc:\n        rationale: compiler\n        repository: git+https://src.fedoraproject.org/rpms/ghc\n        cache: https://src.fedoraproject.org/repo/pkgs/ghc\n        ref: 9.2\n        buildorder: 1\n        arches: [aarch64, armv7hl, i686, ppc64le, s390x, x86_64]\n...\n',
```
gnaponie, line 152: Is this needed to be that long for this test? I'm concerned (besides readability) that flake8 will fail.

freshmaker/kojiservice.py
gnaponie: Thank you!

gnaponie:
Anyhow, besides the very small comment (you're free to consider that as optional) the change looks good to me. +1

gnaponie (approved):

chuazhan:
> Anyhow, besides the very small comment (you're free to consider that as optional) the change looks good to me. +1 I deeply apologize for the delay in the review, I had some days of PTO. Thank you for the patience!

Thanks for the review. No problem at all.

tests/test_kojiservice.py
chuazhan: Ack. Make it shorter

freshmaker/kojiservice.py
chuazhan: Yes. But it seems sth come be better, change it a bit. 
I think this is a package we will use, not wrapping a new service, so I prefer not to put it in a separate module.

I tried to put import gi, gi.require_version('Modulemd', '2.0')  # noqa E402 at the end of import chain and there is no flake8 issue, but I guess there is some convention to put the freshmaker model at the end.

chuazhan:
Rebased and updated

freshmaker/kojiservice.py
hluk: The idea was to create simple file `freshmaker/modulemd.py`:

```python
import gi
gi.require_version('Modulemd', '2.0')
from gi.repository import Modulemd  # noqa E402
```

And use `from freshmaker.modulemd import Modulemd`.

chuazhan: Yeah. I agree that this will resolve the flake issues. But do we have strong opinion to create a new file for fixing flake8 issues, is this a convention for fixing this kind of issues, I am not sure.

PR #133, opened 2021-11-15
JSON logging into file

ElenaKarolinaSemanova:
Make JSON logger to write structured data to a file.
(JIRA: CWFHEALTH-778)

### DIFF
--- a/fedmsg.d/freshmaker-logging.py
+++ b/fedmsg.d/freshmaker-logging.py
@@ -11,7 +11,7 @@ config = dict(
         formatters={
             'bare': {
                 'datefmt': '%Y-%m-%d %H:%M:%S',
-                'format': '[%(asctime)s][%(name)10s %(levelname)7s] %(message)s',
+                'format': '[%(asctime)s][%(name)10s %(levelname)7s] %(message)s'
             },
         },
         loggers=dict(
--- a/yum-packages.txt
+++ b/yum-packages.txt
@@ -39,4 +39,5 @@ python3-systemd
 python3-semver
 python3-tabulate
 python3-vcrpy
+python3-json-logger
 systemd


ElenaKarolinaSemanova:
@apaplaus @mulaievaRH Please have a look

test-requirements.txt
```diff
@@ -8,3 +8,4 @@ vcrpy
 requests_mock
 tox
 mypy
+python-json-logger
```
apaplaus, line 11: Maybe move it to common `requirements.txt`?

apaplaus (commented):
As for modifying file `fedmsg.d/freshmaker-logging.py`. It's not very efficient, as if you check Dockerfile we are erasing that file at some point. And actual configuration for logging is done in **fresmaker-playbooks** repository in `roles/freshmaker-openshift/templates/fedmsg_logging.py` file. So you need to make changes there. And maybe @qixiang can tell why we are keeping this file in the repo at all, since we don't use it for the deployment. Maybe we should at least leave a comment to not use this file for actual configuration somewhere, because as you see it could be a little bit confusing.
And as for actual change, not sure what `()` key is doing in fedmsg logger config. Maybe you could attach some examples of logs produced by this config?
And the last thing, did you try to use it with Splunk? And if so, what do logs look nice there?

ElenaKarolinaSemanova:
@apaplaus @mulaievaRH @qixiang Please review

qixiang:
Andrei's comment is correct, our deployments are using the image with `fedmsg.d/freshmaker-logging.py` removed, so you need to update our ansible-playbook to apply the new logger formatter.

`fedmsg.d/freshmaker-logging.py` is still in code repo for local development purpose. The fedmsg configuration values used at runtime are determined by checking in the following order:
1. Built-in defaults
2. Global config file (/etc/fedmsg.d/*.py)
3. Local Config file (./fedmsg.d/*.py)
4. Command line arguments

While running freshmaker/fedmsg scripts locally, we don't want to change the global config file because the global config file can be used by other components on host.

And actually this change won't enable json format logging for freshmaker logs, since we don't have freshmaker logs in `./fedmsg.d/freshmaker-logging.py` (it's commented out).

So to enable the json format logging for freshmaker, you need to:
1. Update that in our ansible-playbooks.
2. Build docker image with `python-json-logger` installed.

fedmsg.d/freshmaker-logging.py
```diff
@@ -3,15 +3,16 @@
         handlers={
             'console': {
                 'class': 'logging.StreamHandler',
-                'formatter': 'bare',
+                'formatter': 'json',
```
ElenaKarolinaSemanova, line 6: @qixiang Should I also add the new json formatter instead of changing the bare one here?

qixiang: I think we don't need to add json formatter for development purpose, so the only change you need to do in this repo is  adding the `python3-json-logger` package to `yum-packages.txt`, so the image (used for our deployments) can be built with the new required rpm installed.

apaplaus (approved):

gnaponie:
@ElenaKarolinaSemanova the change looks good. I've removed the WIP, and I'll merge the PR now.

PR #134, opened 2021-11-16
Refactor the bundle rebuild handler

qixiang:
The main changes are:

1. While rebuilding unreleased bundles, freshmaker tries to get
   bundle data from Pyxis, but actually Pyxis doesn't have bundle
   data for unreleased bundles, this is fixed to use brew to get
   bundle data.
2. Instead of getting all latest bundles first and check one by one,
   use related image digest to get impacted bundles.
3. The manual case and auto case were handled differently, now they
   are handled with same workflow.
4. Manual rebuild request is refactored to only use advisory to find
   and construct the updated CSV data. "bundle_images" in request is
   not used anymore.

JIRA: CWFHEALTH-914

### DIFF
--- a/freshmaker/events.py
+++ b/freshmaker/events.py
@@ -468,49 +468,23 @@ class BotasErrataShippedEvent(ErrataBaseEvent):
         super().__init__(msg_id, advisory, dry_run=dry_run)
 
 
-class ManualBundleRebuild(BaseEvent):
+class ManualBundleRebuildEvent(ErrataBaseEvent):
     """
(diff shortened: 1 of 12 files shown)


qixiang:
@apaplaus @gnaponie @mulaievaRH could you help to review this? I know it's not easy to review a single big change, however it's hard to break down this change to small commits.

gnaponie:
Quite a lot of changes! I've checked them from here and they seem good to me.

apaplaus (approved):

PR #136, opened 2021-11-19
Update bob msg handler

ronnyhlim:
JIRA: https://issues.redhat.com/browse/CWFHEALTH-859

**Update Bob message handler to avoid usage of Docker-Pulp**
- updated `get_docker_repo_tags` to parse external repo names from ET response
        old response returned:  **{'foo-container-1-1': {'foo-526': ['5.26', 'latest']}}**
        new response returns: **{'foo-526': ['5.26', 'latest']}**
- updated `rebuild_images_depending_on_advisory` to use updated `get_docker_repo_tags` and removed usage of Docker-pulp
- Removed docker pulp credentials from `config.py` and removed all instances of docker pulp usage from unit tests.

### DIFF
--- a/freshmaker/errata.py
+++ b/freshmaker/errata.py
@@ -253,15 +253,12 @@ class Errata(object):
         get_advisory_cdn_docker_file_list
         :param int errata_id: Errata advisory ID.
         :rtype: dict
-        :return: Dict of advisory builds with repo and tag config:
+        :return: Dict of repos and tags config:
             {
-                'build_NVR': {
-                    'cdn_repo1': [
-                        'tag1',
-                        'tag2'
-                    ],
-                    ...
-                },
+                'cdn_repo1': [
+                    'tag1',
+                    'tag2'
+                ],
                 ...
             }
         """
@@ -276,15 +273,11 @@ class Errata(object):
                         "returned None.")
             return None
 
-        repo_tags = dict()
-        for build_nvr in response:
-            if build_nvr not in repo_tags:
-                repo_tags[build_nvr] = dict()
-            repos = response[build_nvr]['docker']['target']['repos']
-            for repo in repos:
-                tags = repos[repo]['tags']
-                repo_tags[build_nvr][repo] = tags
-        return repo_tags
+        return {
+            repo: repo_data['tags']
+            for build_data in response.values()
+            for repo, repo_data in build_data['docker']['target']['external_repos'].items()
+        }
 
     def advisories_from_event(self, event):
         """
--- a/freshmaker/config.py
+++ b/freshmaker/config.py
@@ -304,18 +304,6 @@ class Config(object):
             'type': str,
             'default': '',
             'desc': 'Password to login Pulp.'},
-        'pulp_docker_server_url': {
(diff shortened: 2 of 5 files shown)


freshmaker/errata.py
```diff
@@ -281,7 +278,11 @@ def get_docker_repo_tags(self, errata_id):
             for repo in repos:
                 tags = repos[repo]['tags']
                 repo_tags[build_nvr][repo] = tags
-        return repo_tags
+
+        tags = dict()
+        for repos in repo_tags.values():
+            tags.update(repos)
+        return tags
```
hluk, line 285: Looks like all this and the old code can be rewritten as:

```python
        return {
            repo: repo_data['tags']
            for build_data in response.values()
            for repo, repo_data in build_data['docker']['target']['repos'].items()
        }
```

ronnyhlim: agreed, this is a much cleaner way to do this. changes made

freshmaker/errata.py
```diff
@@ -273,15 +270,11 @@ def get_docker_repo_tags(self, errata_id):
...
-                repo_tags[build_nvr] = dict()
-            repos = response[build_nvr]['docker']['target']['repos']
-            for repo in repos:
-                tags = repos[repo]['tags']
-                repo_tags[build_nvr][repo] = tags
-        return repo_tags
+        return {
+            repo: repo_data['tags']
+            for build_data in response.values()
+            for repo, repo_data in build_data['docker']['target']['repos'].items()
```
hluk, line 276: Using similar change in metaxor as reference: Shouldn't this use key `'external_repos'` instead of `'repos'`?

ronnyhlim: hmm so the original method uses `repos` (https://github.com/redhat-exd-rebuilds/freshmaker/blob/master/freshmaker/errata.py#L283) and when I run unit tests using 
`for repo, repo_data in build_data['docker']['target']['external_repos'].items()` there is a key_error on `external_repos`.

hluk: Do unit tests have correct data?

ronnyhlim: yup, I believe so because when _response_ is created from passing the _errata_id_ into `xmlrpc.get_advisory_cdn_docker_file_list`, we get a dict that has "repos" as one of the keys. The unit tests reflect this so I believe it's correct. However, in case I am mistaken do you think we should change the 
`xmlrpc.get_advisory_cdn_docker_file_list` response to be 'external_repos' instead of 'repos'? (and make the same change in this method?
Screenshot from 2021-11-24 00-21-15
m

hluk: The "external_repos" is not mentioned in ET XMLRPC docs but I found it in ET source code. Looks like the "external_repos" value is compatible with quay.io - not sure if we need it here.

mbirger-rh (changes requested):

freshmaker/errata.py
ronnyhlim: ahh changes made :+1:

mbirger-rh (approved):

PR #137, opened 2021-11-22
Get signed state for builds from updated ET API

hluk:
New "with_sig_key" GET parameter was added to Errata Tool API in
RHELWF-4641. With this we can list all attached builds with "is_signed"
property:

    /api/v1/erratum/{id}/builds_list?with_sig_key=1

### DIFF
--- a/freshmaker/errata.py
+++ b/freshmaker/errata.py
@@ -227,6 +227,9 @@ class Errata(object):
     def _get_attached_builds(self, errata_id):
         return self._errata_http_get(f"advisory/{errata_id}/builds.json")
 
+    def _get_builds_by_product(self, errata_id):
+        return self._errata_rest_get(f"/erratum/{errata_id}/builds_list?with_sig_key=1")
+
     @region.cache_on_arguments()
     def _advisories_from_nvr(self, nvr):
         """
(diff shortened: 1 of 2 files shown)


freshmaker/errata.py
```diff
@@ -311,22 +314,14 @@ def builds_signed(self, errata_id):
...
-        # For each NVR, check that all the rpms are signed.
-        for nvr in nvrs:
-            log.info("Checking whether the build %s is signed", str(nvr))
-            build = self._errata_rest_get("build/%s" % str(nvr))
-            if "rpms_signed" not in build or not build["rpms_signed"]:
-                return False
-
-        return True
+        data_by_product = self._get_builds_by_product(errata_id)
+        log.warning(data_by_product)
```
qixiang, line 318: This log message is not so readable and includes content (product data, rpms in builds for different arches) we don't care about when there are many builds attached, I'd suggest logging a warning message with build NVR and return `False` only when a build is not signed.

hluk: Ooops, I used it to debug some problems. Removing.

qixiang (approved):
:+1:

PR #138, opened 2021-11-29
Add methods to check if an image can be rebuilt

chuazhan:
Change get_images_by_nvr to make it support passing
include_fields param.
Add an api to get the image's metadata and use these
metadata with auto_rebuild_tags api to check if this
image can be rebuilt or not.

JIRA: RHELWF-3779

### DIFF
--- a/freshmaker/pyxis.py
+++ b/freshmaker/pyxis.py
@@ -273,15 +273,18 @@ class Pyxis(object):
                           'filter': q_filter}
         return self._pagination('images', request_params)
 
-    def get_images_by_nvr(self, nvr):
+    def get_images_by_nvr(self, nvr, include=None):
         """
         Get images by image's NVR
 
         :param str nvr: NVR of image
+        :param list include: included fields in image data.
         :return: images
         :rtype: list
         """
         request_params = {"include": "data.architecture,data.brew,data.repositories"}
+        if include:
+            request_params = {'include': ','.join(include)}
         return self._pagination(f'images/nvr/{nvr}', request_params)
 
     def get_auto_rebuild_tags(self, registry, repository):
@@ -313,4 +316,20 @@ class Pyxis(object):
         for label in images[0].get("parsed_data", {}).get("labels", []):
             if label["name"] == "com.redhat.delivery.operator.bundle" and label["value"] == "true":
                 return True
+        return False
+
+    def image_is_tagged_auto_rebuild(self, nvr):
+        include = ["data.repositories.registry", "data.repositories.repository",
+                   "data.repositories.tags.name", "data.repositories.published"]
+        images = self.get_images_by_nvr(nvr, include=include)
+        if images:
+            # Only use item 0 for getting necessary metadata as the difference between
+            # different items are the arches info, the metadata we want are the same.
+            image = images[0]
+            for repo in image['repositories']:
+                if not repo['published']:
+                    continue
+                auto_rebuild_tags = self.get_auto_rebuild_tags(repo['registry'], repo['repository'])
+                if set(tag['name'] for tag in repo['tags']) & set(auto_rebuild_tags):
+                    return True
         return False
--- a/tests/test_pyxis.py
+++ b/tests/test_pyxis.py
@@ -598,3 +598,37 @@ class TestQueryPyxis(helpers.FreshmakerTestCase):
         }]
         is_bundle = self.px.is_bundle("foobar-bundle-1-234")
         self.assertFalse(is_bundle)
+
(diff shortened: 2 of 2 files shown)


freshmaker/pyxis.py
```diff
@@ -273,15 +273,16 @@ def get_images_by_digest(self, digest):
                           'filter': q_filter}
         return self._pagination('images', request_params)
 
-    def get_images_by_nvr(self, nvr):
+    def get_images_by_nvr(self, nvr, include_fields=[]):
```
qixiang, line 276: Mutable default arguments is a command python pitfall: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

And I'd suggest just use `include` instead of `include_fields` since Pyxis is using `include`.

freshmaker/pyxis.py
```diff
@@ -273,15 +273,16 @@ def get_images_by_digest(self, digest):
...
         """
         Get images by image's NVR
 
         :param str nvr: NVR of image
+        :param list include_fields: included fields in image data.
         :return: images
         :rtype: list
         """
-        request_params = {"include": "data.architecture,data.brew,data.repositories"}
+        request_params = {'include': ','.join(include_fields)} if include_fields else {}
```
qixiang, line 285: The default `include` has been changed, it breaks the existing code, you can do it in this way:
```
request_params = {'include': 'data.architecture,data.brew,data.repositories'}
if include_fields:
    request_params.update({'include': ','.join(include_fields})
```

freshmaker/pyxis.py
```diff
@@ -314,3 +315,50 @@ def is_bundle(self, nvr):
             if label["name"] == "com.redhat.delivery.operator.bundle" and label["value"] == "true":
                 return True
         return False
+
+    def get_image_metadata_by_nvr(self, nvr):
```
qixiang, line 319: `get_image_metadata_by_nvr` and `get_rebuild_image_id` are ambiguous names or not describing the purpose very well, I'd suggest something like this (not tested):
```
def image_is_tagged_auto_rebuild(self, nvr):
    images = get_images_by_nvr(nvr, include=...)
    for image in images:
        for repo in image['repositories']:
            if not repo['published']:
                continue
            auto_rebuild_tags = self.get_auto_rebuild_tags(repo['registry'], repo['repository'])
            if set(image['tags']) & set(auto_rebuild_tags):
                return True
    return False
```

qixiang (changes requested):

tests/test_pyxis.py
```diff
@@ -598,3 +602,70 @@ def test_is_bundle_false(self, page):
...
+
+    @patch('freshmaker.pyxis.Pyxis.get_auto_rebuild_tags')
+    @patch('freshmaker.pyxis.Pyxis._pagination')
+    def test_non_rebuild_image(self, page, get_auto_rebuild_tags):
+        page.return_value = [
+            {'_links': {},
+             'repositories': [{'_links': {},
+                               '_id': '087721b',
+                               'published': True,
+                               'registry': 'registry.access.redhat.com',
```
gnaponie, line 663: Could you change this to some fake string? This is public/open source, so maybe let's not mention red hat.

freshmaker/pyxis.py
chuazhan: Yeah. A good point for the pitfall. I was considering to use a string, but I didn't choose it, because I need to tell the user to input a string with comma.

freshmaker/pyxis.py
chuazhan: Yeah. I am thinking to make the method more generic, so refactor it a bit. With a default request_params sounds ok for me.

tests/test_pyxis.py
chuazhan: Ack. Fixing

freshmaker/pyxis.py
chuazhan: Ok. Create a new method get_image_metadata_by_nvr is considering to use it when getting rpm manifest data in future. But yeah, let's consider it in following patch, maybe we don't need it. Ack. Fixing

freshmaker/pyxis.py
chuazhan: Is it only for empty list default value, maybe include_fields=None?

hluk (approved):

gnaponie:
Much more cleaner! Looks good +1

gnaponie (approved):

chuazhan:
Rename rebuild_image_id to a better one is_tagged_auto_rebuild

qixiang (approved):

PR #139, opened 2021-12-01
CSV file can endswith ".csv.yaml"

qixiang:
Besides of ".clusterserviceversion.yaml", CSV file can also endswith
".csv.yaml", so freshmaker should check both the suffixes.


gnaponie (approved):
Looks good!

PR #144, opened 2021-12-02
Filter out rebuild image candidates in handle method

chuazhan:
JIRA: RHELWF-3779

### DIFF
--- a/freshmaker/config.py
+++ b/freshmaker/config.py
@@ -427,6 +427,11 @@ class Config(object):
             'desc': 'A list of bundle name-version entries that do not need to have an auto-rebuild '
                     'tag to be rebuilt. This only applies to the HandleBotasAdvisory handler'
         },
+        'flatpak_server_url ': {
+            'type': str,
+            'default': '',
+            'desc': 'Root url of Flatpak index service'
+        },
     }
 
     def __init__(self, conf_section_obj):
--- a/freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
+++ b/freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
@@ -20,13 +20,150 @@
 # SOFTWARE.
 #
 # Written by Valerij Maljulin <vmaljuli@redhat.com>
+# Written by Chuang Zhang <chuazhan@redhat.com>
 
-from freshmaker.handlers import ContainerBuildHandler
+import requests
+from http import HTTPStatus
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
+
+from freshmaker import conf, db, log
+from freshmaker.errata import Errata
 from freshmaker.events import FlatpakModuleAdvisoryReadyEvent
+from freshmaker.handlers import ContainerBuildHandler, fail_event_on_handler_exception
+from freshmaker.kojiservice import koji_service
+from freshmaker.lightblue import LightBlue
+from freshmaker.models import Event
+from freshmaker.pyxis import Pyxis
+from freshmaker.types import EventState
 
 
 class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
-    name = 'RebuildFlatpakApplicationOnModuleReady'
+    # Module ready means Flatpak module advisory is in QE status
+    # and all attach build is in signed status
+    name = "RebuildFlatpakApplicationOnModuleReady"
 
     def can_handle(self, event):
         return isinstance(event, FlatpakModuleAdvisoryReadyEvent)
+
+    @fail_event_on_handler_exception
+    def handle(self, event):
+
+        if event.dry_run:
+            self.force_dry_run()
+
+        self.event = event
+
+        db_event = Event.get_or_create_from_event(db.session, event)
+        self.set_context(db_event)
+
+        auto_build_images_list = self._get_auto_rebuild_image_list(event)
+        rebuild_images = self._filter_images_with_higher_rpm_nvr(
+            event, auto_build_images_list
+        )
+        if not auto_build_images_list or not rebuild_images:
+            msg = "There is no image can be rebuilt. " f"message_id: {event.msg_id}"
+            db_event.transition(EventState.SKIPPED, msg)
+            db.session.commit()
+            self.log_info(msg)
+            return []
+
+        images_nvr = ",".join([rebuild_image.nvr for rebuild_image in rebuild_images])
+        self.log_info("Following images %s will be rebuilt" % images_nvr)
+
+        # TODO: Return empty list so far as this method is still in progress.
+        # Will think about whether to return the real BaseEvent objects in future.
+        return []
+
+    def _get_requests_session(cls, retry_options={}):
+        """
+        Create a requests session. Reference from:
+        https://github.com/release-engineering/cachito/blob/master/cachito/workers/requests.py
+
+        :param dict retry_options: overwrite options for initialization of Retry instance
+        :return: the configured requests session
+        :rtype: requests.Session
+        """
+        DEFAULT_RETRY_OPTIONS = {
+            "total": 5,
+            "read": 5,
+            "connect": 5,
+            "backoff_factor": 1.3,
+            "status_forcelist": (500, 502, 503, 504),
+        }
+        session = requests.Session()
+        retry_options = {**DEFAULT_RETRY_OPTIONS, **retry_options}
+        adapter = HTTPAdapter(max_retries=Retry(**retry_options))
+        session.mount("http://", adapter)
+        session.mount("https://", adapter)
+        return session
+
+    def _get_auto_rebuild_image_list(self, event):
+        """
+        Get rebuild image list which can be auto rebuilt.
+
+        :param event FlatpakModuleAdvisoryReadyEvent: The event this handler
+            is currently handling.
+        :rtype: list
+        :return: List of images which can be auto rebuilt.
+        """
+        if not conf.pyxis_server_url:
+            raise ValueError("'PYXIS_SERVER_URL' parameter should be set")
+        self._pyxis = Pyxis(conf.pyxis_server_url)
+
+        if not conf.flatpak_server_url:
+            raise ValueError(
+                "'FLATPAK_SERVER_URL' parameter should be set"
+            )
+
+        errata = Errata()
+        module_nvrs = errata.get_cve_affected_build_nvrs(event.advisory.errata_id, True)
+        rebuild_images_list = list()
+
+        with koji_service(
+            conf.koji_profile, log, login=False, dry_run=self.dry_run
+        ) as session:
+            for module_nvr in module_nvrs:
+                mmd = session.get_modulemd(module_nvr)
+                content_index_url = "{}/released/contents/modules/{}:{}.json".format(
+                    conf.flatpak_server_url,
+                    mmd.get_module_name(),
+                    mmd.get_stream_name(),
+                )
+                req_session = self._get_requests_session()
+                response = req_session.get(content_index_url)
+                status_code = response.status_code
+                if status_code == HTTPStatus.OK:
+                    images_info = response.json().get("Images", [])
+                    for image_info in images_info:
+                        image_nvr = image_info["ImageNvr"]
+                        is_tagged_auto_build = self._pyxis.image_is_tagged_auto_rebuild(
+                            image_nvr
+                        )
+                        if is_tagged_auto_build:
+                            rebuild_images_list.append(image_nvr)
+                else:
+                    self.log_error("Fetching module %s data failed.", module_nvr)
+
+        return rebuild_images_list
+
+    def _filter_images_with_higher_rpm_nvr(self, event, rebuild_images_list):
+        """
+        Filter out images which have higher nvr in rebuild_images_list than
+        the advisory's.
+
+        :param event FlatpakModuleAdvisoryReadyEvent: The event this handler
+            is currently handling.
+        :rtype: list
+        :return: List of ContainerImage instances which can be auto rebuilt.
+        """
+        errata = Errata()
+        errata_rpm_nvrs = errata.get_cve_affected_rpm_nvrs(event.advisory.errata_id)
+        lb = LightBlue(
+            server_url=conf.lightblue_server_url,
+            cert=conf.lightblue_certificate,
+            private_key=conf.lightblue_private_key,
+            event_id=self.current_db_event_id,
+        )
+        images = lb.get_images_by_nvrs(rebuild_images_list, rpm_nvrs=errata_rpm_nvrs)
+        return images
--- a/tests/handlers/koji/test_rebuild_flatpak_application_on_module_ready.py
+++ b/tests/handlers/koji/test_rebuild_flatpak_application_on_module_ready.py
@@ -2,19 +2,23 @@
 
 from unittest.mock import patch
 
+from freshmaker import db
 from freshmaker.errata import ErrataAdvisory
 from freshmaker.events import FlatpakModuleAdvisoryReadyEvent
 from freshmaker.handlers.koji import RebuildFlatpakApplicationOnModuleReady
+from freshmaker.models import Event
+from freshmaker.types import EventState
 from tests import helpers
 
 
-@patch("freshmaker.events.conf.parsers",
-       [
-           "freshmaker.parsers.errata:ErrataAdvisorySigningChangedParser",
-           "freshmaker.parsers.errata:ErrataAdvisoryStateChangedParser",
-       ])
+@patch(
+    "freshmaker.events.conf.parsers",
+    [
+        "freshmaker.parsers.errata:ErrataAdvisorySigningChangedParser",
+        "freshmaker.parsers.errata:ErrataAdvisoryStateChangedParser",
+    ],
+)
 class TestFlatpakModuleAdvisoryReadyEvent(helpers.ModelsTestCase):
-
     def _patch(self, to_patch):
         patcher = patch(to_patch, autospec=True)
         self.addCleanup(patcher.stop)
@@ -25,16 +29,27 @@ class TestFlatpakModuleAdvisoryReadyEvent(helpers.ModelsTestCase):
 
         self.consumer = self.create_consumer()
 
-        self.get_pulp_repository_ids = self._patch("freshmaker.errata.Errata.get_pulp_repository_ids")
+        self.get_pulp_repository_ids = self._patch(
+            "freshmaker.errata.Errata.get_pulp_repository_ids"
+        )
         self.get_pulp_repository_ids.return_value = ["rhel-8-for-x86_64-hidden-rpms"]
 
         self.builds_signed = self._patch("freshmaker.errata.Errata.builds_signed")
         self.builds_signed.return_value = True
 
-        self.from_advisory_id = self._patch("freshmaker.errata.ErrataAdvisory.from_advisory_id")
+        self.from_advisory_id = self._patch(
+            "freshmaker.errata.ErrataAdvisory.from_advisory_id"
+        )
         self.advisory = ErrataAdvisory(123, "RHSA-123", "QE", ["module"], "Critical")
         self.from_advisory_id.return_value = self.advisory
         self.handler = RebuildFlatpakApplicationOnModuleReady()
+        self.mock_get_auto_rebuild_image_list = self._patch(
+            "freshmaker.handlers.koji.RebuildFlatpakApplicationOnModuleReady._get_auto_rebuild_image_list"
+        )
+        self.mock_filter_images_with_higher_rpm_nvr = self._patch(
+            "freshmaker.handlers.koji.RebuildFlatpakApplicationOnModuleReady._filter_images_with_higher_rpm_nvr"
+        )
+        self.event = FlatpakModuleAdvisoryReadyEvent("123", self.advisory)
 
     def tearDown(self):
         self.consumer = self.create_consumer()
@@ -49,7 +64,7 @@ class TestFlatpakModuleAdvisoryReadyEvent(helpers.ModelsTestCase):
                 "content_types": ["module"],
                 "errata_status": "QE",
                 "errata_id": 123,
-            }
+            },
         }
         event = self.consumer.get_abstracted_msg(msg)
         self.assertIsInstance(event, FlatpakModuleAdvisoryReadyEvent)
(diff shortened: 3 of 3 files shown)


freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,108 @@
...
 class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
     name = 'RebuildFlatpakApplicationOnModuleReady'
 
     def can_handle(self, event):
         return isinstance(event, FlatpakModuleAdvisoryReadyEvent)
+
+    @fail_event_on_handler_exception
+    def handle(self, event):
+        """
+        TODO: Update in future.
```
hluk, line 48: Leftover TODO.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,108 @@
...
+
+        :param event FlatpakModuleAdvisoryReadyEvent: The event this handler
+            is currently handling.
+        :rtype: list
+        :return: List of images which can be auto rebuilt.
+        """
+        errata = Errata()
+        module_nvrs = errata.get_cve_affected_build_nvrs(event.advisory.errata_id, True)
+        rebuild_images_list = list()
+        flatpak_server_url = "https://flatpaks.engineering.redhat.com"
```
hluk, line 89: This should be in the configuration. The default could be either a dummy value or empty (which would effectively disable handling flatpaks).

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,108 @@
...
+
+        with koji_service(conf.koji_profile, log, login=False, dry_run=self.dry_run) as session:
+            for module_nvr in module_nvrs:
+                mmd = session.get_modulemd(module_nvr)
+                content_index_url = '{}/released/contents/modules/{}:{}.json'.format(flatpak_server_url,
+                                                                                     mmd.get_module_name(),
+                                                                                     mmd.get_stream_name())
+                response = requests.get(content_index_url)
+                status_code = response.status_code
+                if status_code == HTTPStatus.OK:
```
hluk, line 100: How about log on error?

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,108 @@
...
+        flatpak_server_url = "https://flatpaks.engineering.redhat.com"
+        self._pyxis = Pyxis(conf.pyxis_server_url)
+
+        with koji_service(conf.koji_profile, log, login=False, dry_run=self.dry_run) as session:
+            for module_nvr in module_nvrs:
+                mmd = session.get_modulemd(module_nvr)
+                content_index_url = '{}/released/contents/modules/{}:{}.json'.format(flatpak_server_url,
+                                                                                     mmd.get_module_name(),
+                                                                                     mmd.get_stream_name())
+                response = requests.get(content_index_url)
```
hluk, line 98: Better to use `requests.Session` with some retry configured.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,108 @@
...
+
+        rebuild_images = self._filter_images_with_higher_rpm_nvr(event, auto_build_images_list)
+        if not rebuild_images:
+            msg = ("There is no image with higher rpm nvr can be rebuilt. "
+                   f"message_id: {event.msg_id}")
+            db_event.transition(EventState.SKIPPED, msg)
+            db.session.commit()
+            self.log_info(msg)
+            return []
+
```
hluk, line 76: Missing return statement? Docstring for `BaseHandler.handle()` says it returns list.

hluk (changes requested):

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Do you mean to remove the docstring totally? And we can add it when everything is ready?

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ack. Fixing!

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ack. Fixing

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Yeah. Actually I am not sure what to return in current case as the implementation is not complete

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: I am thinking to leave a placeholder there and update it in future.

hluk: Not sure if there is anything worth noting - if not, I'm OK with removing it.

BTW, since this is on code review and not marked as draft, I assume it's ready. :)

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ack. Fixing

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: return [] so far and see if needed to be updated in future

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Yeah. It is ready. The thing is this handle method will be added more implementations for future tickets. So I am assuming to place a holder here to remind me to change it in future. Maybe it's worthless, let's remove it for no confusion :D

freshmaker/config.py
```diff
@@ -427,6 +427,11 @@ class Config(object):
             'desc': 'A list of bundle name-version entries that do not need to have an auto-rebuild '
                     'tag to be rebuilt. This only applies to the HandleBotasAdvisory handler'
         },
+        'flatpak_server_url ': {
+            'type': str,
+            'default': '',
```
hluk, line 432: How do we want to handle it if the string is empty? Is it OK to just ignore the case?

tests/handlers/koji/test_rebuild_flatpak_application_on_module_ready.py
```diff
@@ -35,6 +38,9 @@ def setUp(self):
         self.advisory = ErrataAdvisory(123, "RHSA-123", "QE", ["module"], "Critical")
         self.from_advisory_id.return_value = self.advisory
         self.handler = RebuildFlatpakApplicationOnModuleReady()
+        self.mock_get_auto_rebuild_image_list = self._patch('freshmaker.handlers.koji.RebuildFlatpakApplicationOnModuleReady._get_auto_rebuild_image_list')
+        self.mock_filter_images_with_higher_rpm_nvr = self._patch('freshmaker.handlers.koji.RebuildFlatpakApplicationOnModuleReady._filter_images_with_higher_rpm_nvr')
```
gnaponie, line 42: Could you put those 2 lines on multiple lines? For readability.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,112 @@
...
+        return []
```
hluk, line 75: A `TODO` comment would be probably useful here to explain why the empty list is returned. And perhaps at least log `rebuild_images` for now?

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,112 @@
...
+        if not auto_build_images_list:
+            msg = ("There is no auto rebuild image can be rebuilt. "
+                   f"message_id: {event.msg_id}")
+            db_event.transition(EventState.SKIPPED, msg)
+            db.session.commit()
+            self.log_info(msg)
+            return []
+
+        rebuild_images = self._filter_images_with_higher_rpm_nvr(event, auto_build_images_list)
+        if not rebuild_images:
```
gnaponie, line 67: Optional: there's some repeated code in this block and the previous one (about transitioning the event to SKIPPED), you might consider re-arranging the code so that you don't repeat it.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,112 @@
...
+        self._pyxis = Pyxis(conf.pyxis_server_url)
+
+        with koji_service(conf.koji_profile, log, login=False, dry_run=self.dry_run) as session:
+            for module_nvr in module_nvrs:
+                mmd = session.get_modulemd(module_nvr)
+                content_index_url = '{}/released/contents/modules/{}:{}.json'.format(conf.flatpak_server_url,
+                                                                                     mmd.get_module_name(),
+                                                                                     mmd.get_stream_name())
+                session = requests.Session()
+                adapter = HTTPAdapter(max_retries=3)
```
hluk, line 98: Oh, this does not handle the actual HTTP request timeouts - from [docs](https://2.python-requests.org/en/master/api/#requests.adapters.HTTPAdapter):

> max_retries – The maximum number of retries each connection should attempt. Note, this applies only to failed DNS lookups, socket connections and connection timeouts, never to requests where data has made it to the server. By default, Requests does not retry failed connections. If you need granular control over the conditions under which we retry a request, import urllib3’s Retry class and pass that instead.

See a good example of a well-tested retry implementation from cachito: https://github.com/release-engineering/cachito/blob/db1acca3f5c56add57d7f82b55c44122e93ea394/cachito/workers/requests.py#L48

hluk (changes requested):

freshmaker/config.py
chuazhan: Not quite sure about your point. At least there is a default empty string instead of no attribute.

tests/handlers/koji/test_rebuild_flatpak_application_on_module_ready.py
chuazhan: Ok. Let me black it. Maybe there will be some unrelated changes.

chuazhan: Ok. Let me black it. Maybe there will be some unrelated changes.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ack. Fixing

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ack. Fixing.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ack. Fixing

freshmaker/config.py
hluk: I mean, empty string would lead to bad URL to flatpak API - I don't see that handled.

We could skip flatpak handling if this is empty - this could be useful also for some troubleshooting in near future.

chuazhan: Ok. Got it. Fixing

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,148 @@
...
+        :param event FlatpakModuleAdvisoryReadyEvent: The event this handler
+            is currently handling.
+        :rtype: list
+        :return: List of images which can be auto rebuilt.
+        """
+        if not conf.pyxis_server_url:
+            raise ValueError("'PYXIS_SERVER_URL' parameter should be set")
+        self._pyxis = Pyxis(conf.pyxis_server_url)
+
+        if not conf.flatpak_server_url or "://" not in conf.flatpak_server_url:
```
hluk, line 112: What's the reason for checking the `://` in URL? I would rather not do any assumptions about format of the URL.

```suggestion
        if not conf.flatpak_server_url:
```

hluk (approved):
One comment, otherwise LGTM.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Actually no strong opinion. Fixing.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,148 @@
...
+        )
+        if not auto_build_images_list or not rebuild_images:
+            msg = "There is no image can be rebuilt. " f"message_id: {event.msg_id}"
+            db_event.transition(EventState.SKIPPED, msg)
+            db.session.commit()
+            self.log_info(msg)
+            return []
+
+        images_nvr = ",".join([rebuild_image.nvr for rebuild_image in rebuild_images])
+        self.log_info("Following images %s will be rebuilt" % images_nvr)
```
qixiang, line 70: Images get from `self._get_auto_rebuild_image_list(event)` are directly affected images, probably you should also check if there are any images need to be rebuilt as dependency, I believe it should not happen with normal flatpak applications images, but it may be possible for flatpak runtime or sdk images, so it would be safe to also check that.

The context behind this is, for example we have two images:
Image `foo`'s Dockerfile:
```
FROM xxx
RUN yum install -y titi
```
Image `bar`'s Dockerfile
```
FROM foo
RUN yum install -y tata
```
In image `foo`, there are both `titi` and `tata` installed, when there is a CVE fix for `titi`, image `bar` is impacted, for some reason, image `foo` is not enabled for auto-rebuild, so only `bar` appears in our rebuild list. However if you only rebuild `bar`, the `titi` rpm in rebuilt image `bar` won't be updated. In such cases, you need to rebuild image `foo` as dependency first, and then use the rebuilt `foo` as parent of `bar`, so there will be two batches of images to be rebuilt:
```
batch1: foo (rebuild_reason: dependency)
batch2: bar (rebuild_reason: directly_affeced)
```
You can do this in another PR.

qixiang: Maybe we can just assume that all flatpak modules are installed from the impacted images returned by flatpak index service at this moment, if that can happen we can fix it later, or you can check this with flatpak team to confirm.

chuazhan: Ok. Will check with Flatpak people.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,148 @@
...
+        self.event = event
+
+        db_event = Event.get_or_create_from_event(db.session, event)
+        self.set_context(db_event)
+
+        auto_build_images_list = self._get_auto_rebuild_image_list(event)
+        rebuild_images = self._filter_images_with_higher_rpm_nvr(
+            event, auto_build_images_list
+        )
+        if not auto_build_images_list or not rebuild_images:
```
gnaponie, line 62: Is there a reason why you need to check both `auto_build_images_list` and `rebuild_images`? Shouldn't `rebuild_images` be a filtered version of `rebuild_images`?

chuazhan: auto_build_images_list are the images which have the auto build tag. If there are auto_build_images_list, then we will check them against the rpm version in the advisory's.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,148 @@
...
+from freshmaker.kojiservice import koji_service
+from freshmaker.lightblue import LightBlue
+from freshmaker.models import Event
+from freshmaker.pyxis import Pyxis
+from freshmaker.types import EventState
 
 
 class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
-    name = 'RebuildFlatpakApplicationOnModuleReady'
+    name = "RebuildFlatpakApplicationOnModuleReady"
```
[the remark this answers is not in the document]
chuazhan: I think ready means the Flatpak module advisory is in QE status and all attach build is in signed status, then we can trigger to rebuild Flatpack application.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: I got the answer from Owen Taylor "The index is returning all the images that should be rebuilt - there is no docker-style image layered / derived images with Flatpaks.". So let's rebuild the images flatpak index service returned.

chuazhan:
@qixiang @gnaponie Can you help to review again as the question has been answered and we don't need to change.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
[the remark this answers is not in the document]
gnaponie: Optional change: maybe let's add a comment (even somewhere else) that explain that?

gnaponie (approved):
Let's wait for @qixiang's review, but it looks good to me.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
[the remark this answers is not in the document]
chuazhan: Ok. Updated. Add the comment before the naming

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,150 @@
...
+from freshmaker.lightblue import LightBlue
+from freshmaker.models import Event
+from freshmaker.pyxis import Pyxis
+from freshmaker.types import EventState
 
 
 class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
-    name = 'RebuildFlatpakApplicationOnModuleReady'
+    # Module ready means Flatpak module advisory is in QE status
+    # and all attach build is in signed status
```
qixiang, line 43:
```suggestion
    # and all attach builds are signed
```

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,150 @@
...
+
+        db_event = Event.get_or_create_from_event(db.session, event)
+        self.set_context(db_event)
+
+        auto_build_images_list = self._get_auto_rebuild_image_list(event)
+        rebuild_images = self._filter_images_with_higher_rpm_nvr(
+            event, auto_build_images_list
+        )
+        if not auto_build_images_list or not rebuild_images:
+            msg = "There is no image can be rebuilt. " f"message_id: {event.msg_id}"
```
qixiang, line 65: It'd be better to log different reasons of no image to rebuild:
1. No images are impacted by the advisory
2. Images are not enabled for auto rebuild.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,150 @@
...
+                mmd = session.get_modulemd(module_nvr)
+                content_index_url = "{}/released/contents/modules/{}:{}.json".format(
+                    conf.flatpak_server_url,
+                    mmd.get_module_name(),
+                    mmd.get_stream_name(),
+                )
+                req_session = self._get_requests_session()
+                response = req_session.get(content_index_url)
+                status_code = response.status_code
+                if status_code == HTTPStatus.OK:
```
qixiang, line 136: Probably we can just check `response.ok` (`True` if `status_code` is less than 400) instead of comparing the status code with `HTTPStatus.OK` (value == 200).
```suggestion
                if response.ok:
```

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -20,13 +20,150 @@
...
+            conf.koji_profile, log, login=False, dry_run=self.dry_run
+        ) as session:
+            for module_nvr in module_nvrs:
+                mmd = session.get_modulemd(module_nvr)
+                content_index_url = "{}/released/contents/modules/{}:{}.json".format(
+                    conf.flatpak_server_url,
+                    mmd.get_module_name(),
+                    mmd.get_stream_name(),
+                )
+                req_session = self._get_requests_session()
```
qixiang, line 133: It'd be better to move this to a separate API or even API in a new module, and decorator it with `freshmaker.utils.retry` or handle the network error exceptions.

qixiang (approved):
LGTM, I have some minor comments, but since you'll have further patches to this, I'm fine with improvements in coming PRs.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ack. Fixing

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ack. Fixing

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ack. Fixing

chuazhan:
As @qixiang mentions, let me merge this first and change the comments in coming patch

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ok. Let me push the coming patch first and fix this at the meanwhile.

PR #146, opened 2021-12-02
Change ImageGroup methods to be more efficient

ronnyhlim:
This change is in regards to the comment mentioned from https://github.com/redhat-exd-rebuilds/freshmaker/pull/122

**tl;dr**:
It'd be simpler and more efficient to change this:
``` 
return all(
            [self.name == other.name and self.version == other.version and self.repos == other.repos]
        )
```
to
```
return self.name == other.name and self.version == other.version and self.repos == other.repos
```
because as of now, we are creating a list with a single boolean value and then calling the all function against it.


qixiang (approved):
:+1:

PR #149, opened 2021-12-07
Update requirements for a security bug

gnaponie:
Bug ref: https://github.com/python-ldap/python-ldap/issues/424


qixiang (approved):
:+1:

PR #150, opened 2021-12-08
Find bundles by digests from all available arches

qixiang:
Freshmaker find bundles by the manifest schema2 digest, but this
digest values are different for arches, seems bundle_path_digest
is the one from x86_64 image, but not sure about this, so the 
safest way is querying with digests from all available arches.


apaplaus (approved):

gnaponie:
Looks good +1

gnaponie (approved):

PR #151, opened 2021-12-09
Don't check auto rebuild tags for bundles specified in manual rebuild

qixiang:
Bundles specified explicitly in manual rebuild request can include
unpublished images, freshmaker should not check the auto rebuild
tags for such bundles.


gnaponie:
looks good

PR #154, opened 2021-12-14
Resolve and record builds in database

chuazhan:
Add new method to resolve odcs compose id for flatpak
image.

Add logic to filter out compose in sets renew and new.
Link composes to build and record builds in database.

JIRA: RHELWF-3780

### DIFF
--- a/freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
+++ b/freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
@@ -22,8 +22,10 @@
 # Written by Valerij Maljulin <vmaljuli@redhat.com>
 # Written by Chuang Zhang <chuazhan@redhat.com>
 
+import json
+import koji
 import requests
-from http import HTTPStatus
+from odcs.common.types import PungiSourceType
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
 
@@ -33,14 +35,15 @@ from freshmaker.events import FlatpakModuleAdvisoryReadyEvent
 from freshmaker.handlers import ContainerBuildHandler, fail_event_on_handler_exception
 from freshmaker.kojiservice import koji_service
 from freshmaker.lightblue import LightBlue
-from freshmaker.models import Event
+from freshmaker.models import Event, Compose
+from freshmaker.odcsclient import create_odcs_client
 from freshmaker.pyxis import Pyxis
-from freshmaker.types import EventState
+from freshmaker.types import ArtifactType, ArtifactBuildState, EventState, RebuildReason
 
 
 class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
     # Module ready means Flatpak module advisory is in QE status
-    # and all attach build is in signed status
+    # and all attached builds are signed.
     name = "RebuildFlatpakApplicationOnModuleReady"
 
     def can_handle(self, event):
@@ -57,12 +60,23 @@ class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
         db_event = Event.get_or_create_from_event(db.session, event)
         self.set_context(db_event)
 
-        auto_build_images_list = self._get_auto_rebuild_image_list(event)
-        rebuild_images = self._filter_images_with_higher_rpm_nvr(
-            event, auto_build_images_list
+        self.errata = Errata()
+        self.advisory_module_nvrs = self.errata.get_cve_affected_build_nvrs(
+            event.advisory.errata_id, True
         )
-        if not auto_build_images_list or not rebuild_images:
-            msg = "There is no image can be rebuilt. " f"message_id: {event.msg_id}"
+
+        rebuild_images = []
+        auto_build_image_modules_mapping = self._get_auto_rebuild_image_mapping()
+        if auto_build_image_modules_mapping:
+            images = list(auto_build_image_modules_mapping.keys())
+            rebuild_images = self._filter_images_with_higher_rpm_nvr(images)
+        if not auto_build_image_modules_mapping or not rebuild_images:
+            msg = (
+                "Images are not enabled for auto rebuild. "
+                if not auto_build_image_modules_mapping
+                else "No images are impacted by the advisory. "
+            )
+            msg = f"{msg} message_id: {event.msg_id}"
             db_event.transition(EventState.SKIPPED, msg)
             db.session.commit()
             self.log_info(msg)
@@ -70,6 +84,14 @@ class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
 
         images_nvr = ",".join([rebuild_image.nvr for rebuild_image in rebuild_images])
         self.log_info("Following images %s will be rebuilt" % images_nvr)
+
+        builds = self._record_builds(rebuild_images, auto_build_image_modules_mapping)
+        if not builds:
+            msg = "No container images to rebuild for advisory %r" % event.advisory.name
+            self.log_info(msg)
+            db_event.transition(EventState.SKIPPED, msg)
+            db.session.commit()
+            return []
 
         # TODO: Return empty list so far as this method is still in progress.
         # Will think about whether to return the real BaseEvent objects in future.
@@ -98,42 +120,35 @@ class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
         session.mount("https://", adapter)
         return session
 
-    def _get_auto_rebuild_image_list(self, event):
+    def _get_auto_rebuild_image_mapping(self):
         """
-        Get rebuild image list which can be auto rebuilt.
+        Get rebuild image mapping which images can be auto rebuilt.
 
-        :param event FlatpakModuleAdvisoryReadyEvent: The event this handler
-            is currently handling.
-        :rtype: list
-        :return: List of images which can be auto rebuilt.
+        :rtype: dict
+        :return: A dict which key is the image which can be rebuilt
+            and value are modules which are enabled in this image.
         """
         if not conf.pyxis_server_url:
             raise ValueError("'PYXIS_SERVER_URL' parameter should be set")
         self._pyxis = Pyxis(conf.pyxis_server_url)
 
         if not conf.flatpak_server_url:
-            raise ValueError(
-                "'FLATPAK_SERVER_URL' parameter should be set"
-            )
+            raise ValueError("'FLATPAK_SERVER_URL' parameter should be set")
 
-        errata = Errata()
-        module_nvrs = errata.get_cve_affected_build_nvrs(event.advisory.errata_id, True)
-        rebuild_images_list = list()
-
+        image_modules_mapping = {}
+        req_session = self._get_requests_session()
         with koji_service(
             conf.koji_profile, log, login=False, dry_run=self.dry_run
         ) as session:
-            for module_nvr in module_nvrs:
-                mmd = session.get_modulemd(module_nvr)
+            for advisory_module_nvr in self.advisory_module_nvrs:
+                mmd = session.get_modulemd(advisory_module_nvr)
                 content_index_url = "{}/released/contents/modules/{}:{}.json".format(
                     conf.flatpak_server_url,
                     mmd.get_module_name(),
                     mmd.get_stream_name(),
                 )
-                req_session = self._get_requests_session()
                 response = req_session.get(content_index_url)
-                status_code = response.status_code
-                if status_code == HTTPStatus.OK:
+                if response.ok:
                     images_info = response.json().get("Images", [])
                     for image_info in images_info:
                         image_nvr = image_info["ImageNvr"]
@@ -141,29 +156,211 @@ class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
                             image_nvr
                         )
                         if is_tagged_auto_build:
-                            rebuild_images_list.append(image_nvr)
+                            image_modules_mapping.setdefault(image_nvr, [])
+                            if (
+                                advisory_module_nvr
+                                not in image_modules_mapping[image_nvr]
+                            ):
+                                image_modules_mapping[image_nvr].append(
+                                    advisory_module_nvr
+                                )
                 else:
-                    self.log_error("Fetching module %s data failed.", module_nvr)
+                    self.log_error(
+                        "Fetching module %s data failed.", advisory_module_nvr
+                    )
 
-        return rebuild_images_list
+        return image_modules_mapping
 
-    def _filter_images_with_higher_rpm_nvr(self, event, rebuild_images_list):
+    def _filter_images_with_higher_rpm_nvr(self, images):
         """
-        Filter out images which have higher nvr in rebuild_images_list than
+        Filter out images which have higher NVR in rebuild_images than
         the advisory's.
 
-        :param event FlatpakModuleAdvisoryReadyEvent: The event this handler
-            is currently handling.
+        :param images list: A list of image NVRS.
+        :return: a list of ContainerImage instances which can be auto rebuilt.
         :rtype: list
-        :return: List of ContainerImage instances which can be auto rebuilt.
         """
-        errata = Errata()
-        errata_rpm_nvrs = errata.get_cve_affected_rpm_nvrs(event.advisory.errata_id)
+        errata_rpm_nvrs = self.errata.get_cve_affected_rpm_nvrs(
+            self.event.advisory.errata_id
+        )
         lb = LightBlue(
             server_url=conf.lightblue_server_url,
             cert=conf.lightblue_certificate,
             private_key=conf.lightblue_private_key,
             event_id=self.current_db_event_id,
         )
-        images = lb.get_images_by_nvrs(rebuild_images_list, rpm_nvrs=errata_rpm_nvrs)
+        images = lb.get_images_by_nvrs(images, rpm_nvrs=errata_rpm_nvrs)
         return images
+
+    def _outdated_composes(self, original_odcs_compose_ids, module_name_stream_set):
+        """
+        Generate outdated composes.
+
+        :param original_odcs_compose_ids list: Original compose ids.
+        :param module_name_stream_set set: Module name stream set from an advisory.
+        :return: a set of outdated composes.
+        :rtype: set
+        """
+        outdated_composes = set()
+        for compose_id in original_odcs_compose_ids:
+            compose = self.odcs.get_compose(compose_id)
+            source_type = compose.get("source_type")
+            if source_type != PungiSourceType.MODULE:
+                outdated_composes.add(compose_id)
+                continue
+
+            name_stream_set = {f"{n}:{s}" for n, s, v, c in _compose_sources(compose)}
+
+            if name_stream_set.isdisjoint(module_name_stream_set):
+                outdated_composes.add(compose_id)
+
+        return outdated_composes
+
+    def _missing_composes(
+        self,
+        original_odcs_compose_ids,
+        module_name_stream_set,
+        module_name_stream_version_set,
+    ):
+        """
+        Generate missing composes.
+
+        :param original_odcs_compose_ids list: Original compose ids.
+        :param module_name_stream_set set: Module name stream set from an advisory.
+        :param module_name_stream_version_set set: Module name stream version set from an advisory.
+        :return: a set of missing composes.
+        :rtype: set
+        """
+        missing_composes = set()
+        for compose_id in original_odcs_compose_ids:
+            compose = self.odcs.get_compose(compose_id)
+            source_type = compose.get("source_type")
+            if source_type == PungiSourceType.MODULE:
+                name_stream_set = {f"{n}:{s}" for n, s, v, c in _compose_sources(compose)}
+                mapping = {f"{n}:{s}": f"{n}:{s}:{v}" for n, s, v, c in _compose_sources(compose)}
+
+                if not name_stream_set.isdisjoint(module_name_stream_set):
+                    missing_composes.update(
+                        mapping[name_stream]
+                        for name_stream in name_stream_set.difference(module_name_stream_set)
+                    )
+                missing_composes.update(module_name_stream_version_set)
+
+        return missing_composes
+
+    def _record_builds(self, images, auto_build_image_modules_mapping):
+        """
+        Records the images to database.
+
+        :param images list: a list of ContainerImage instances.
+        :param auto_build_image_modules_mapping dict: a dict which key is the
+            original image's NVR and value is modules list enabled in the image.
+        :return: a mapping between docker image build NVR and
+            corresponding ArtifactBuild object representing a future rebuild of
+            that docker image. It is extended by including those docker images
+            stored into database.
+        :rtype: dict
+        """
+        db_event = Event.get_or_create_from_event(db.session, self.event)
+
+        # Cache for ODCS module composes. Key is white-spaced, sorted, list
+        # of module's NAME:STREAM:VERSION. Value is Compose database object.
+        odcs_cache = {}
+
+        # Dict with {brew_build_nvr: ArtifactBuild, ...} mapping.
+        builds = {}
+
+        with koji_service(
+            conf.koji_profile, log, login=False, dry_run=self.dry_run
+        ) as session:
+            for image in images:
+                self.set_context(db_event)
+
+                image.resolve_commit()
+                image.resolve_original_odcs_compose_ids(False)
+                nvr = image.nvr
+                image_name = koji.parse_NVR(nvr)["name"]
+                build = self.record_build(
+                    self.event,
+                    image_name,
+                    ArtifactType.IMAGE,
+                    state=ArtifactBuildState.PLANNED.value,
+                    original_nvr=nvr,
+                    rebuild_reason=RebuildReason.DIRECTLY_AFFECTED.value,
+                )
+                # Set context to particular build so logging shows this build
+                # in case of error.
+                self.set_context(build)
+
+                module_name_stream_version_set = set()
+                module_name_stream_set = set()
+                module_nvrs = auto_build_image_modules_mapping[nvr]
+                for module_nvr in module_nvrs:
+                    mmd = session.get_modulemd(module_nvr)
+                    name = mmd.get_module_name()
+                    stream = mmd.get_stream_name()
+                    version = mmd.get_version()
+                    module_name_stream_set.add("%s:%s" % (name, stream))
+                    module_name_stream_version_set.add(
+                        ":".join([name, stream, str(version)])
+                    )
+                original_odcs_compose_ids = image["original_odcs_compose_ids"]
+                outdated_composes = self._outdated_composes(
+                    original_odcs_compose_ids, module_name_stream_set
+                )
+
+                build.build_args = json.dumps(
+                    {
+                        "repository": image["repository"],
+                        "commit": image["commit"],
+                        "target": image["target"],
+                        "branch": image["git_branch"],
+                        "arches": image["arches"],
+                        "renewed_odcs_compose_ids": list(outdated_composes),
+                        "flatpak": image.get("flatpak", False),
+                        "isolated": image.get("isolated", True),
+                    }
+                )
+                db.session.commit()
+
+                missing_composes = self._missing_composes(
+                    original_odcs_compose_ids,
+                    module_name_stream_set,
+                    module_name_stream_version_set,
+                )
+                for compose_source in missing_composes:
+                    if compose_source in odcs_cache:
+                        db_compose = odcs_cache[compose_source]
+                    else:
+                        compose = create_odcs_client().new_compose(
+                            compose_source, "module"
+                        )
+                        db_compose = Compose(odcs_compose_id=compose["id"])
+                        db.session.add(db_compose)
+                        db.session.commit()
+                        odcs_cache[compose_source] = db_compose
+
+                    if db_compose:
+                        build.add_composes(db.session, [db_compose])
+                        db.session.commit()
+                builds[nvr] = build
+
+            # Reset context to db_event.
+            self.set_context(db_event)
+
+        return builds
+
+
+def _compose_sources(compose):
+    """
+    Get compose sources .
+
+    :param compose dict: Compose info from ODCS.
+    :return: a list of lists in format [[N1, S1, V1, C1], [N2, S2, V2, C2]...].
+    :rtype: list
+    """
+    source_value = compose.get("source", "")
+    sources = []
+    for source in source_value.split():
+        sources.append(source.strip().split(":"))
+    return sources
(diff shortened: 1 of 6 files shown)


chuazhan:
@qixiang @hluk @gnaponie Please help to review, thanks!

tests/handlers/koji/test_rebuild_flatpak_application_on_module_ready.py
```diff
@@ -178,5 +180,172 @@ def test_event_state_updated_when_no_images_with_higher_rpm_nvr(self):
...
         )
+
+    def test_get_module_name_stream_set(self):
+        build = {
+            "build_id": 1850907,
+            "epoch": None,
+            "extra": {
+                "typeinfo": {
+                    "module": {
+                        "modulemd_str": '---\ndocument: modulemd\nversion: 2\ndata:\n  name: ghc\n  stream: "9.2"\n  version: 3620211101111632\n  context: d099bf28\n  summary: Haskell GHC 9.2\n  description: >-\n    This module provides the Glasgow Haskell Compiler version 9.2.1\n',
```
gnaponie, line 193: could you put this string on multiple lines for readability?

chuazhan: Ok. Fixed.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -167,3 +179,152 @@ def _filter_images_with_higher_rpm_nvr(self, event, rebuild_images_list):
         )
         images = lb.get_images_by_nvrs(rebuild_images_list, rpm_nvrs=errata_rpm_nvrs)
         return images
+
+    def _get_module_name_stream_set(self, event):
```
qixiang, line 183: There is `self.event` can be used to get the advisory id, unnecessary to have `event` in api params.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -167,3 +179,152 @@ def _filter_images_with_higher_rpm_nvr(self, event, rebuild_images_list):
...
+                ArtifactType.IMAGE,
+                state=ArtifactBuildState.PLANNED.value,
+                original_nvr=nvr,
+                rebuild_reason=RebuildReason.DIRECTLY_AFFECTED.value,
+            )
+            # Set context to particular build so logging shows this build
+            # in case of error.
+            self.set_context(build)
+
+            module_name_stream_set = self._get_module_name_stream_set(event)
```
qixiang, line 290: For every image, you're calling this API and the underline errata API once which is unnecessary.

freshmaker/lightblue.py
```diff
@@ -467,6 +467,34 @@ def resolve(self, lb_instance, children=None):
             err = "Cannot resolve the container image: %s" % e
             self.log_error(err)
 
+    def resolve_flatpak_odcs_compose_ids(self):
```
qixiang, line 470: There is not much different between this API and `resolve_original_odcs_compose_ids`, could you just add a param to the original API:
```
def resolve_original_odcs_compose_ids(self, from_most_original_build=True):
    ...
    original_nvr = self.nvr
    if from_most_original_build:
        original_nvr =  ArtifactBuild.get_most_original_nvr(self.nvr)
    ...
```

tests/handlers/koji/test_rebuild_flatpak_application_on_module_ready.py
```diff
@@ -178,5 +180,174 @@ def test_event_state_updated_when_no_images_with_higher_rpm_nvr(self):
...
+        get_build.return_value = build
+        get_cve_affected_build_nvrs = self._patch(
+            "freshmaker.errata.Errata.get_cve_affected_build_nvrs"
+        )
+        get_cve_affected_build_nvrs.return_value = ["ghc-9.2-3620211101111632.d099bf28"]
+        ret = self.handler._get_module_name_stream_set(self.event)
+        self.assertEqual(ret, {"ghc:9.2"})
+
+    @patch("freshmaker.odcsclient.create_odcs_client")
+    def test_prepare_data_for_compose(self, create_odcs_client):
```
gnaponie, line 219: It seems like there's a lot of repeated code in this test and it becomes hard to read it. Any chance we can shorten this a bit? Consider using `parametrize` or maybe a for loop.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -167,3 +179,152 @@ def _filter_images_with_higher_rpm_nvr(self, event, rebuild_images_list):
...
+            for source in source_value.split():
+                group = source.strip().split(":")[:2]
+                name_stream_set.add(group[0] + ":" + group[1])
+
+            # If all the sources are not in advisory, renew them
+            if len(name_stream_set - module_name_stream_set) == len(name_stream_set):
+                renewed_odcs_compose_set.add(compose_id)
+            else:
+                # If some source in advisory, create composes for all separately
+                new_odcs_compose_set.update(name_stream_set)
```
qixiang, line 244: I believe odcs module compose source should be module nsvc instead of just name and stream, can you check this?

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -33,14 +34,15 @@
...
+from freshmaker.odcsclient import create_odcs_client
 from freshmaker.pyxis import Pyxis
-from freshmaker.types import EventState
+from freshmaker.types import ArtifactType, ArtifactBuildState, EventState, RebuildReason
 
 
 class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
     # Module ready means Flatpak module advisory is in QE status
-    # and all attach build is in signed status
+    # and all attach build are signed.
```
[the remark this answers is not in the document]
chuazhan: Ack. Fixing

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -167,3 +179,152 @@ def _filter_images_with_higher_rpm_nvr(self, event, rebuild_images_list):
...
+        """
+        errata = Errata()
+        module_nvrs = errata.get_cve_affected_build_nvrs(event.advisory.errata_id, True)
+
+        module_name_stream_set = set()
+        with koji_service(
+            conf.koji_profile, log, login=False, dry_run=self.dry_run
+        ) as session:
+            for module_nvr in module_nvrs:
+                mmd = session.get_modulemd(module_nvr)
```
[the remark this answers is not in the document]
chuazhan: Actually I was confused with the [doc](https://fedora-modularity.github.io/libmodulemd/latest/modulemd-2.0-Modulemd.html#modulemd-read-packager-string) too. But when I was testing it, the returned value is a _ResultTuple like "(<GType ModulemdModuleStreamV2 (94271507489296)>, object=<Modulemd.ModuleStreamV2 object at 0x7fe3eea26700 (ModulemdModuleStreamV2 at 0x55bd4bb78940)>)"

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ack. Fixing

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Sure. Fixing.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Yes. When I use method GET, I can see the source format is nvsc, but I checked the [doc](https://pagure.io/odcs/#:~:text=Requesting%20new%20compose), when we create a new compose, for module source type, the source can be "White-space separated NAME:STREAM or NAME:STREAM:VERSION of modules to include in compose." I am wondering if the VERSION should be created automatically.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -167,3 +179,152 @@ def _filter_images_with_higher_rpm_nvr(self, event, rebuild_images_list):
...
+            is a set of required data for creating new composes
+        :rtype: tuple
+        """
+        renewed_odcs_compose_set = set()
+        new_odcs_compose_set = set()
+        for compose_id in original_odcs_compose_ids:
+            compose = self.odcs.get_compose(compose_id)
+            source_type = compose.get("source_type")
+            # source_type of module composes is 2
+            if source_type != 2:
```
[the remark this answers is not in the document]
chuazhan: Yeah, looks like a magic number, but this code imitates the pulp compose which source_type is 4 and that doesn't exist in ArtifactType. Do we need to create a new one, looks like ComposeSourceType, @qixiang @gnaponie any idea?

tests/handlers/koji/test_rebuild_flatpak_application_on_module_ready.py
chuazhan: The data has a small difference. Let me separate the test cases for readability.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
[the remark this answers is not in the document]
qixiang: ODCS compose type is different from freshmaker `ArtifactType`, you can use the types from ODCS:
```
>>> from odcs.common.types import PungiSourceType
>>> PungiSourceType.MODULE
2
```

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
qixiang: If you request compose with only module `name` + `stream`, I guess ODCS should get the latest modules from that `name` and `stream`, which can be different from the modules in the trigger advisory, so I believe at least `version` is also required, but I'm not sure about `context`, I'd suggest checking with ODCS team if that is not documented.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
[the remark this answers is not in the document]
chuazhan: Ack. Thanks

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ok. Checked with ODCS people and as you mentioned name + stream get the latest modules. Adding version as well

freshmaker/lightblue.py
chuazhan: Ok. Try to fix it.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -167,3 +179,152 @@ def _filter_images_with_higher_rpm_nvr(self, event, rebuild_images_list):
...
+        """
+        Get set of modules with format NAME:STREAM in an advisory.
+
+        :param event FlatpakModuleAdvisoryReadyEvent: The event this handler
+            is currently handling.
+        :return: a set with item NAME:STREAM
+        :rtype: set
+        """
+        errata = Errata()
+        module_nvrs = errata.get_cve_affected_build_nvrs(event.advisory.errata_id, True)
```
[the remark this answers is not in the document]
chuazhan: Fix the former part, not quite sure about why to save the mapping between nvr and nsvc

tests/handlers/koji/test_rebuild_flatpak_application_on_module_ready.py
```diff
@@ -178,5 +188,159 @@ def test_event_state_updated_when_no_images_with_higher_rpm_nvr(self):
...
+        self.assertEqual(ret, {"ghc:9.2:3620211101111632"})
+
+    @patch("freshmaker.odcsclient.create_odcs_client")
+    def test_prepare_data_for_compose_not_module_type(self, create_odcs_client):
+        odcs = create_odcs_client.return_value
+        # Test for compose source_type not module type
+        odcs.get_compose.return_value = {
+            "arches": "x86_64",
+            "id": 985716,
+            "owner": "auto/liver.englab.brq.redhat.com",
```
gnaponie, line 206: could you make a fake url instead of "redhat" (since this is a public repo)?

gnaponie:
I've just left one comment, but the change looks good to me +1

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -167,3 +180,149 @@ def _filter_images_with_higher_rpm_nvr(self, event, rebuild_images_list):
...
+            name_stream_set = set()
+            source_value = compose.get("source", "")
+            for source in source_value.split():
+                group = source.strip().split(":")[:3]
+                name_stream_set.add(":".join(group))
+
+            # If all the sources are not in advisory, renew them
+            if len(name_stream_set - module_name_stream_version_set) == len(
+                name_stream_set
+            ):
```
qixiang, line 240: Could you explain on this? I can't understand why this means all sources are not in advisory. And `version` should not be included while checking this.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -167,3 +180,149 @@ def _filter_images_with_higher_rpm_nvr(self, event, rebuild_images_list):
...
+                renewed_odcs_compose_set.add(compose_id)
+            else:
+                # If some source in advisory, create composes for all separately
+                new_odcs_compose_set.update(name_stream_set)
```
qixiang, line 244: So when some sources are in advisory, you'll create new composes for all these old modules, right? I believe this is wrong, for example, there is old modules:

- module-a:stable:123
- module-b:el8:345

And now you have module-b:el8:350 in advisory, you should request 2 new module composes (in case you want to create one compose for each module):
- compose: module-a:stable:123
- compose: module-b:el8:350

compose for `module-b:el8:345` should be excluded, in your code, there will be 3 composes with all these modules included.

qixiang (changes requested):
It overall looks good to me, but still have something need to be fixed, especially the code of getting new odcs module composes.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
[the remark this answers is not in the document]
qixiang: I forgot this, could you try it in our freshmaker image? I guess it could be caused by some different version of packages, but I didn't dig into this, error from what I tried with `quay.io/factory2/freshmaker:prod`:
```
>  _, mmd = Modulemd.read_packager_string(modulemd_str, None, None)
                                                                                                                         
TypeError: cannot unpack non-iterable ModuleStreamV2 object
```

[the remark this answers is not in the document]
chuazhan: Ack. Fixed.
Add root cause for others, seems rpm package install is different with pip install, rpm will provide overrides which will help to unpack and strip GType, but pip install will not, so the return value is different.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -167,3 +180,149 @@ def _filter_images_with_higher_rpm_nvr(self, event, rebuild_images_list):
...
+            source_type = compose.get("source_type")
+            if source_type != PungiSourceType.MODULE:
+                renewed_odcs_compose_set.add(compose_id)
+                continue
+
+            name_stream_set = set()
+            source_value = compose.get("source", "")
+            for source in source_value.split():
+                group = source.strip().split(":")[:3]
+                name_stream_set.add(":".join(group))
```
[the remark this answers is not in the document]
chuazhan: Ack. Fixed. Actually these part has been refactored

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ok. As discussed offline, has change this part with intersection and difference, please check the latest codes

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ack. Fixed

chuazhan:
@qixiang @chuazhan @hluk Updated, please help to review

chuazhan:
@qixiang @gnaponie @hluk Can you help to take a look? Thanks

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -167,3 +180,156 @@ def _filter_images_with_higher_rpm_nvr(self, event, rebuild_images_list):
...
+            if source_type != PungiSourceType.MODULE:
+                renewed_odcs_compose_set.add(compose_id)
+                continue
+
+            name_stream_set = set()
+            mapping = dict()
+            source_value = compose.get("source", "")
+            for source in source_value.split():
+                source = source.strip()
+                name_stream = source.rsplit(":", 2)[0]
```
qixiang, line 237: Is it possible that the module compose was generated from `<name>:<stream>`/`<name>:<stream>:<version>` in old build?

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -167,3 +180,156 @@ def _filter_images_with_higher_rpm_nvr(self, event, rebuild_images_list):
...
+            inter_section = name_stream_set.intersection(module_name_stream_set)
+            if not inter_section:
+                renewed_odcs_compose_set.add(compose_id)
+            else:
+                # If some module in advisory, create composes for all differences,
+                # inter_section part will be added in following step
+                difference = name_stream_set.difference(module_name_stream_set)
+                for item in difference:
+                    new_odcs_compose_set.add(mapping[item])
+            # All modules in advisory should create compose
```
qixiang, line 252: Why all modules in advisory should be enabled for all impacted images? If there are multiple modules in advisory, each module can impact different images, for example, an advisory has `module-a` and `module-b`, `module-a` only impacts `a-flatpak-image`, `module-b` only impacts `b-flatpak-image`, in this case, you only need to request `module-a` compose for image `a-flatpak-image`.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
qixiang: Ignore this, tested and verified odcs will expand the module to `n:s:v:c` even it's requested with just `n:s` or `n:s:v`.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ack. Fixed

chuazhan:
@qixiang @hluk @gnaponie  Please help to review, thanks!

qixiang (approved):
LGTM

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -57,20 +60,38 @@ def handle(self, event):
...
+                else "No images are impacted by the advisory. "
+            )
+            msg = f"{msg} message_id: {event.msg_id}"
             db_event.transition(EventState.SKIPPED, msg)
             db.session.commit()
             self.log_info(msg)
             return []
 
-        images_nvr = ",".join([rebuild_image.nvr for rebuild_image in rebuild_images])
+        images_nvr = ",".join([rebuild_image.nvr() for rebuild_image in rebuild_images])
```
hluk, line 84: I might be wrong, but this change looks like a typo and seems there are no tests to catch this - i.e. missing test where `_filter_images_with_higher_rpm_nvr()` returns non-empty list.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -98,72 +119,219 @@ def _get_requests_session(cls, retry_options={}):
...
             server_url=conf.lightblue_server_url,
             cert=conf.lightblue_certificate,
             private_key=conf.lightblue_private_key,
             event_id=self.current_db_event_id,
         )
-        images = lb.get_images_by_nvrs(rebuild_images_list, rpm_nvrs=errata_rpm_nvrs)
+        images = lb.get_images_by_nvrs(images, rpm_nvrs=errata_rpm_nvrs)
         return images
+
+    def _prepare_data_for_compose(
```
hluk, line 189: This function is a bit difficult to understand. Can you split this into two function, each with simpler return type (`_outdated_composes()`, `_missing_composes()`)?

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -98,72 +119,219 @@ def _get_requests_session(cls, retry_options={}):
...
+                            compose_source, "module"
+                        )
+                        db_compose = Compose(odcs_compose_id=compose["id"])
+                        db.session.add(db_compose)
+                        db.session.commit()
+                        odcs_cache[compose_source] = db_compose
+
+                    if db_compose:
+                        build.add_composes(db.session, [db_compose])
+                        db.session.commit()
```
hluk, line 331: Can there be just a single DB commit at the end of this method?

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Good catch. yeah, It is a typo. I updated the test_record_builds a bit to cover this.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ok. I split it in two methods,  but it seems it will lose some efficiency.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: I am assuming after every db operation, it will do a commit. But I am not sure we will strictly following this.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -98,72 +120,232 @@ def _get_requests_session(cls, retry_options={}):
...
+                    name_stream_set.add(name_stream)
+                    mapping[name_stream] = name_stream_version
+
+                inter_section = name_stream_set.intersection(module_name_stream_set)
+                if inter_section:
+                    # If some module in advisory, create composes for all differences,
+                    # inter_section part will be added in following step
+                    difference = name_stream_set.difference(module_name_stream_set)
+                    for item in difference:
+                        missing_composes.add(mapping[item])
```
hluk, line 245:
```suggestion
                missing_composes.update(
                    mapping[name_stream]
                    for name_stream in name_stream_set.difference(module_name_stream_set)
                )
```

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -98,72 +120,232 @@ def _get_requests_session(cls, retry_options={}):
...
+        for compose_id in original_odcs_compose_ids:
+            compose = self.odcs.get_compose(compose_id)
+            source_type = compose.get("source_type")
+            if source_type == PungiSourceType.MODULE:
+                name_stream_set = set()
+                mapping = dict()
+                source_value = compose.get("source", "")
+                for source in source_value.split():
+                    source = source.strip()
+                    name_stream = source.rsplit(":", 2)[0]
```
hluk, line 234: Can you move this duplicate code to a separate function?

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -98,72 +120,232 @@ def _get_requests_session(cls, retry_options={}):
...
-            for module_nvr in module_nvrs:
-                mmd = session.get_modulemd(module_nvr)
+            for advisory_module_nvr in self.advisory_module_nvrs:
+                mmd = session.get_modulemd(advisory_module_nvr)
                 content_index_url = "{}/released/contents/modules/{}:{}.json".format(
                     conf.flatpak_server_url,
                     mmd.get_module_name(),
                     mmd.get_stream_name(),
                 )
                 req_session = self._get_requests_session()
```
hluk, line 149: This should initialized once outside of the loop.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -98,72 +120,232 @@ def _get_requests_session(cls, retry_options={}):
...
+            name_stream_set = set()
+            source_value = compose.get("source", "")
+            for source in source_value.split():
+                source = source.strip()
+                name_stream = source.rsplit(":", 2)[0]
+                name_stream_set.add(name_stream)
+
+            # If all the sources are not in advisory, renew them
+            inter_section = name_stream_set.intersection(module_name_stream_set)
+            if not inter_section:
```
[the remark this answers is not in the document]
chuazhan: Ok. We don't need to create a new set for checking.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ack. Fixed

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: I think we still need to check if the two sets are joint. I updated the codes, you can have a look with the new codes and see if it is proper.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: advisory_module_nvr is different, so every advisory_module_nvr will be initialized, right?

hluk: What I meant is that creating `requests.Session()` every loop iteration should be avoided.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -98,72 +120,256 @@ def _get_requests_session(cls, retry_options={}):
...
             server_url=conf.lightblue_server_url,
             cert=conf.lightblue_certificate,
             private_key=conf.lightblue_private_key,
             event_id=self.current_db_event_id,
         )
-        images = lb.get_images_by_nvrs(rebuild_images_list, rpm_nvrs=errata_rpm_nvrs)
+        images = lb.get_images_by_nvrs(images, rpm_nvrs=errata_rpm_nvrs)
         return images
+
+    def _parse_compose_source(self, source_value, name_stream_only=True):
```
hluk, line 195: This function hides too many things again. It can be also free function since it does not depend on `self`.

Too simplify you can return/yield the split sources from compose object. So at call sites you would have:

```python
name_stream_set = {f'{n}:{s}' for n,s,v in _compose_sources(compose)}
...
mapping = {f'{n}:{s}':f'{n}:{s}:{v}' for n,s,v in _compose_sources(compose)}
name_stream_set = set(mapping.values())
```

The main things to de-duplicate and move to separate function/method are:
- Getting the sources from compose object
- Splitting into n:s:v

hluk (changes requested):

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -98,72 +120,256 @@ def _get_requests_session(cls, retry_options={}):
...
+        for compose_id in original_odcs_compose_ids:
+            compose = self.odcs.get_compose(compose_id)
+            source_type = compose.get("source_type")
+            if source_type != PungiSourceType.MODULE:
+                outdated_composes.add(compose_id)
+                continue
+
+            name_stream_set, _ = self._parse_compose_source(compose.get("source", ""))
+
+            # If all the sources are not in advisory, renew them
```
hluk, line 236: No need for comment - the code is clear.

```suggestion
```

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -98,72 +120,256 @@ def _get_requests_session(cls, retry_options={}):
...
+        """
+        missing_composes = set()
+        for compose_id in original_odcs_compose_ids:
+            compose = self.odcs.get_compose(compose_id)
+            source_type = compose.get("source_type")
+            if source_type == PungiSourceType.MODULE:
+                name_stream_set, mapping = self._parse_compose_source(compose.get("source", ""), False)
+
+                # If the two sets are joint, union compose's name_stream_version set
+                # and advisory module's name_stream_version set in missing composes.
```
hluk, line 265: No need for comment - it explains what the code does instead of why - the code is clear enough.

```suggestion
```

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Oh. You mean line "req_session = self._get_requests_session()"? I was thinking the difference parts.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ok. I was thinking to emphasize advisory, but module_name_stream_set's meaning can be seen in doc string.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Ack. Removed

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
hluk: Yeah, it something we missed in previous PR.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Got it. Updated

hluk (approved):
LGTM! :+1:

PR #157, opened 2022-01-11
Support manual rebuild of Flatkpak app images

hluk:
JIRA: RHELWF-3860

### DIFF
--- a/freshmaker/events.py
+++ b/freshmaker/events.py
@@ -486,3 +486,9 @@ class ManualBundleRebuildEvent(ErrataBaseEvent):
     @property
     def search_key(self):
         return str(self.advisory.errata_id)
+
+
+class FlatpakApplicationManualBuildEvent(ManualRebuildWithAdvisoryEvent):
+    """
+    Event triggered when manual rebuild of a flatpak is requested.
+    """
--- a/freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
+++ b/freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
@@ -23,6 +23,8 @@
 # Written by Chuang Zhang <chuazhan@redhat.com>
 
 import json
+from collections import defaultdict
+
 import koji
 import requests
 from odcs.common.types import PungiSourceType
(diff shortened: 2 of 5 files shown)


freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -65,33 +94,43 @@ def handle(self, event):
...
+        """Raises SkipEventException if the event should be skipped."""
+        image_modules_mapping = self._image_modules_mapping()
+        if not image_modules_mapping:
+            raise SkipEventException("No images are impacted by the advisory.")
+
+        if event.manual and event.container_images:
+            image_modules_mapping = {
+                image: modules for image, modules in image_modules_mapping.items()
+                if image in event.container_images
+            }
```
qixiang, line 116: Only published images will be present in flatpak index service, but unpublished images can be specified in request's `container_images`, especially when it's requested by release-driver where the `container_images` are always unreleased yet, however flatpak is a little special, it won't be handled by release-driver, product team is responsible for testing and releasing them, so we can ignore this at this moment.

Sometimes the requests can also be sent by person with unreleased images specified as well, I think it's not a priority to support this at this moment, I'd suggest a more meaningful skip reason like:
```
f"None of the specified images are listed in flatpak index service as latest published images impacted by the advisory: {event.container_images}"
```

freshmaker/events.py
```diff
@@ -486,3 +486,10 @@ def __init__(self, msg_id, advisory, container_images,
     @property
     def search_key(self):
         return str(self.advisory.errata_id)
+
+
+class FlatpakApplicationManualBuildEvent(ManualRebuildWithAdvisoryEvent):
+    """
+    Event triggered when Release Driver requests manual rebuild
+    OR when manual rebuild of a flatpak is requested by person.
```
qixiang, line 494: Flatpak advisories won't be handled by release-driver, it's better to update this doc, actually release-driver requests are just manual requests which is not very special, it worth nothing to mention release-driver in any of the manual rebuild event.

qixiang (approved):
LGTM, just some minor comments.

gnaponie:
Agreed with Qixiang's comments. Besides those it looks good.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
hluk: I've updated the skip reason.

freshmaker/events.py
hluk: I removed release-driver mention from the docstring.

chuazhan (approved):
LGTM

qixiang:
:+1:

PR #159, opened 2022-01-21
Make Freshmaker ignore operator hotfix images

ronnyhlim:
CWFHEALTH-908: "Ensure Freshmaker ignores hotfix images and does not try to rebuild them, or use them to trigger other rebuilds."

### DIFF
--- a/freshmaker/handlers/botas/botas_shipped_advisory.py
+++ b/freshmaker/handlers/botas/botas_shipped_advisory.py
@@ -215,6 +215,12 @@ class HandleBotasAdvisory(ContainerBuildHandler):
             # images in related images
             bundle_nvrs = self._get_impacted_bundles(original_digests_by_nvr.values())
 
+        # Create a new list which only contains images that are not hotfixes
+        for nvr in bundle_nvrs[:]:
+            if self._pyxis.is_hotfix_image(nvr):
+                bundle_nvrs.remove(nvr)
+                log.debug("Excluding hotfix bundle: %s", nvr)
+
         if not bundle_nvrs:
             return None, f"No bundle image is impacted by {self.event.advisory.errata_id}, skip."
 
--- a/freshmaker/pyxis.py
+++ b/freshmaker/pyxis.py
@@ -339,3 +339,17 @@ class Pyxis(object):
                 if set(tag['name'] for tag in repo['tags']) & set(auto_rebuild_tags):
                     return True
         return False
+
+    def is_hotfix_image(self, image_nvr):
+        """
+        Checks if image_nvr is a hotfix image
+
+        :param str image_nvr: image_nvr
+        :return: True if image is a hotfix image, otherwise False
+        :rtype: bool
+        """
+        image = self.get_images_by_nvr(image_nvr, include=["data.parsed_data"])
+        if not image:
+            raise Exception("Image %s was not found in Pyxis", image_nvr)
+        # images for different arches contain the same label names, so just check the first image
+        return any(label["name"] == "com.redhat.hotfix" for label in image[0]["parsed_data"]["labels"])
--- a/tests/test_pyxis.py
+++ b/tests/test_pyxis.py
@@ -601,6 +601,36 @@ class TestQueryPyxis(helpers.FreshmakerTestCase):
         is_bundle = self.px.is_bundle("foobar-bundle-1-234")
         self.assertFalse(is_bundle)
 
+    @patch('freshmaker.pyxis.Pyxis._pagination')
+    def test_is_hotfix_image_true(self, page):
+        page.return_value = [{
+            "parsed_data": {
+                "labels": [
+                    {"name": "com.redhat.hotfix", "value": "bz1234567"},
+                    {"name": "architecture", "value": "ppc64le"},
+                    {"name": "brew", "value": "brew_value_1"},
+                    {"name": "repositories", "value": "repositories_value_1"},
+                ]
+            },
+        }]
+        is_hotfix_image = self.px.is_hotfix_image("foobar-bundle-1-234")
+        self.assertTrue(is_hotfix_image)
+
+    @patch('freshmaker.pyxis.Pyxis._pagination')
+    def test_is_hotfix_image_false(self, page):
+        page.return_value = [{
+            "parsed_data": {
+                "labels": [
+                    {"name": "not_com.redhat.hotfix", "value": "bz1234567"},
+                    {"name": "architecture", "value": "ppc64le"},
+                    {"name": "brew", "value": "brew_value_2"},
+                    {"name": "repositories", "value": "repositories_value_2"},
+                ]
+            },
+        }]
+        is_hotfix_image = self.px.is_hotfix_image("foobar-bundle-1-234")
+        self.assertFalse(is_hotfix_image)
+
     @patch('freshmaker.pyxis.Pyxis.get_auto_rebuild_tags')
     @patch('freshmaker.pyxis.Pyxis._pagination')
     def test_image_is_tagged_auto_rebuild(self, page, get_auto_rebuild_tags):
--- a/tests/handlers/botas/test_botas_shipped_advisory.py
+++ b/tests/handlers/botas/test_botas_shipped_advisory.py
@@ -238,6 +238,7 @@ class TestBotasShippedAdvisory(helpers.ModelsTestCase):
 
         self.pyxis().get_manifest_list_digest_by_nvr.side_effect = gmldbn
         self.pyxis().get_operator_indices.return_value = []
+        self.pyxis().is_hotfix_image.return_value = False
         # Doens't matter what this method will return, because we override method
         # that uses return value
         self.pyxis().get_latest_bundles.return_value = ["some", "bundles", "info"]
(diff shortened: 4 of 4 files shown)


ronnyhlim:
hi @qixiang, the source code is finished but I'm not sure how to Mock a Koji call for testing. I tried to duplicate the unit test for "get_ocp_versions_range" but It seems that even when I pass in a value for an "id" in the unit test it is saying that the id field is missing in the test error. Do you have any suggestions?

qixiang:
> hi @qixiang, the source code is finished but I'm not sure how to Mock a Koji call for testing. I tried to duplicate the unit test for "get_ocp_versions_range" but It seems that even when I pass in a value for an "id" in the unit test it is saying that the id field is missing in the test error. Do you have any suggestions?

The test fails happen in `tests/handlers/botas/test_botas_shipped_advisory.py`, not the new test case you added, the existing test cases need to be adjusted because you changed the implementation.

And I'd prefer using Pyxis instead of koji/brew to get the lables.

ronnyhlim:
> > hi @qixiang, the source code is finished but I'm not sure how to Mock a Koji call for testing. I tried to duplicate the unit test for "get_ocp_versions_range" but It seems that even when I pass in a value for an "id" in the unit test it is saying that the id field is missing in the test error. Do you have any suggestions?
> 
> The test failures happen in `tests/handlers/botas/test_botas_shipped_advisory.py`, not the new test case you added, the existing test cases need to be adjusted because you changed the implementation.
> 
> And I'd prefer using Pyxis instead of koji/brew to get the lables.

Changes were made to use Pyxis instead of Koji. Also, I'm not too familiar with the Mock tests in Freshmaker; what updates in the unit tests should be made to accommodate for these changes? I see the issue is with the assertionNotEqual within `test_botas_shipped_advisory.py` but I don't know what to add to make this unit test work :open_mouth:

freshmaker/pyxis.py
```diff
@@ -339,3 +339,32 @@ def image_is_tagged_auto_rebuild(self, nvr):
                 if set(tag['name'] for tag in repo['tags']) & set(auto_rebuild_tags):
                     return True
         return False
+
+    def remove_hotfix_images(self, bundle_nvrs):
```
qixiang, line 343: Please add an API like `is_hotfix_image(self, image_nvr)` in pyxis, then call the new API in some way like:
```
bundle_nvrs = [nvr in bundle_nvrs if not self._pyxis.is_hotfix_image(nvr)]
```

freshmaker/pyxis.py
```diff
@@ -339,3 +339,32 @@ def image_is_tagged_auto_rebuild(self, nvr):
...
+        :rtype: list
+        """
+        for bundle_nvr in bundle_nvrs:
+            images = self.get_images_by_nvr(bundle_nvr, include=["data.parsed_data"])
+            if not images:
+                log.warning("Image %s is not found in Pyxis, ignore it.", bundle_nvr)
+                continue
+
+            hotfix_image_flag = False
+            labels = images["parsed_data"]["labels"]
```
qixiang, line 358: `images` are images for all available arches, you need to check the labels for one of the image (e.g. images[0]["parsed_data"]["labels"], or check if any of the images has the label, `images["parsed_data"]["labels"]` will raise error.
you can just do something like this:
```
# images for different arches should have same label names, so just check the first image
return any(label["name"] == "com.redhat.hotfix" for label in images[0]["parsed_data"]["labels"])
```

qixiang (changes requested):
You can mock the new API call (`is_hotfix_image`) to always return `False` for existing test cases in `tests/handlers/botas/test_botas_shipped_advisory.py`, then add a new test case to test hotfix images are really filtered out in the botas handler.

And for the new API call, you should add new test cases in `test_pyxis.py`.

freshmaker/pyxis.py
ronnyhlim: ahh gotcha I see, calling a method to return a bool and creating a list comprehension off it is a much better way. I'll change my code accordingly!

freshmaker/pyxis.py
ronnyhlim: I see, I added the [0] index in and used your way of any() to return a bool

ronnyhlim:
update: Mocked the `is_hotfix_image` to always return False for existing test cases and added new test cases in test_pyxis.py. Will add more tests tomorrow to test that hotfix_images are really filtered

tests/test_pyxis.py
```diff
@@ -601,6 +601,36 @@ def test_is_bundle_false(self, page):
         is_bundle = self.px.is_bundle("foobar-bundle-1-234")
         self.assertFalse(is_bundle)
 
+    @patch('freshmaker.pyxis.Pyxis._pagination')
+    def test_is_hotfix_image_true(self, page):
+        page.return_value = [{
+            "parsed_data": {
+                "labels": [
+                    {"name": "com.redhat.hotfix", "value": True},
```
qixiang, line 609: `value` should be some string, please use a string instead, not sure about the real format, you can just give it something like `1.2hotfix`.

tests/test_pyxis.py
```diff
@@ -601,6 +601,36 @@ def test_is_bundle_false(self, page):
...
+        }]
+        is_hotfix_image = self.px.is_hotfix_image("foobar-bundle-1-234")
+        self.assertTrue(is_hotfix_image)
+
+    @patch('freshmaker.pyxis.Pyxis._pagination')
+    def test_is_hotfix_image_false(self, page):
+        page.return_value = [{
+            "parsed_data": {
+                "labels": [
+                    {"name": "not_com.redhat.hotfix", "value": True},
```
qixiang, line 624: As above.

qixiang (commented):
LGTM, just minor comments. @ronnyhlim btw, do you know whether normal images (operator and operand images) are also impacted?

tests/test_pyxis.py
ronnyhlim: Found an image with an actual hotfix label - https://brewweb.engineering.redhat.com/brew/archiveinfo?archiveID=5886789
I'll use it for the test value

ronnyhlim:
> LGTM, just minor comments. @ronnyhlim btw, do you know whether normal images (operator and operand images) are also impacted?

So I've asked and was told that an operator hotfix will consist of an operator image (containing the hotfix) and an operator bundle image (to allow index delivery and customer deployment). With that as you mentioned, we would need to change the `rebuild_images_on_rpm_advisory_change.py`. Would we want to update **filter_out_not_allowed_builds()?**

freshmaker/pyxis.py
```diff
@@ -339,3 +339,23 @@ def image_is_tagged_auto_rebuild(self, nvr):
...
+        """
+        image = self.get_images_by_nvr(image_nvr, include=["data.parsed_data"])
+        if not image:
+            raise Exception("Image %s was not found in Pyxis", image_nvr)
+        # images for different arches contain the same label names, so just check the first image
+        if any(label["name"] == "com.redhat.hotfix" for label in image[0]["parsed_data"]["labels"]):
+            log.debug(
+                "Excluding %s from bundle_nvrs because it is a hotfix image",
+                image_nvr,
+            )
```
qixiang, line 359: This log should be in the handler, this API can be used in different cases, not only for excluding hotfix images.

qixiang:
> So I've asked and was told that an operator hotfix will consist of an operator image (containing the hotfix) and an operator bundle image (to allow index delivery and customer deployment). With that as you mentioned, we would need to change the `rebuild_images_on_rpm_advisory_change.py`. Would we want to update **filter_out_not_allowed_builds()?**

I filed CWFHEALTH-1132 to track this.

`filter_out_not_allowed_builds` is used to filter with the configurable rules, which is not suitable for this case, and you will need to get all hotfix images first before processing which is non-operational.

https://github.com/redhat-exd-rebuilds/freshmaker/blob/e23a4ffb8f0317082005ca2ecb3df634a611a025/freshmaker/lightblue.py#L1623 is a good place to filter out these hotfix images.

btw, please avoid pasting internal links to public.

ronnyhlim:
> > So I've asked and was told that an operator hotfix will consist of an operator image (containing the hotfix) and an operator bundle image (to allow index delivery and customer deployment). With that as you mentioned, we would need to change the `rebuild_images_on_rpm_advisory_change.py`. Would we want to update **filter_out_not_allowed_builds()?**
> 
> I filed CWFHEALTH-1132 to track this.
> 
> `filter_out_not_allowed_builds` is used to filter with the configurable rules, which is not suitable for this case, and you will need to get all hotfix images first before processing which is non-operational.
> 
> https://github.com/redhat-exd-rebuilds/freshmaker/blob/e23a4ffb8f0317082005ca2ecb3df634a611a025/freshmaker/lightblue.py#L1623
> 
> is a good place to filter out these hotfix images.
> btw, please avoid pasting internal links to public.

gotcha, all changes have been made and internal links have been removed!

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -215,6 +215,15 @@ def _get_bundles_to_rebuild(self):
             # images in related images
             bundle_nvrs = self._get_impacted_bundles(original_digests_by_nvr.values())
 
+        def is_hotfix(nvr):
```
qixiang, line 218: It's unnecessary to introduce a nested function, but I'm fine with this.

freshmaker/lightblue.py
```diff
@@ -1621,6 +1621,10 @@ def find_images_to_rebuild(
             rpm_nvrs, content_sets, filter_fnc, published,
             release_categories, leaf_container_images=leaf_container_images)
 
+        # Create a new list of images which contain non-hotfix images
+        is_hotfix_img = any(label["name"] == "com.redhat.hotfix" for label in images[0]["parsed_data"]["labels"])
```
qixiang, line 1625: iirc, the images here are images of all NVRs, each one represent a NVR with only one arch, you're checking the first image instead of all images. I'd prefer you submit another PR for CWFHEALTH-1132.

qixiang (changes requested):

freshmaker/lightblue.py
ronnyhlim: I actually had just noticed this as I pushed it up. Essentially, I have to remove the [0] correct?

And sure I can submit a separate PR for this

qixiang: >  remove the [0]

You need to check labels in each images, it should be something like:
```
images = [
    img
    for img in images
    if any(label["name"] == "com.redhat.hotfix" for label in img["parsed_data"]["labels"])
]
```
but I didn't test it, you need to test and verify this with the images data.

freshmaker/handlers/botas/botas_shipped_advisory.py
ronnyhlim: Ahh, I used a separate function because I didn't know if you could implement a logger within a list comprehension. I don't mind changing it but I don't see anything online about that

freshmaker/lightblue.py
ronnyhlim: no problem, I'll write the code to check each image. I assigned CWFHEALTH-1132 to myself and will work on that after this  PR gets merged

freshmaker/handlers/botas/botas_shipped_advisory.py
qixiang: You don't have to insist with list comprehension, you can do it in loop:
```
for nvr in bundle_nvrs[:]:
    if self._pyxis.is_hotfix_image(nvr):
        bundle_nvrs.remove(nvr)
        log.debug("Excluding hotfix bundle: %s", nvr)
```

qixiang (approved):
minor comment, lgtm

freshmaker/handlers/botas/botas_shipped_advisory.py
ronnyhlim: I'm not sure why I was so insistent on using a list comprehension :sweat_smile: , but I changed it to a normal loop as suggested (it's definitely cleaner than using nested funcs)

PR #162, opened 2022-03-24
Fix the issue to create multiple composes

chuazhan:
The original idea is to create several small composes
so that they can be reused in different images, but as
metioned in RHELWF-5880, "Creating multiple ODCS composes
risks ODCS pulling in module builds that we don't want
through module dependency resolution.". So the fix is
just replace the compose old version module with new one.

JIRA: RHELWF-5880

### DIFF
--- a/freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
+++ b/freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
@@ -227,46 +227,46 @@ class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
 
         return lb.get_images_by_nvrs(images)
 
-    def _outdated_composes(self, original_odcs_compose_ids, module_name_stream_set):
+    def _reused_composes(self, original_odcs_compose_ids, module_name_stream_set):
         """
-        Generate outdated composes.
+        Generate reused composes.
 
         :param original_odcs_compose_ids list: Original compose ids.
         :param module_name_stream_set set: Module name stream set from an advisory.
-        :return: a set of outdated composes.
+        :return: a set of reused composes.
         :rtype: set
         """
-        outdated_composes = set()
+        reused_composes = set()
         for compose_id in original_odcs_compose_ids:
             compose = self.odcs.get_compose(compose_id)
             source_type = compose.get("source_type")
             if source_type != PungiSourceType.MODULE:
-                outdated_composes.add(compose_id)
+                reused_composes.add(compose_id)
                 continue
 
             name_stream_set = {f"{n}:{s}" for n, s, v, c in _compose_sources(compose)}
 
             if name_stream_set.isdisjoint(module_name_stream_set):
-                outdated_composes.add(compose_id)
+                reused_composes.add(compose_id)
 
-        return outdated_composes
+        return reused_composes
 
-    def _missing_composes(
+    def _updated_compose_source(
         self,
         original_odcs_compose_ids,
         module_name_stream_set,
         module_name_stream_version_set,
     ):
         """
-        Generate missing composes.
+        Generate updated compose source.
 
         :param original_odcs_compose_ids list: Original compose ids.
         :param module_name_stream_set set: Module name stream set from an advisory.
         :param module_name_stream_version_set set: Module name stream version set from an advisory.
-        :return: a set of missing composes.
+        :return: a string of updated compose source.
         :rtype: set
         """
-        missing_composes = set()
+        updated_composes = set()
         for compose_id in original_odcs_compose_ids:
             compose = self.odcs.get_compose(compose_id)
             source_type = compose.get("source_type")
@@ -280,15 +280,15 @@ class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
                 }
 
                 if not name_stream_set.isdisjoint(module_name_stream_set):
-                    missing_composes.update(
+                    updated_composes.update(
                         mapping[name_stream]
                         for name_stream in name_stream_set.difference(
                             module_name_stream_set
                         )
                     )
-                missing_composes.update(module_name_stream_version_set)
+                updated_composes.update(module_name_stream_version_set)
 
-        return missing_composes
+        return " ".join(sorted(updated_composes))
 
     def _record_builds(self, images, image_modules_mapping):
         """
@@ -345,7 +345,7 @@ class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
                     module_name_stream_set.add(f"{name}:{stream}")
                     module_name_stream_version_set.add(f"{name}:{stream}:{version}")
                 original_odcs_compose_ids = image["original_odcs_compose_ids"]
-                outdated_composes = self._outdated_composes(
+                reused_composes = self._reused_composes(
                     original_odcs_compose_ids, module_name_stream_set
                 )
 
@@ -356,7 +356,7 @@ class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
                         "target": image["target"],
                         "branch": image["git_branch"],
                         "arches": image["arches"],
-                        "renewed_odcs_compose_ids": list(outdated_composes),
+                        "renewed_odcs_compose_ids": list(reused_composes),
                         "flatpak": image.get("flatpak", False),
                         "isolated": image.get("isolated", True),
                         "original_parent": None,
@@ -364,13 +364,13 @@ class RebuildFlatpakApplicationOnModuleReady(ContainerBuildHandler):
                 )
                 db.session.commit()
 
-                missing_composes = self._missing_composes(
+                compose_source = self._updated_compose_source(
                     original_odcs_compose_ids,
                     module_name_stream_set,
                     module_name_stream_version_set,
                 )
                 arches = sorted(image["arches"].split())
-                for compose_source in missing_composes:
+                if compose_source:
                     if compose_source in odcs_cache:
                         db_compose = odcs_cache[compose_source]
                     else:
--- a/tests/handlers/koji/test_rebuild_flatpak_application_on_module_ready.py
+++ b/tests/handlers/koji/test_rebuild_flatpak_application_on_module_ready.py
@@ -236,16 +236,16 @@ class TestFlatpakModuleAdvisoryReadyEvent(helpers.ModelsTestCase):
         original_odcs_compose_ids = ["985716"]
         module_name_stream_set = set(["nodejs:14"])
         module_name_stream_version_set = set(["nodejs:14:8040020211213111158"])
-        outdated_composes = self.handler._outdated_composes(
+        reused_composes = self.handler._reused_composes(
             original_odcs_compose_ids, module_name_stream_set
         )
-        missing_composes = self.handler._missing_composes(
+        updated_compose_source = self.handler._updated_compose_source(
             original_odcs_compose_ids,
             module_name_stream_set,
             module_name_stream_version_set,
         )
-        self.assertEqual(outdated_composes, {"985716"})
-        self.assertEqual(missing_composes, set())
+        self.assertEqual(reused_composes, {"985716"})
+        self.assertEqual(updated_compose_source, "")
 
     @patch("freshmaker.odcsclient.create_odcs_client")
     def test_prepare_data_for_compose_all_sources_not_in_adv(self, create_odcs_client):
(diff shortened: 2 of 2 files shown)


freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -280,15 +280,15 @@ def _missing_composes(
...
                         mapping[name_stream]
                         for name_stream in name_stream_set.difference(
                             module_name_stream_set
                         )
                     )
-                missing_composes.update(module_name_stream_version_set)
+                updated_composes.update(module_name_stream_version_set)
 
-        return missing_composes
+        return " ".join(updated_composes)
```
hluk, line 291: As a generic function interface, it would be more useful to return the list and let caller handle the concatenation if needed.

```suggestion
        return updated_composes
```

hluk: Ah, I see this is meant to return source parameter for ODCS. In that case returning the string makes sense but it would be better to mention ODCS in docstring and function name.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
```diff
@@ -364,13 +364,13 @@ def _record_builds(self, images, image_modules_mapping):
...
+                compose_source = self._updated_composes(
                     original_odcs_compose_ids,
                     module_name_stream_set,
                     module_name_stream_version_set,
                 )
                 arches = sorted(image["arches"].split())
-                for compose_source in missing_composes:
+                if compose_source:
                     if compose_source in odcs_cache:
                         db_compose = odcs_cache[compose_source]
```
hluk, line 375: Doesn't the `compose_source` need to be sorted before concatenation so it can be used as cache key consistently?

qixiang:
Could you explain more on the issue and why single compose makes differences? 

In the mentioned issue RHELWF-4651 I see a scratch build which is a rebuild for some image, but I don't see anything about original build, so I can't tell what's wrong. And in description of RHELWF-5880, there are duplicated modules in composes, which is definitely wrong, but it's not something related to generating a single or multiple composes.

For a case like:

Original image build has 2 composes:
```
compose 1: ["foo:rhel8:123:123", "bar:rhel8:123:123"]
compose 2: ["qux:rhel8:123:123"]
```
A module advisory with attached modules:
```
["bar:rhel8:333:333", "baz:rhel8:444:444"]
```
The expected rebuild should has compose(s) enabled with the following modules:
```
"foo:rhel8:123:123", "qux:rhel8:123:123", "bar:rhel8:333:333", "baz:rhel8:444:444"
```
No matter they're in a single compose or in multiple composes (but without duplicate).

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Yeah. I will update the docstring and function name.

freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py
chuazhan: Good point. I was tweaking the test codes. Will fix it in the method.

chuazhan:
> Could you explain more on the issue and why single compose makes differences?
> 


> In the mentioned issue RHELWF-4651 I see a scratch build which is a rebuild for some image, but I don't see anything about original build, so I can't tell what's wrong. And in description of RHELWF-5880, there are duplicated modules in composes, which is definitely wrong, but it's not something related to generating a single or multiple composes.
> 
> For a case like:
> 
> Original image build has 2 composes:
> 
> ```
> compose 1: ["foo:rhel8:123:123", "bar:rhel8:123:123"]
> compose 2: ["qux:rhel8:123:123"]
> ```
> 
> A module advisory with attached modules:
> 
> ```
> ["bar:rhel8:333:333", "baz:rhel8:444:444"]
> ```
> 
> The expected rebuild should has compose(s) enabled with the following modules:
> 
> ```
> "foo:rhel8:123:123", "qux:rhel8:123:123", "bar:rhel8:333:333", "baz:rhel8:444:444"
> ```
> 
> No matter they're in a single compose or in multiple composes (but without duplication).

chuazhan:
> Could you explain more on the issue and why single compose makes differences?
> 


> In the mentioned issue RHELWF-4651 I see a scratch build which is a rebuild for some image, but I don't see anything about original build, so I can't tell what's wrong. And in description of RHELWF-5880, there are duplicated modules in composes, which is definitely wrong, but it's not something related to generating a single or multiple composes.
> 
> For a case like:
> 
> Original image build has 2 composes:
> 
> ```
> compose 1: ["foo:rhel8:123:123", "bar:rhel8:123:123"]
> compose 2: ["qux:rhel8:123:123"]
> ```
> 
> A module advisory with attached modules:
> 
> ```
> ["bar:rhel8:333:333", "baz:rhel8:444:444"]
> ```
> 
> The expected rebuild should has compose(s) enabled with the following modules:
> 
> ```
> "foo:rhel8:123:123", "qux:rhel8:123:123", "bar:rhel8:333:333", "baz:rhel8:444:444"
> ```
> 
> No matter they're in a single compose or in multiple composes (but without duplication).

chuazhan:
> Could you explain more on the issue and why single compose makes differences?
> 
> In the mentioned issue RHELWF-4651 I see a scratch build which is a rebuild for some image, but I don't see anything about original build, so I can't tell what's wrong. And in description of RHELWF-5880, there are duplicated modules in composes, which is definitely wrong, but it's not something related to generating a single or multiple composes.
> 
> For a case like:
> 
> Original image build has 2 composes:
> 
> ```
> compose 1: ["foo:rhel8:123:123", "bar:rhel8:123:123"]
> compose 2: ["qux:rhel8:123:123"]
> ```
> 
> A module advisory with attached modules:
> 
> ```
> ["bar:rhel8:333:333", "baz:rhel8:444:444"]
> ```
> 
> The expected rebuild should has compose(s) enabled with the following modules:
> 
> ```
> "foo:rhel8:123:123", "qux:rhel8:123:123", "bar:rhel8:333:333", "baz:rhel8:444:444"
> ```
> 
> No matter they're in a single compose or in multiple composes (but without duplication).

The original build is {'libreoffice:flatpak:8050020210830125255', 'python36:3.6:8050020210811103506', 'flatpak-runtime:el8:8050020210831114120'}
And this patch just replaces the 'libreoffice:flatpak:8050020210830125255' with the latest version from the module build 'libreoffice:flatpak:8050020220215203934'.
This is the new build: https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=44083666
Although it is failed, the build version is updated as expected. The difference with the previous implementation is to merge composes into one compose source, both have removed duplicate items.

qixiang:
> The original build is {'libreoffice:flatpak:8050020210830125255', 'python36:3.6:8050020210811103506', 'flatpak-runtime:el8:8050020210831114120'} And this patch just replaces the 'libreoffice:flatpak:8050020210830125255' with the latest version from the module build 'libreoffice:flatpak:8050020220215203934'. This is the new build:  brew task 44083666, Although it is failed, the build version is updated as expected. The difference with the previous implementation is to merge composes into one compose source, both have removed duplicate items.

1. Please update your commit message which is misleading, the issue is not about creating multiple composes or single composes, it's about old corresponding module was not replaced with new module.
2. Please don't include internal urls in public sites, you can use brew task xxx, build id xxx and etc.
3. Some function/variable names are misleading as well, please add some docstring to explain more. For example, you have 
```
                outdated_composes = self._outdated_composes(
                    original_odcs_compose_ids, module_name_stream_set
                )
                ...
                        "renewed_odcs_compose_ids": list(outdated_composes)
```
where the name indicates the composes are outdated, but actually they're not and you're reusing them in new build, this is misleading.
4. In `_updated_compose_source`, you replaced the old modules when there are updates in advisory, but just dropped modules which are newer than modules in advisory, this is wrong.

chuazhan:
> 4. In `_updated_compose_source`, you replaced the old modules when there are updates in advisory, but just dropped modules which are newer than modules in advisory, this is wrong.

This implementation is based on the assumption that the modules in the advisory are always newer than the original build.
Is possible it is not true? If so, do we have any way to compare the module version?

qixiang:
> > 4. In `_updated_compose_source`, you replaced the old modules when there are updates in advisory, but just dropped modules which are newer than modules in advisory, this is wrong.
> 
> This implementation is based on the assumption that the modules in the advisory are always newer than the original build. Is possible it is not true? If so, do we have any way to compare the module version?

well, I misread it, it doesn't compare version at all, just replaced modules with same NAME:STREAM. I'm not sure whether this can happen, it's unlike the rpm case, which image rebuilds are triggered by rpm shipped event, in this case, rebuilds are triggered before modules are shipped (though to hidden repos), so if there are multiple module advisories exist (for same module), in theory this can happen.

chuazhan:
> > > 4. In `_updated_compose_source`, you replaced the old modules when there are updates in advisory, but just dropped modules which are newer than modules in advisory, this is wrong.
> > 
> > 
> > This implementation is based on the assumption that the modules in the advisory are always newer than the original build. Is possible it is not true? If so, do we have any way to compare the module version?
> 
> well, I misread it, it doesn't compare version at all, just replaced modules with same NAME:STREAM. I'm not sure whether this can happen, it's unlike the rpm case, which image rebuilds are triggered by rpm shipped event, in this case, rebuilds are triggered before modules are shipped (though to hidden repos), so if there are multiple module advisories exist (for same module), in theory this can happen. But I don't have strong opinion to support this corner case at this moment.

Thanks, @qixiang for the information sharing. Actually, I am thinking even the module advisory is an old one, that is the one that users want to ship with ET(I think that will not happen :D).

chuazhan:
Updated

chuazhan:
@qixiang @hluk Can you help to review again? Thanks.

qixiang:
> The original idea is to create several small composes
> so that they can be reused in different images, but as
> metioned in RHELWF-5880, "Creating multiple ODCS composes
> risks ODCS pulling in module builds that we don't want
> through module dependency resolution.". So this fix is
> to replace old module build in compose source instead of
> creating serveral small composes.

As mentioned above, could you update the commit message which is misleading? the original issue is not about creating multiple composes or single compose, the original issue is caused by both old modules and new modules are enabled in rebuild task.

I see `get_cve_affected_rpm_nvrs` is still being used by https://github.com/redhat-exd-rebuilds/freshmaker/blob/05987c7e68ca349bdcf337ca79e333041a308c35/freshmaker/handlers/koji/rebuild_flatpak_application_on_module_ready.py#L215-L217 I believe this is wrong, please fix that in a separate commit.

qixiang:
> > The original idea is to create several small composes
> > so that they can be reused in different images, but as
> > metioned in RHELWF-5880, "Creating multiple ODCS composes
> > risks ODCS pulling in module builds that we don't want
> > through module dependency resolution.". So this fix is
> > to replace old module build in compose source instead of
> > creating serveral small composes.
> 
> As mentioned above, could you update the commit message which is misleading? the original issue is not about creating multiple composes or single compose, the original issue is caused by both old modules and new modules are enabled in rebuild task.

As we talked on chat, the issue is probably that when you submit an ODCS module compose, ODCS will try to resolve the dependency of the requested modules and add the missing modules in generated compose. For example, when you request a module compose with source `libreoffice:flatpak:...`, ODCS will automatically add `python36:3.6... flatpak-runtime:...` after dependency resolving, and the `python36` `flatpak-runtime` are latest published modules, so if we have `flatpak-runtime` or `python36` in advisory and only want the new ones, this will cause some risks of pulling unexpected packages.

This is should be documented in comments to avoid changing of breaking the mono compose to small ones.

chuazhan:
> As we talked on chat, the issue is probably that when you submit an ODCS module compose, ODCS will try to resolve the dependency of the requested modules and add the missing modules in generated compose. For example, when you request a module compose with source `libreoffice:flatpak:...`, ODCS will automatically add `python36:3.6... flatpak-runtime:...` after dependency resolving, and the `python36` `flatpak-runtime` are latest published modules, so if we have `flatpak-runtime` or `python36` in advisory and only want the new ones, this will cause some risks of pulling unexpected packages.
> 
> This is should be documented in comments to avoid changing of breaking the mono compose to small ones.

Ack. Update the commit message, please help to review again, thanks!

qixiang:
:+1:

hluk (approved):

PR #163, opened 2022-03-24
Remove RHC bob trigger from Freshmaker

ronnyhlim:
RHC Bob will be decommissioned and the trigger in Freshmaker needs to be removed too.


apaplaus (approved):
Seems fine, but I would wait for qwan to check it as well.

qixiang:
LGTM

qixiang (approved):

PR #164, opened 2022-03-28
Find images only in repositories have auto rebuild tags enabled

qixiang:
While querying LightBlue to find out impacted images, the query should
only be performed against repositories which have auto rebuild tags
enabled, this will reduce the time required for the image query.

And change the '$or' operators in query to '$in' which is more efficient.

JIRA: CWFENGINE-644

Signed-off-by: Qixiang Wan <qwan@redhat.com>


mulaievaRH (approved):
LGTM

PR #165, opened 2022-04-01
Add new method to get build rpm nvrs

chuazhan:
As the flatpak application advisory is not
restricted to CVE related, we need to use new
API to get build rpm nvrs.


freshmaker/errata.py
```diff
@@ -514,3 +514,23 @@ def get_attached_build_nvrs(self, errata_id):
             for build in builds
             for nvr in build.keys()
         }
+
+    def get_build_rpm_nvrs(self, errata_id):
```
qixiang, line 518: Any reason of creating this new API? there is already `get_binary_rpm_nvrs` exist.

chuazhan: Ah. I haven't noticed it. I was trying to add a new api like get_cve_affected_rpm_nvrs. Thanks for pointing this out. Fixed.

qixiang:
:+1:

hluk (approved):

PR #166, opened 2022-04-06
Freshmaker configuration refactoring

ElenaKarolinaSemanova:
Refactor and simplify freshmaker configuration files and set up
Jira: CWFHEALTH-1159

### DIFF
--- a/freshmaker/config.py
+++ b/freshmaker/config.py
@@ -32,6 +32,8 @@ import threading
 from os import sys  # type: ignore
 from freshmaker import logger
 
+PROD_CONFIG_FILE = '/etc/freshmaker/config.py'
+
 
 def any_(*rules):
     """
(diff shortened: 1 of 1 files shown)


ElenaKarolinaSemanova:
I refactored `freshmaker/config.py` and `conf/config.py`, but I didn't merge them into one file, because as I understand it, `freshmaker/config.py` is the module where freshmaker config is loaded from config file, and both `conf/config.py` and `/etc/freshmaker/config.py` are config files themeselves.

freshmaker/config.py
```diff
@@ -57,65 +57,88 @@ def all_(*rules):
...
     conf = Config(config_section_obj)
     app.config.from_object(config_section_obj)
     return conf
 
 
+def init_config(app):
+    """
+    Configure Freshmaker
+    """
+    prod_config_file = '/etc/freshmaker/config.py'
```
mulaievaRH, line 131: What do you think if we would create a global var PROD_CONFIG_FILE = ''/etc/freshmaker/config.py'' and after it will check if  'FRESHMAKER_CONFIG_FILE'  exists as env var, config_file = FRESHMAKER_CONFIG_FILE. Otherwise config_file = PROD_CONFIG_FILE. This way we can delete def get_config_file(prod_config_file).

freshmaker/config.py
```diff
@@ -463,6 +484,10 @@ def set_item(self, key, value):
         # Create the empty self._key attribute, so we can assign to it.
         setattr(self, "_" + key, None)
 
+        self.create_property(key)
+        self.create_attribute(key, value)
```
mulaievaRH, line 488: Sorry, I didn't get, what was the point to add this?

freshmaker/config.py
```diff
@@ -502,7 +525,7 @@ def set_item(self, key, value):
 
     def _setifok_log_backend(self, s):
         if s is None:
-            self._log_backend = "console"
+            self._log_backend = "journal"
```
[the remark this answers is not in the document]
mulaievaRH: Yes, I checked and we have `journal` everywhere. I would say it's safe to change it here.
@qixiang , wdyt?

freshmaker/config.py
```diff
@@ -57,65 +57,88 @@ def all_(*rules):
...
+
+    if not config_module:
+        config_module = get_config_module_from_file(config_file)
+
+    return get_freshmaker_config(app, config_section, config_module)
+
+
 class Config(object):
     """Class representing the freshmaker configuration."""
     _defaults = {
```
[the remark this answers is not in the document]
mulaievaRH: Yes, they should stay

conf/config.py
```diff
@@ -236,9 +236,9 @@ class DevConfiguration(BaseConfiguration):
 
 
 class TestConfiguration(BaseConfiguration):
+    DEBUG = True
```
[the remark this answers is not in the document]
ElenaKarolinaSemanova: Honestly, I looked back and I think there wasn't a reason for that so I can revert it.

freshmaker/config.py
ElenaKarolinaSemanova: I split the code to sub methods for better decomposition @mulaievaRH

freshmaker/config.py
ElenaKarolinaSemanova: I like this suggestion, thank you

mulaievaRH:
LGTM

freshmaker/config.py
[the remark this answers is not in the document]
qixiang: I think keeping 'console' as default makes more sense, when this is not specified, it's better to logging to 'console' rather than to 'journal'.

freshmaker/config.py
```diff
@@ -478,22 +503,20 @@ def set_item(self, key, value):
         delx = lambda self: delattr(self, "_" + key)
         setattr(Config, key, property(getx, setx, delx))
```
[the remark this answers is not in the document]
qixiang: This makes some slight differences, for example, with `self`, you will run into an error when you call `config_instance._<key>`, while `config_instance.<key>` works in both cases. I don't know the exact reason why `Config` is used here, but it'd be safe to keep as it as, and I think use class makes more sense, at least it's documentedhttps://docs.python.org/3/reference/datamodel.html#implementing-descriptors.
```

File: freshmaker/config.py (hunk 2 above)
Comment (qixiang) [reply freshmaker/config.py:481]: IMO, I think it would be better to not split it, because `create_property` and `create_attribute` should always be called together and maybe also in order, when you split it to two functions, there is no mechanism to make sure they will be called in the right way (this will bring some confusions to future devs).

File: freshmaker/config.py [hunk 7]
```diff
@@ -125,7 +141,7 @@ class Config(object):
             'desc': 'Debug mode'},
         'log_backend': {
             'type': str,
-            'default': None,
+            'default': 'journal',
```
ElenaKarolinaSemanova, line 144: What do you think about this @qixiang , I guess it should stay `None` rather than `journal`?

qixiang: right, keep it as `None` is better.

qixiang (approved):

PR #167, opened 2022-04-12
Fix returning invalid value from module handler

hluk:
The return value should be list of unhandled events.

JIRA: RHELWF-5907


qixiang (approved):
:+1:

PR #177, opened 2022-08-09
pyxis_gql: add new API get_repository_by_registry_path

qixiang:
New API get_repository_by_registry_path to get container repository by
registry path.

### DIFF
--- a/freshmaker/pyxis_gql.py
+++ b/freshmaker/pyxis_gql.py
@@ -118,3 +118,34 @@ class PyxisGQL:
             page_num += 1
 
         return repositories
+
+    def get_repository_by_registry_path(self, repository, registry):
+        """Get image repository by registry path
+
+        :param str repository: repository name
+        :param str registry: registry name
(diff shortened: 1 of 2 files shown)


ronnyhlim (approved):

qixiang:
> LGTM! +1 general question though, https://github.com/redhat-exd-rebuilds/freshmaker/pull/177/files#diff-79784fd03ea58eb32fde6d829709747cbf637b3ff3b022fe272e117e27644e7bR131-R132 looks like it's getting the repository, but one of the arguments passed in is already the repository. What is the reason for that?

This API accepts `registry` and `repository`, then returns the repository info we specified in the query, includes `release_categories`, `auto_rebuild_tags`, `registry` and `repository`, `registry` and `repository` in the result is actually not so necessary, but it's good to keep it consistent with the `find_repositories` API, I'll probably add a class to represent the container repository and then return an instance instead of a dict, which would be easier.

PR #181, opened 2022-10-05
Implement find_images_by_names with Pyxis GraphQL

ElenaKarolinaSemanova:
Implement find_images_by_names with Pyxis GraphQL with same projection as LightBlue's result, Jira: CWFHEALTH-1251

### DIFF
--- a/freshmaker/pyxis_gql.py
+++ b/freshmaker/pyxis_gql.py
@@ -341,3 +341,57 @@ class PyxisGQL:
             page_num += 1
 
         return images
+
+    def find_images_by_names(self, names):
+        """
+        Find all the images for a specific list of names.
+        :param names list: list of names we want to find images for.
+        :return: list of container images matching the requested names.
+        :rtype: list of ContainerImages
+        """
+        images = []
+        query_filter = {"and": []}
+        query_filter["and"].append({"brew": {"package": {"in": names}}})
+        # Only query for published images
+        query_filter["and"].append(
+            {"repositories_elemMatch": {"published": {"eq": True}}}
+        )
+
+        ds = self.dsl_schema
+        page_num = 0
+        page_size = 50
+
+        while True:
+            query_dsl = ds.Query.find_images(
+                page=page_num,
+                page_size=page_size,
+                filter=query_filter,
+            ).select(
+                ds.ContainerImagePaginatedResponse.error.select(
+                    ds.ResponseError.status,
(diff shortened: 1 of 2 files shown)


ElenaKarolinaSemanova:
@qixiang @mulaieva Please review

freshmaker/pyxis_gql.py
```diff
@@ -90,7 +90,10 @@ def _get_image_projection(self, include_rpms=True):
         projection = [
             ds.ContainerImage.architecture,
             ds.ContainerImage.brew.select(
+                ds.Brew.package,
```
qixiang, line 93: I realized that we have some fields included implicitly in LightBlue result, so when I say return same data as the LightBlue result, it's actually inaccurate, some of these fields are not used by any code at all (e.g. `ds.Brew.completion_date`), so could you remove these new fields in this PR? I'll check the existing code later to confirm whether these fields are used.

freshmaker/pyxis_gql.py
```diff
@@ -90,7 +90,10 @@ def _get_image_projection(self, include_rpms=True):
...
+                ds.Brew.nvra,
```
qixiang, line 94:
```suggestion
```

freshmaker/pyxis_gql.py
```diff
@@ -90,7 +90,10 @@ def _get_image_projection(self, include_rpms=True):
...
                 ds.Brew.build,
+                ds.Brew.completion_date,
```
qixiang, line 96:
```suggestion
```

tests/test_pyxis_gql.py
```diff
@@ -319,3 +319,105 @@ def test_pyxis_graphql_find_images_by_nvr():
...
+            "total": 2,
+        }
+    }
+
+    pyxis_gql = PyxisGQL(
+        url="graphql.pyxis.local", cert=("/path/to/crt", "/path/to/key")
+    )
+    flexmock(Client).should_receive("execute").and_return(copy.deepcopy(result))
+
+    images = pyxis_gql.find_images_by_names("foo")
```
qixiang, line 422: In the mocked data, the image name is `foobar-container` instead of `foo`.

ElenaKarolinaSemanova:
@qixiang I resolved your suggestions, you can take a final look

qixiang (approved):
LGTM

PR #182, opened 2022-10-18
Exclude hidden content_set

chuazhan:
As described in RHELWF-5866, making compose failed when including hidden content_set, probably the reason is ODCS thinks it is unpublished content set.

JIRA: RHELWF-5866

### DIFF
--- a/freshmaker/config.py
+++ b/freshmaker/config.py
@@ -444,6 +444,11 @@ class Config(object):
             'default': 'lightblue',
             'desc': 'An option to get image metadata from LightBlue or Pyxis GraphQL. Can be set to "pyxis_graphql" or "lightblue".'
         },
+        'exclude_content_sets_pattern': {
+            'type': str,
+            'default': '-hidden-rpms$',
+            'desc': 'Pattern for content sets which will be excluded while generating composes'
+        },
     }
 
     def __init__(self, conf_section_obj):
--- a/freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
+++ b/freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
@@ -24,6 +24,7 @@
 
 import json
 import koji
+import re
 
 from freshmaker import conf, db
 from freshmaker.events import (
@@ -318,6 +319,9 @@ class RebuildImagesOnRPMAdvisoryChange(ContainerBuildHandler):
                     missing_content_sets = set()
                     for content_set in image["content_sets"]:
                         if content_set in image["compose_sources"]:
+                            continue
+                        # Exclude hidden content set.
+                        if re.search(conf.exclude_content_sets_pattern, content_set):
                             continue
                         missing_content_sets.add(content_set)
 
--- a/tests/handlers/koji/test_rebuild_images_on_rpm_advisory_change.py
+++ b/tests/handlers/koji/test_rebuild_images_on_rpm_advisory_change.py
@@ -1195,6 +1195,47 @@ class TestRecordBatchesImages(helpers.ModelsTestCase):
         self.assertEqual(ArtifactBuildState.PLANNED.value, parent_image.state)
         self.mock_prepare_pulp_repo.assert_called()
 
+    def test_record_batches_not_generate_hidden_pulp_repos(self):
+        batches = [
+            [ContainerImage({
+                "brew": {
+                    "completion_date": "20170420T17:05:37.000-0400",
+                    "build": "rhel-server-docker-7.3-82",
+                    "package": "rhel-server-docker"
+                },
+                'parsed_data': {
+                    'layers': [
+                        'sha512:12345678980',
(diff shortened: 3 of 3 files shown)


freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
```diff
@@ -319,6 +320,9 @@ def _record_batches(self, batches, event, builds=None):
                     for content_set in image["content_sets"]:
                         if content_set in image["compose_sources"]:
                             continue
+                        # Exclude hidden content set.
+                        if re.search(".+(-hidden-).+", content_set):
```
hluk, line 324:
```suggestion
                        if content_set.endswith("-hidden-rpms"):
```

But this should probably go into Freshmaker configuration in our internal repo as a list of regexps, because it is an internal thing.

chuazhan: I am not sure, do we have an external freshmaker service to deploy?

hluk: No, but any internal and potentially sensitive data should be kept in our internal git repos.

chuazhan: Ok. updated

freshmaker/config.py
```diff
@@ -439,6 +439,11 @@ class Config(object):
             'default': '',
             'desc': 'Root url of Flatpak index service'
         },
+        'hidden_content_set_pattern': {
```
hluk, line 442: This is not pattern but suffix.

In the configuration, using a regex pattern seems like a better more flexible solution: `exclude_content_sets_pattern`

freshmaker/config.py
```diff
@@ -439,6 +439,11 @@ class Config(object):
...
+            'type': str,
+            'default': '-hidden-rpms',
+            'desc': 'Pattern for hidden content_set'
```
qixiang, line 445:
```suggestion
            'desc': 'Pattern for content sets which will be excluded while generating composes'
```

freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
```diff
@@ -319,6 +319,9 @@ def _record_batches(self, batches, event, builds=None):
                     for content_set in image["content_sets"]:
                         if content_set in image["compose_sources"]:
                             continue
+                        # Exclude hidden content set.
+                        if content_set.endswith(conf.hidden_content_set_pattern):
```
qixiang, line 323: Could you check with regex matching instead of `endswith`?

freshmaker/config.py
chuazhan: Ack. Fixing.

freshmaker/config.py
chuazhan: Ack. Fixing.

freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
chuazhan: Not sure about which is better between regex and endswith, my original code is using regex, Lukas commented better with endswith, both seems fine for me.

hluk: I commented on `content_set.endswith("-hidden-rpms")` - so previously it looked like suffix-only, but it would be better with full regex match.

hluk: Also, the config name is confusing because it contains `pattern`, but it is still simple string (and suffix) - not a regex nor glob expression. Either use `hidden_content_set_suffix` or use `hidden_content_set_pattern` but with regex matching.

chuazhan: Ah,ok, got your point finally. Thanks for your explanation. Again, which one you prefer? endswith or regex, I use regex originally because I don't know the correct format for hidden set, I think string "hidden" must be there. So use "hidden" regex, I am assuming the "-hidden-rpms" should be more accurate after Lukas review, so I changed to endswith.

chuazhan:
Codes updated, please help to review again, thanks.

freshmaker/config.py
```diff
@@ -444,6 +444,11 @@ class Config(object):
             'default': 'lightblue',
             'desc': 'An option to get image metadata from LightBlue or Pyxis GraphQL. Can be set to "pyxis_graphql" or "lightblue".'
         },
+        'exclude_content_sets_pattern': {
+            'type': str,
+            'default': '.+(-hidden-rpms)',
```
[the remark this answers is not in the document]
chuazhan: fullmatch check the entire string.

hluk (approved):

qixiang (approved):

PR #184, opened 2022-10-26
Add a new option to get image metadata from LightBlue or Pyxis

ElenaKarolinaSemanova:
Add an option in Freshmaker to switch between using LightBlue and Pyxis GraphQL; Jira: CWFHEALTH-1246


qixiang (approved):
LGTM

PR #185, opened 2022-11-09
Fix failure building python-ldap

hluk:
Fixes `/usr/bin/ld: cannot find -lldap_r` error when compiling python-ldap on Fedora 36.

The single dependency was updated manually in the generated requirements.txt file.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -159,7 +159,7 @@ python-fedora==1.1.1
     #   odcs
 python-json-logger==0.1.7
     # via -r requirements.in
-python-ldap==3.3.1
+python-ldap==3.4.3
     # via
     #   -r requirements.in
     #   pyldap


hluk:
Let me know if running `pip-compile requirements.in` makes more sense - but I guess this would potentially break things.

qixiang:
> Let me know if running `pip-compile requirements.in` makes more sense - but I guess this would potentially break things.

I'm fine with updating the requirements.txt, I tried to figure out a solution to run `pip-compile` to produce a requirements file with package versions equal to the existing environment, the purpose is to install our dependencies with yum and rpms first, then generate a requirements based on what the dependencies provided by these rpms, but I didn't find a solution :(

qixiang (approved):
:+1:

hluk:
BTW, for gating apps I started using **Poetry** to handle dependencies with version pinning and try to avoid using any system packages whenever possible. Dependabot can auto-update the poetry.lock file nicely though it often creates conflicting PRs (which just means I wait after merging a PR for dependabot to auto-resolve conflicts).

See: https://github.com/release-engineering/greenwave (mainly `pyproject.toml`, `tox.ini`, `Dockerfile`, `poetry.lock`)

qixiang:
> BTW, for gating apps I started using **Poetry** to handle dependencies with version pinning and try to avoid using any system packages whenever possible. Dependabot can auto-update the poetry.lock file nicely though it often creates conflicting PRs (which just means I wait after merging a PR for dependabot to auto-resolve conflicts).
> 
> See: https://github.com/release-engineering/greenwave (mainly `pyproject.toml`, `tox.ini`, `Dockerfile`, `poetry.lock`)

nice, I'll take a look later :)

PR #187, opened 2022-11-09
Check availability parent_build_id

mulaievaRH:
Check if parent_build_id exists and if not, FM skip it.
JIRA: CWFHEALTH-1628

### DIFF
--- a/freshmaker/lightblue.py
+++ b/freshmaker/lightblue.py
@@ -1224,6 +1224,11 @@ class LightBlue(object):
         # We need to resolve the image in here because "parent_image_builds" needs to be there
         # and it gets populated when the image gets resolved.
         child_image.resolve(self)
+
+        # Some images have `parent_image_builds` but are built in a multi-stage way and don't have
+        # `parent_build_id`, in this situation FM doesn't try to find the parent image and skip.
+        if not child_image.get("parent_build_id"):
+            return None
         # If the parent is not in `parent_brew_build` we can try to look for the parent in Brew,
         # using the field `parent_image_builds` (searching for the nvr), which should always be there.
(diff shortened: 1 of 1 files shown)


freshmaker/lightblue.py
```diff
@@ -1218,6 +1218,8 @@ def find_parent_brew_build_nvr_from_child(self, child_image):
         :rtype: str
 
         """
+        if not child_image.get("parent_build_id"):
+            return None
```
qixiang, line 1222: In case this image has not been resolved, it will return before resolving which is unexpected. So we need to do this after resolving https://github.com/redhat-exd-rebuilds/freshmaker/blob/f9bac30955792b1d8a6e0c43447a11afbe970334/freshmaker/lightblue.py#L1226

Please add a comment there to explain why we need to check the `parent_build_id`.

qixiang (changes requested):

freshmaker/lightblue.py
```diff
@@ -1224,6 +1224,11 @@ def find_parent_brew_build_nvr_from_child(self, child_image):
         # We need to resolve the image in here because "parent_image_builds" needs to be there
         # and it gets populated when the image gets resolved.
         child_image.resolve(self)
+
+        # Some images are built in a multi-stage way and they don't have a parent
+        # In this situation FM doesn't try to find the parent image and skip
```
qixiang, line 1229:
```suggestion
        # Some images have `parent_image_builds` but are built in a multi-stage way and don't have
        # `parent_build_id`, in this situation FM doesn't try to find the parent image and skip.
```

mulaievaRH:
@qixiang , changes accepted. Could it be merged now?

qixiang (approved):
:+1:

PR #188, opened 2022-11-09
WIP: Update image_verifier.py to use APIs in Pyxis GraphQL module

ElenaKarolinaSemanova:
Update image_verifier.py to use APIs in Pyxis GraphQL module, Jira: CWFHEALTH-1255


qixiang:
Hi @ElenaKarolinaSemanova While reviewing your proposal of supporting both LightBlue and Pyxis as the metadata source, I having been thinking on just drop the LightBlue support for the image_verifier module, the reason of supporting both the two metadata source is some handlers are complicated in its implementation and we want to run dev with Pyxis to test it carefully, and make it possible to switch to LightBlue when we see critical bugs on prod with Pyixs. However, I think the image_verifier is much simpler and we can perform the tests easily, so it probably doesn't worth the effort to supporting both the metadata source, what do you think?

ElenaKarolinaSemanova:
@qixiang ptal

qixiang:
> @qixiang ptal

Could you test it locally to verify these APIs work as expected? A simple way to test it locally could be something like:
```
from freshmaker.image_verifier import ImageVerifier
image_verifier = ImageVerifier()
image_verifier.verify_repository("some/repo")
image_verifier.verify_image("some-image-nvr")
```
And please squash the commits to remove the unnecessary commit which the changes have been dropped.

freshmaker/image_verifier.py
```diff
@@ -180,15 +153,17 @@ def verify_repository(self, repo_name):
 
         data = {
             "repository": {"auto_rebuild_tags": repo["auto_rebuild_tags"]},
-            "images": {}
+            "images": {},
         }
 
-        images = self.lb.find_images_with_included_rpms(
-            [], [], {repo["repository"]: repo}, include_rpm_manifest=False)
+        images = self.pgql.find_images_by_installed_rpms(
```
ElenaKarolinaSemanova, line 159: While testing locally, I noticed this API is implemented in a way that it needs to have the argument `rpm_names` with some value, otherwise this will always be false `query_filter["and"].append({"rpm_manifest": {"rpms": {"name": {"in": rpm_names}}}})`. However, I also noticed that lb API `find_images_with_included_rpms` will be working fine, even if `rpm_nvrs` is None for example, because the `if` statement (https://github.com/redhat-exd-rebuilds/freshmaker/blob/master/freshmaker/lightblue.py#L996) will not pass so the api will just ignore that `rpm_nvrs` is None and more further. So I'm not sure if this is intended behaviour of the PyxisGQL api? @qixiang

ElenaKarolinaSemanova:
@qixiang  ptal

freshmaker/image_verifier.py
```diff
@@ -147,21 +110,31 @@ def verify_image(self, image_nvr):
         :rtype: dict
         :return: Dict with image NVR as key and list of content_sets as values.
         """
-        repo = self._get_repository_from_image(image_nvr)
-        self._verify_repository_data(repo)
+        images = self.pgql.find_images_by_nvrs([image_nvr], include_rpms=False)
```
qixiang, line 113: We have `find_images_by_nvr` in pyxis_gql which is better for the purpose.

qixiang (changes requested):
I haven't reviewed the code in very detail, only added some comments to some visible issues. Please rebase the PR on top of the latest code, so we can pull the PR to local more easily, and when you finish the refactoring of this module, please try to use the APIs locally to test these APIs can work.

freshmaker/image_verifier.py
```diff
@@ -22,121 +22,84 @@
...
-        self.lb = lb if lb else LightBlue(
-            server_url=conf.lightblue_server_url,
-            cert=conf.lightblue_certificate,
-            private_key=conf.lightblue_private_key)
+        self.pgql = (
+            pgql
+            if pgql
+            else PyxisGQL(
+                url="https://graphql.pyxis.engineering.redhat.com/graphql/",
+                cert=("/path/to/crt", "/path/to/key",),
```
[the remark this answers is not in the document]
ElenaKarolinaSemanova: Should I put actual paths there instead? @qixiang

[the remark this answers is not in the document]
qixiang: This should be configurable like the LightBlue cert: https://github.com/redhat-exd-rebuilds/freshmaker/blob/b035d1ca862f0571a0d1f4780fca15d15f9e347f/freshmaker/config.py#L278-L285

ElenaKarolinaSemanova:
@qixiang Please take a look

tests/test_image_verifier.py
```diff
@@ -21,179 +21,314 @@
...
+                }
+            )
+        ]
         self.assertRaisesRegex(
-            ValueError, r'.*found in multiple repositories in Lightblue.',
-            self.verifier.verify_image, "foo/bar")
+            ValueError,
+            r".*found in multiple repositories in Pyxis.",
+            self.verifier.verify_image,
+            "foo-1-1",
```
ElenaKarolinaSemanova, line 327: @qixiang Previously there was a "foo/bar" argument, but I think the argument "foo-1-1", which is the image's brew build, makes more sense

freshmaker/image_verifier.py
```diff
@@ -22,120 +22,106 @@
...
         if len(repos) != 1:
-            raise ValueError("Multiple records found in Lightblue for repository %s." % repo_name)
+            raise ValueError(
+                "Multiple records found in Pyxis for image repository %s." % repo_name
+            )
 
         return repos[0]
 
-    def _get_repository_from_image(self, nvr):
+    def _get_repository_from_image(self, image):
```
ElenaKarolinaSemanova, line 104: @qixiang My workaround for getting repository from image

qixiang:
hi @ElenaKarolinaSemanova I made some changes to the pyxis_gql module in #194, could you rebase this PR and update the push target to `pyxis` branch once #194 is merged?

freshmaker/image_verifier.py
```diff
@@ -22,120 +22,119 @@
...
         Raises ValueError in case of error.
         """
+        if "brew" not in image:
+            raise ValueError("'brew' field not set in image")
+
+        if "build" not in image["brew"]:
+            raise ValueError("'brew.build' field not set in image")
+
         if not image["content_sets"]:
             raise ValueError(
```
ElenaKarolinaSemanova, line 91: While testing `verify_image` and `verify_repository`, sometimes it was failing on this exception because an image had empty content_sets, but I suppose that's alright because the methods were throwing exceptions when they were supposed to?

freshmaker/image_verifier.py
```diff
@@ -22,120 +22,119 @@
...
+            for exc in conf.unpublished_exceptions:
+                if image_repo["repository"] == exc["repository"]:
+                    break
+            else:
+                raise ValueError(
+                    "Only images with published repositories or unpublished exceptions can be rebuilt, but this image's repository is not published."
+                )
+
+        repos = self.pgql.find_repositories_by_repository_name(image_repo["repository"])
         return repos[0]
```
ElenaKarolinaSemanova, line 138: @qixiang Is the logic to return the first repo alright?

tests/test_image_verifier.py
```diff
@@ -21,179 +21,317 @@
...
+            ValueError,
+            r"Cannot get repository.*",
+            self.verifier.verify_repository,
+            "foo/bar",
+        )
 
     def test_get_verify_repository_multiple_repos(self):
-        self.lb.find_container_repositories.return_value = ["foo", "bar"]
+        self.pgql.find_repositories_by_repository_name.return_value = [
+            ContainerRepository(
```
ElenaKarolinaSemanova, line 48: @qixiang Is it neessary for repositories to be mocked also with `content_sets` as here for example https://github.com/redhat-exd-rebuilds/freshmaker/blob/pyxis/tests/test_lightblue.py#L572?

freshmaker/image_verifier.py
```diff
@@ -183,21 +179,22 @@ def verify_repository(self, repo_name):
             "images": {}
         }
 
-        images = self.lb.find_images_with_included_rpms(
-            [], [], {repo["repository"]: repo}, include_rpm_manifest=False)
+        images = self.pgql.find_images_by_repository(repo)
```
ElenaKarolinaSemanova, line 182: @qixiang Images returned by this API have different format than images returned by `find_images_with_included_rpms` in LighBlue. In `find_images_with_included_rpms`, the images are post-processed a few times beginning here https://github.com/redhat-exd-rebuilds/freshmaker/blob/pyxis/freshmaker/lightblue.py#L1011 .  Should I implement such postprocessing to pyxis as well, since the output in pyxis isn't really similar without this postprocessing that happens in LightBlue?

freshmaker/pyxis_gql.py
```diff
@@ -496,3 +548,58 @@ def find_images_by_names(self, names):
...
+        :param string repository: repository name to filter by
+        :return: List of image data
+        :rtype: list
+        """
+        images = []
+
+        query_filter = {}
+        query_filter["and"] = []
+
+        query_filter["and"].append({"repositories": {"repository": {"eq": repo["repository"]}}})
```
qixiang, line 564: `repo` is a string by design in your doc string, but here it's used as a dict. I agree the api should be designed to accept a string, use dict as a parameter is unnecessary and making the API hard to use.

freshmaker/pyxis_gql.py
```diff
@@ -496,3 +548,58 @@ def find_images_by_names(self, names):
...
+        query_filter["and"].append({"repositories": {"published": {"eq": True}}})
+        query_filter["and"].append({"repositories": {"tags": {"name": {"in": repo["auto_rebuild_tags"]}}}})
+
+        ds = self.dsl_schema
+        page_num = 0
+        page_size = 50
```
qixiang, line 570: `PYXIS_PAGE_SIZE` should be used instead.

freshmaker/pyxis_gql.py
```diff
@@ -374,7 +426,7 @@ def find_images_by_installed_rpms(
 
         :param list rpm_names: List of rpm names
         :param list content_sets: List of content sets
-        :param list repositories: List of repository paths
+        :param dict repositories: Dictionary of repository names as keys and repositories as values
```
qixiang, line 429: Please don't change this, this API is called by other code as well, or probably you were meant to update the docstring of `find_images_by_repository`?

freshmaker/image_verifier.py
```diff
@@ -22,120 +22,119 @@
...
-        :param LightBlue lb: Lightblue instance to use to verify images.
-            When None, new default Lightblue class is created.
+        :param PyxisGQL pgql: PyxisGQL instance to use to verify images.
+            When None, new default PyxisGQL class is created.
         """
-        self.lb = lb if lb else LightBlue(
-            server_url=conf.lightblue_server_url,
-            cert=conf.lightblue_certificate,
-            private_key=conf.lightblue_private_key)
+        self.pgql = (
```
qixiang, line 35: could you rename the var to `self.pyxis` to keep it consistent with code in other modules?

freshmaker/image_verifier.py
```diff
@@ -22,120 +22,119 @@
...
         """
-        Verifies the Lightblue ContainerImage data.
+        Verifies the PyxisGQL ContainerImage data.
         Raises ValueError in case of error.
         """
+        if "brew" not in image:
+            raise ValueError("'brew' field not set in image")
+
+        if "build" not in image["brew"]:
+            raise ValueError("'brew.build' field not set in image")
```
qixiang, line 88: Any reason for adding this? In theory all images we handle should have `brew` field and we don't need to worry about this.

freshmaker/image_verifier.py
qixiang: Yes, it's supposed to be, when image is missing `content_sets`, freshmaker is unable to rebuild it, and the API should report error.

freshmaker/image_verifier.py
```diff
@@ -22,120 +22,119 @@
...
+        nvr = image["brew"]["build"]
 
-        repos = self.lb.find_container_repositories(query)
-        if not repos:
-            raise ValueError("Cannot get repository for image %s from Lightblue." % nvr)
-        if len(repos) != 1:
+        if "repositories" not in image or not image["repositories"]:
+            raise ValueError("Cannot get repository for image %s from Pyxis." % nvr)
+
+        if len(image["repositories"]) != 1:
```
qixiang, line 122: unlike the querying LightBlue repo way which should always returns a single repository (at least for redhat repositories), when you check published image's repositories it should always have at least 2 repositories, one is the internal OSBS build repo, so can't perform such check.

freshmaker/image_verifier.py
```diff
@@ -22,120 +22,119 @@
...
             raise ValueError(
-                "Image %s found in multiple repositories in Lightblue." % nvr)
+                "Image %s found in multiple repositories in Pyxis." % nvr)
 
+        image_repo = image["repositories"][0]
```
qixiang, line 126: As above, this could return internal OSBS build repository.

freshmaker/image_verifier.py
qixiang: Yes.

freshmaker/image_verifier.py
```diff
@@ -22,120 +22,119 @@
...
+                    break
+            else:
+                raise ValueError(
+                    "Only images with published repositories or unpublished exceptions can be rebuilt, but this image's repository is not published."
+                )
+
+        repos = self.pgql.find_repositories_by_repository_name(image_repo["repository"])
         return repos[0]
 
     def verify_image(self, image_nvr):
```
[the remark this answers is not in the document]
qixiang: that's fine.

tests/test_image_verifier.py
qixiang: Unlike the LightBlue APIs, our find_repositories_* APIs from pyxis_gql.py only returns list of dicts, not list of `ContainerRepository`.

tests/test_image_verifier.py
qixiang: right

tests/test_pyxis_gql.py
```diff
@@ -109,6 +109,50 @@ def test_pyxis_graphql_get_repository_by_registry_path():
     assert repository == result["get_repository_by_registry_path"]["data"]
 
 
+def test_pyxis_graphql_find_repositories_by_repository_name():
```
qixiang, line 112: I don't mind if you don't add the unit tests for new APIs in pyxis_gql.py, the test cases (here and the existing ones) actually doesn't make much sense, APIs from pyxis_gql.py are wrappers of pyxis calls, we mocked the APIs to return the fixtures and then compare the result with fixtures which should always pass, you can just bypass these tests and focus on other parts.

tests/test_image_verifier.py
ElenaKarolinaSemanova: I guess I should also only return dicts for images, not ContainerImage? @qixiang

freshmaker/pyxis_gql.py
ElenaKarolinaSemanova: @qixiang I meant this function, because I think the repository argument is a dictionary, not a list

qixiang: the repository names are what needed by this API, so it's unnecessary to have a dict as the parameter.

freshmaker/image_verifier.py
qixiang: Yes, we'd keep the same data structure in pyxis_gql module as what pyxis returns, in lightblue.py the images returned by LightBlue is also a list of dict, but converted to `ContainerImage` and `ContainerRepository`, so we also need to convert the result returned by pyxis_gql APIs to such objects in `lightblue.py` instead of `pyxis_gql.py`

tests/test_image_verifier.py
qixiang: right, APIs in pyxis_gql.py don't convert results to `ContainerImage` or `ContainerRepository`.

freshmaker/image_verifier.py
```diff
@@ -147,21 +139,22 @@ def verify_image(self, image_nvr):
...
 
-        images = self.lb.get_images_by_nvrs([image_nvr], include_rpm_manifest=False)
         if not images:
+            raise ValueError("No images found for the specified NVR")
+
+        if len(images) != 1:
             raise ValueError(
-                "No published images tagged by %r found in repository" % (
-                    repo["auto_rebuild_tags"]))
+                "Multiple instances of the image with `%s` nvr found in Pyxis." % images[0]["brew"]["build"])
```
ElenaKarolinaSemanova, line 149: @qixiang Does this make sense?

qixiang: The old LightBlue API (`find_container_images`) returns a list of `ContainerImage` and each of that represent an image with data for all available architectures, in other words, LightBlue returns a list of images (for a specific NVR) and then it was converted to a single `ContainerImage` with content sets from all architectures.

Our Pyxis API (`find_images_by_nvr`) returns a list of images, in most cases, the `len(images)` will be >1 because we didn't convert images of an NVR to a single object. We can verify each of images in this case:
```
content_sets = set()
for image in images:
    self._verify_image_data(image)
    content_sets.update(image["content_sets"])
...

return {images[0]["brew"]["build"]: sorted(content_sets)}
```

freshmaker/image_verifier.py
```diff
@@ -21,124 +21,117 @@
...
+        if "repositories" not in image or not image["repositories"]:
+            raise ValueError("Cannot get repository for image %s from Pyxis." % image["brew"]["build"])
 
-        repos = self.lb.find_container_repositories(query)
-        if not repos:
-            raise ValueError("Cannot get repository for image %s from Lightblue." % nvr)
-        if len(repos) != 1:
-            raise ValueError(
-                "Image %s found in multiple repositories in Lightblue." % nvr)
+        repos = list(filter(lambda x: 'osbs' not in x['repository'], image["repositories"]))
```
qixiang, line 119: You can filter with:
```
[r for r in image["repositories"] if r["registry"] != "conf.image_build_repository_registries"]
```

freshmaker/image_verifier.py
```diff
@@ -21,124 +21,117 @@
...
+        image_repo = repos[0]
 
+        if "published" not in image_repo or not image_repo["published"]:
+            for exc in conf.unpublished_exceptions:
+                if image_repo["repository"] == exc["repository"]:
+                    break
+            else:
+                raise ValueError(
+                    "Only images with published repositories or unpublished exceptions can be rebuilt, but this image's repository is not published."
+                )
```
qixiang, line 129: We should check against repo data returned from Pyxis's repository API, not the repo data in image, so you should find the repository from Pyxis first then verify it, the data show in image means how this image is present in the repository, not the repository itself.

But since we'll verify the repository data later:
```
def verify_image(...):
    ...
    repo = self._get_repository_from_image(image)                                                                                                                                     
    self._verify_repository_data(repo)
```
it's unnecessary to verify it here.

freshmaker/image_verifier.py
```diff
@@ -21,124 +21,117 @@
...
+
+        repos = self.pyxis.find_repositories_by_repository_name(image_repo["repository"])
```
qixiang, line 131: `pyxis.get_repository_by_registry_path` is preferred.
```
repo = self.pyxis.get_repository_by_registry_path(image_repo["registry"], image_repo["repository"] # returns a single repository
```

freshmaker/image_verifier.py
```diff
@@ -178,26 +172,69 @@ def verify_repository(self, repo_name):
...
-            image_data = {"content_sets": [], "tags": []}
-            for repodata in image['repositories']:
+            image_data: Dict[str, List[str]] = {"content_sets": [], "tags": []}
+            for repodata in image["repositories"]:
                 if repodata["repository"] == repo_name:
                     image_data["tags"] = [t["name"] for t in repodata["tags"]]
                     break
             image_data["content_sets"] = image["content_sets"]
-            data["images"][image.nvr] = image_data
+            data["images"][image["brew"]["build"]] = image_data
```
qixiang, line 232: This can be achieved by (not tested):
```
images_by_nvr = {}
for image in images:
    self._verify_image_data(image)
    nvr = image["brew"]["build"]
    images_by_nvr.setdefault(nvr, {"content_sets": [], "tags": []})
    # Image tags should be same for all architectures, just set it once
    if not images_by_nvr[nvr]["tags"]:
        for repodata in image["repositories"]:
            if repodata["repository"] == repo_name:
                images_by_nvr[nvr]["tags"] = [t["name"] for t in repodata["tags"]]
                break
    for cs in image["content_sets"]:
        if cs not in images_by_nvr[nvr]["content_sets"]:
            images_by_nvr[nvr]["content_sets"].append(cs)
```

freshmaker/image_verifier.py
```diff
@@ -178,26 +172,69 @@ def verify_repository(self, repo_name):
...
 
         if not data["images"]:
             raise ValueError(
-                "No published images tagged by %r found in repository" % (
-                    repo["auto_rebuild_tags"]))
+                "No images found in repository %r"
+                % (repo["repository"])
+            )
```
qixiang, line 238: This is misleading, it doesn't mean we don't find images in this repository, instead of that, we just can't find images in that repository which are tagged with the auto-rebuild tags, so this should not be changed.

freshmaker/pyxis_gql.py
```diff
@@ -194,6 +195,58 @@ def find_repositories(self, published=None, release_categories=None, auto_rebuil
 
         return repositories
 
+    def find_repositories_by_repository_name(self, repository: str = None) -> List:
```
qixiang, line 198: This parameter cannot be `None`.

freshmaker/pyxis_gql.py
```diff
@@ -496,3 +549,58 @@ def find_images_by_names(self, names):
             page_num += 1
 
         return images
+
+    def find_images_by_repository(self, repository: str = None, auto_rebuild_tags: List[str] = []) -> List:
```
qixiang, line 553: `repository` cannot be None.

freshmaker/image_verifier.py
```diff
@@ -178,26 +157,32 @@ def verify_repository(self, repo_name):
...
+        data: Dict[str, Dict[str, Dict[str, List[str]] | List[str]]] = {
             "repository": {"auto_rebuild_tags": repo["auto_rebuild_tags"]},
             "images": {}
         }
 
-        images = self.lb.find_images_with_included_rpms(
-            [], [], {repo["repository"]: repo}, include_rpm_manifest=False)
+        images = self.pyxis.find_images_by_repository(repo["repository"], repo["auto_rebuild_tags"])
+
+        images_by_nvr = {}
```
qixiang, line 167: mypy test failure:
```
freshmaker/image_verifier.py:167: error: Need type annotation for 'images_by_nvr' (hint: "images_by_nvr: Dict[<type>, <type>] = ...")
```

freshmaker/pyxis_gql.py
```diff
@@ -496,3 +549,59 @@ def find_images_by_names(self, names):
             page_num += 1
 
         return images
+
+    def find_images_by_repository(self, repository: str, auto_rebuild_tags: List[str] = []) -> List:
```
qixiang, line 553: Mutable default arguments should be avoided in Python: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

ElenaKarolinaSemanova:
@qixiang I resolved your comments; if you agree, we can merge this MR

qixiang (approved):
LGTM, I'll merge this.

qixiang:
@ElenaKarolinaSemanova there are some conflicts in freshmaker/pyxis_gql.py, could you rebase it?

ElenaKarolinaSemanova:
> @ElenaKarolinaSemanova there are some conflicts in freshmaker/pyxis_gql.py, could you rebase it?

@qixiang Done, we can merge after tests pass

PR #190, opened 2022-11-30
Split repos to small chunks to query LightBlue

qixiang:
LightBlue queries can take long time when:

1. Advisory is shipped to many repos
2. There are many repos are enabled for CVE rebuilds
3. Number of images grows to a large number in underlying database

Currently we run into an issue of LightBlue image query can't finish in 5mins, it doesn't make sense to always increase timeout from LightBlue side as time goes.

Splite the repos to small chunks can generate "smaller" queries which LightBlue can return more quickly.

### DIFF
--- a/freshmaker/lightblue.py
+++ b/freshmaker/lightblue.py
@@ -32,7 +32,7 @@ import dogpile.cache
 import kobo.rpmlib
 from concurrent.futures import ThreadPoolExecutor
 from http import HTTPStatus
-from itertools import groupby
+from itertools import groupby, islice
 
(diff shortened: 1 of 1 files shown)


qixiang:
This blocks our rebuilds for 8.7 images (more details in RHELWF-8101 and CLOUDBLD-12545), RHEL team is waiting for the rebuilds to meet RH SLA, I don't see anyone else online at this moment to review this and give me a +1, I'll just merge this without reviews and deploy it to prod, apologize for it.

@chuazhan @hluk @mulaievaRH

PR #192, opened 2022-12-21
Make compose with name:stream:version:context

chuazhan:
When using name:stream:version to create compose,
all contexts with the same name:stream:version will be returned in the result from MBS which is unexpected.

JIRA: RHELWF-7942


qixiang (approved):
LGTM

PR #193, opened 2023-01-13
Find container repositories with Pyxis GraphQL APIs

qixiang:
The LightBlue class is updated to use Pyxis GraphQL APIs to find    
container repositories, this is just a partial change so there are still    
a lot of variables, functions in this class have not been changed    
accordingly.    
      
JIRA: CWFHEALTH-1708


qixiang:
Close this one, all commits are included by #194, please review that one instead.

PR #198, opened 2023-02-08
Fix product pages API url

qixiang:
Product pages APIs now requires trailing slash in urls, otherwise the APIs return empty results unless we follow the redirects.
```
$ curl https://<product-pages-url>/pp/api/v1/releases/openshift-4.10/schedule-tasks
<nothing>
    
$ curl https://<product-pages-url>/pp/api/v1/releases/openshift-4.10/schedule-tasks/
<result as expected>
    
$ curl -L https://<product-pages-url>/pp/api/v1/releases/openshift-4.10/schedule-tasks
<result as expected>
```

### DIFF
--- a/freshmaker/utils.py
+++ b/freshmaker/utils.py
@@ -205,7 +205,7 @@ def get_ocp_release_date(ocp_version):
 
     ocp_release = f"openshift-{ocp_version}"
 
-    url = f"{conf.product_pages_api_url.rstrip('/')}/releases/{ocp_release}/schedule-tasks"
+    url = f"{conf.product_pages_api_url.rstrip('/')}/releases/{ocp_release}/schedule-tasks/"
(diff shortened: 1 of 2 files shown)


qixiang:
Error  raised by freshmaker:
```
...
  File "/src/freshmaker/pyxis.py", line 143, in ocp_is_released
    ga_date_str = get_ocp_release_date(ocp_version)
  File "/src/freshmaker/utils.py", line 220, in get_ocp_release_date
    return resp.json()[0]['date_finish']
IndexError: list index out of range
```

we should handle the request more properly, but at this moment we need a quick fix to avoid exceptions for bundle rebuilds, so I only fixed the url in this PR.

ElenaKarolinaSemanova:
+1

PR #199, opened 2023-02-08
Pull fedora 35 image from fedora registry

qixiang:
Fedora 35 image has been removed from quay.io/fedora/fedora, use fedora registry, later we should upgrade to newer version.


ronnll:
+1 LGTM

ronnll (approved):

PR #200, opened 2023-02-08
Fix get product version which doesn't have a GA date

qixiang:
A product version can exist in product pages but doesn't have the GA date, freshmaker should also handle this case.


ronnll (approved):
+1 LGTM!

PR #202, opened 2023-02-21
Test

ronnll:


tests/test_lightblue.py
```diff
@@ -1306,9 +1310,77 @@ def test_find_all_repositories(self, exists, gql_client):
...
+                    "architecture": "amd64",
+                    "brew": {"build": "parent-1-2"},
+                    "content_sets": ["dummy-content-set-1"],
+                    "edges": {
+                        "rpm_manifest": {
+                            "data": {
+                                "rpms": [
+                                    {
+                                        "name": "openssl",
+                                        "nvra": "openssl-1.2.3-2.module+el8.0.0+3248+9d514f3b",
```
qixiang, line 1330: This is not a valid `nvra`, it doesn't include `arch`, it will cause problem in code to parse this value.

tests/test_lightblue.py
```diff
@@ -1306,9 +1310,77 @@ def test_find_all_repositories(self, exists, gql_client):
...
+                                        "srpm_name": "openssl",
+                                        "srpm_nevra": "openssl-1.2.3-2.module+el8.0.0+3248+9d514f3b",
```
qixiang, line 1332: This is not a valid `nevra`, it doesn't include `arch`.

tests/test_lightblue.py
```diff
@@ -2379,18 +2714,85 @@ def test_images_with_modular_container_image(
             ["dummy-content-set-1", "dummy-content-set-2"], ["openssl-1.2.3-2.module+el8.0.0+3248+9d514f3b.src"], repositories)
```
qixiang, line 2714: `openssl-1.2.3-2.module+el8.0.0+3248+9d514f3b.src` is not a valid NVR, it includes arch of `src`, it will cause problem when freshmaker parse this and result in unexpected result.

PR #204, opened 2023-02-23
Pyxis kerberos (squashed)

FernandesMF:
(This Pull Requests continues #201.)

Pyxis supports authentication through kerberos. This method is preffered over SSL certificates because we won't need to renew our certificates periodically anymore.

This contribution implements kerberos authentication to the pyxis-gql.py module through the requests-kerberos library.

The Jira card associated with this contribution is https://issues.redhat.com/browse/CWFHEALTH-1762
### Testing remarks

I ran all the unit tests successfully. I also made manual e2e tests to make sure all queries were working properly. I used my personal kerberos credentials and send the requests to the dev pyxis gql endpoint.

I called all methods of the PyxisGQL class except for query, _get_repo_projection and _get_image_projection, which are called by the other methods. I ran the other methods just as in the unit tests and most of them returned empty lists as result (since the queried values were just mock values).

The methods find_images_by_nvr and find_repositories returned actual results. find_repositories took a pretty long time, and the result was quite long. I'm quoting an excerpt of the result from find_repositories below:

``` python
[
 ...
 {'auto_rebuild_tags': [],
  'registry': 'registry.access.redhat.com',
  'release_categories': None,
  'repository': 'rhacm2/acm-cluster-templates-rhel8-operator'}
]
```

### Questions

I have removed the `cert` parameter from `PyxisGQL` and `pyxis_cert` from `ContainerAPI`. Although these are silly changes, they are still breaking changes.

Please let me know if you think something will break in production, or you you think we should take a backwards-compatible approach (like accepting the parameters for now, but issuing a deprecation warning).

### DIFF
--- a/freshmaker/pyxis_gql.py
+++ b/freshmaker/pyxis_gql.py
@@ -21,37 +21,22 @@
 
 from functools import cached_property
 
-from gql import gql, Client
+from gql import Client, gql
 from gql.dsl import DSLQuery, DSLSchema, dsl_gql
 from gql.transport.requests import RequestsHTTPTransport
+from requests_kerberos import OPTIONAL, HTTPKerberosAuth
 
 
 class PyxisGQLRequestError(RuntimeError):
     pass
 
 
-class RequestsHTTPTransportWithCert(RequestsHTTPTransport):
-    """A modified requests transport to support certificate authentication"""
-
-    def __init__(self, *args, **kwargs):
-        self.cert = kwargs.pop("cert", None)
-        if self.cert is None:
-            raise RuntimeError("Missing required keyword argument: cert")
-        super().__init__(*args, **kwargs)
-
-    def connect(self):
-        super().connect()
-        self.session.cert = self.cert
-
-
 class PyxisGQL:
-    def __init__(self, url, cert):
+    def __init__(self, url):
         """Create authenticated Pyxis GraphQL session"""
-        transport = RequestsHTTPTransportWithCert(
-            url=url,
-            retries=3,
-            cert=cert,
-        )
+        pyxis_krb_auth = HTTPKerberosAuth(mutual_authentication=OPTIONAL, force_preemptive=True)
+        transport = RequestsHTTPTransport(url=url, retries=3, auth=pyxis_krb_auth)
+
         # Fetch the schema from the transport using an introspection query
         self._client = Client(transport=transport, fetch_schema_from_transport=True)
 
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ freshmaker.db
 .mypy_cache/
 .dmypy.json
 dmypy.json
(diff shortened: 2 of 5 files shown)


FernandesMF:
(I force-pushed a new commit that happened upstream, and also a correction to the Jira card number in my commit message)

freshmaker/pyxis_gql.py
```diff
@@ -21,37 +21,22 @@
...
-    def __init__(self, url, cert):
+    def __init__(self, url):
         """Create authenticated Pyxis GraphQL session"""
-        transport = RequestsHTTPTransportWithCert(
-            url=url,
-            retries=3,
-            cert=cert,
-        )
+        pyxis_krb_auth = HTTPKerberosAuth(mutual_authentication=OPTIONAL, force_preemptive=True)
+        transport = RequestsHTTPTransport(url=url, retries=3, auth=pyxis_krb_auth, verify=False)
```
qixiang, line 38:
```suggestion
        transport = RequestsHTTPTransport(url=url, retries=3, auth=pyxis_krb_auth)
```
I don't remember why `verify=False` was added before, let's remove it.

qixiang:
LGTM, just a minor comment.
It's better to update in the same PR which is better for checking review comments, so you don't need to do this in the future.

qixiang (approved):

freshmaker/pyxis_gql.py
FernandesMF: I believe this flag turns off the SSL verification during the requests. If I don't set it to false, I run into the following error when I try to make any of the queries:

> `requests.exceptions.SSLError: HTTPSConnectionPool(host='graphql.pyxis.dev.engineering.redhat.com', port=443): Max retries exceeded with url: /graphql/ (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:997)'))) `

I believe this flag will be necessary. Is it okay if we keep it?

qixiang: You need to have the internal Root CA installed, if you're running inside a virtual env and the `python-requests`/`python-certifi` are installed with `pip`, then you have to workaround this by:
```
$ export REQUESTS_CA_BUNDLE=/etc/pki/tls/certs/ca-bundle.crt
```
or
```
$ cp /etc/pki/tls/certs/ca-bundle.crt $(python -c "from requests.certs import where; print(where())")
```
I'll share the doc about installing the internal CA in jira, when you're not sure whether it's safe or not, it's better to not expose any internal links (like the pyxis graphql server url).

FernandesMF: That was exactly what I was missing, thanks a lot!

I removed the 'verify=False' flag and tested everything again just to be sure. I'll squash this together into this commit.

PR #205, opened 2023-03-07
feat: bump fedora version

FernandesMF:
The fedora version that FM is using (f35) reached its end of life. Therefore, we need to update it.

I'm editing the old PR text and adding a few remarks. I'll mark them with "**[<ins>new</ins>]**".

List of changes:
- update package jsonformatter
- register api's v1 and v2 with different names in `view.py`
- gql installation through pip during the build process; I believe it is not available through dnf
- type hint in line 208 of `conf/config.py` (it was failing a flake8 check)

Please let me know if any of these are actually unnecessary!

### Remarks about testing

Things that I managed to do:
- I ran unit tests (and tox) locally, all good!
- **[<ins>new</ins>]**  I built the image and deployed it to the dev environment! I also adapted and ran the "dry_run" script and got a good return:
```json
{
    "builds":[],
    "depending_events":[],
    "depends_on_events":[],
    "dry_run":true,
    "event_type_id":13,
    "id":52026,
    "message_id":"manual_rebuild_1678894295.1448815",
    "requested_rebuilds":[],
    "requester":"mfoganho",
    "requester_metadata":{},
    "search_key":"110818",
    "state":0,
    "state_name":"INITIALIZED",
    "state_reason":null,
    "time_created":"2023-03-15T15:31:35Z",
    "time_done":null,
    "url":"/api/1/events/52026"
}
```
- **[<ins>new</ins>]** I left it running for about day and checked the splunk logs, there was no error so far (but also no real rebuild).
- I ran unit tests inside the container. There was one failing test, but that same test was passing when I ran tox in my local codebase. Perhaps it is because I didn't set up the network for the container correctly?
```
================================================================ short test summary info =================================================================
FAILED tests/test_views.py::TestOpenIDCLogin::test_openidc_manual_trigger_unauthorized - AssertionError: '500 INTERNAL SERVER ERROR' != '401 UNAUTHORIZED'
===================================================== 1 failed, 515 passed, 16552 warnings in 19.46s =====================================================
```

### **[<ins>new</ins>]** Questions
- Do you think this enough testing? Should we wait until a rebuild is actually performed? Or, perhaps, trigger one manually?

JIRA: CWFHEALTH-1788

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:35-x86_64
+FROM registry.fedoraproject.org/fedora:37-x86_64
 
 LABEL \
     name="Freshmaker application" \
@@ -7,7 +7,7 @@ LABEL \
 
 
 # Use Copr repo for python3-rhmsg package
-RUN dnf install -y 'dnf-command(copr)' && dnf copr enable -y mikeb/python-rhmsg
+RUN dnf install -y 'dnf-command(copr)' && dnf copr enable -y qwan/python-rhmsg
 
 COPY yum-packages.txt /tmp/yum-packages.txt
 
@@ -24,7 +24,7 @@ RUN \
     echo '' > requirements.txt && \
     pip3 install . --no-deps
 
-RUN pip install jsonformatter==0.3.1
+RUN pip install jsonformatter==0.3.2 gql[requests]==3.3.0
 
 RUN mkdir /var/log/freshmaker/
 
--- a/conf/config.py
+++ b/conf/config.py
@@ -205,7 +205,7 @@ class BaseConfiguration(object):
     }
 
     # repositories that should be searched for unpublished images, specifically because of EUS base images
-    UNPUBLISHED_EXCEPTIONS = []
+    UNPUBLISHED_EXCEPTIONS: list[dict[str, str]] = []
 
 
 class DevConfiguration(BaseConfiguration):
--- a/freshmaker/views.py
+++ b/freshmaker/views.py
@@ -814,7 +814,7 @@ def register_api_v1():
                              view_func=view,
                              **val['options'])
 
-    app.register_blueprint(monitor_api)
+    app.register_blueprint(monitor_api, name="monitor_api_v1")
 
 
 def register_api_v2():
@@ -829,7 +829,7 @@ def register_api_v2():
                              view_func=view,
                              **val['options'])
 
-    app.register_blueprint(monitor_api)
+    app.register_blueprint(monitor_api, name="monitor_api_v2")
 
 
 register_api_v1()
--- a/requirements.txt
+++ b/requirements.txt
@@ -84,7 +84,7 @@ itsdangerous==2.1.2
     # via flask
 jinja2==3.1.2
     # via flask
-jsonformatter==0.3.1
+jsonformatter==0.3.2
     # via -r requirements.in
 kitchen==1.2.6
     # via


Dockerfile
```diff
@@ -24,14 +24,18 @@ RUN \
     echo '' > requirements.txt && \
     pip3 install . --no-deps
 
-RUN pip install jsonformatter==0.3.1
+RUN pip install jsonformatter==0.3.2 gql[requests]==3.3.0
```
qixiang, line 27: correct, we haven't added gql in Dockerfile and it's not available in yum repos.

Dockerfile
```diff
@@ -24,14 +24,18 @@ RUN \
...
 
 RUN mkdir /var/log/freshmaker/
 
 RUN mkdir -p /usr/share/freshmaker && cp contrib/freshmaker.wsgi /usr/share/freshmaker/
 # Delete the default logging configuration
 RUN rm -f fedmsg.d/freshmaker-logging.py
 
+RUN mkdir /etc/freshmaker
+COPY ./conf/config.py /etc/freshmaker/config.py
+COPY ./conf/koji.conf /etc/koji.conf
```
qixiang, line 37: Freshmaker heavily relies on the internal infrastructure but this is a public repository, so we can't easily include our configurations in this repo, when we deploy freshmaker in OpenShift, the config files will be mounted through volumes, so we don't need to copy these config templates to container.

qixiang (commented):
> List of changes:
> 
> * update package jsonformatter
> * register api's v1 and v2 with different names in `view.py`
> * include koji config file into the container
> * gql installation through pip during the build process; I believe it is not available through dnf
> * type hint in line 208 of `conf/config.py` (it was failing a flake8 check)

Looks good to me except that we don't need to copy the koji config file.

> * I ran unit tests (and tox) locally, all good!

The tests for this PR passed, so yes, you're right.

> * I ran unit tests inside the container. There was one failing test, but that same test was passing when I ran tox in my local codebase. Perhaps it is because I didn't set up the network for the container correctly?
> 
> ```
> ==================================================================================== short test summary info =====================================================================================
> FAILED tests/test_views.py::TestOpenIDCLogin::test_openidc_manual_trigger_unauthorized - AssertionError: '500 INTERNAL SERVER ERROR' != '401 UNAUTHORIZED'
> ========================================================================= 1 failed, 515 passed, 16552 warnings in 19.46s =========================================================================
> ```

You don't have to run the unit tests in container, but this test failure looks to be some defect in our unit test that relies on the infra it runs on, you can investigate this when you have time.

> Things that I didn't manage to do:
> 
> * I did not manage to make a manual test. I tried to adapt `dry_run_script.py` (available in [freshmaker-dev-scripts repository), but I didn't figure how to expose a port of the image and submit a message/request to it. I am not very experienced with containers, though, so maybe this is actually easy.

It's kind of difficult to run the freshmaker containers locally to provide the full functionalities, the main reason is it heavily relies on our internal infrastructure, we have some internal docs (but are old) describe how to run with containers. I'll share the links to you.

> * I did not deploy the image to dev environment, nor check if it was working with all the things it has to interface.
> 

Deploy it to dev env could be the easiest way for us to test the functionalities, I can share more on how we can do this.

> I do not know how to go about these two. If you could provide directions, I would greatly appreciate!
> 
> JIRA: CWFHEALTH-1788

qixiang:
The docs build failed, so seems we have to do more changes to fix that.

Dockerfile
FernandesMF: I see! I removed these lines, thanks for the explanation!

qixiang (approved):
LGTM, thanks.

PR #210, opened 2023-03-28
chore: remove old jenkinsfile

FernandesMF:
The jenkinsfile is obsolete and is not used anymore in this project.


FernandesMF:
I made a small mess with my fork and it is conflicting with the upstream branch. I'll solve that first and then open a new PR.

PR #211, opened 2023-03-28
chore: remove old jenkinsfile, update ubuntu runners

FernandesMF:
The jenkinsfile is obsolete and is not used anymore in this project.

Also, Github Actions will drop support for Ubuntu 18.04 runners:
https://github.com/actions/runner-images/issues/6002
therefore, we need their version in our actions.


qixiang (approved):
LGTM

PR #215, opened 2023-04-14
Cache the Pyxis query results

qixiang:
Use dogpile.cache to cache the Pyxis result fetched by APIs in pyxis_gql.py module.

The `get-or-create` pattern is the entire key to the `dogpile` system, it coordinates a single value creation operation among many concurrent get operations for a particular key. The default caching key is used in the pyxis_gql module, which is "<module>:<fn_name>|<args>", calling a function in multiple instances of PyxisGQL will share the same key and result will be cached.

JIRA: CWFHEALTH-1897

### DIFF
--- a/freshmaker/pyxis_gql.py
+++ b/freshmaker/pyxis_gql.py
@@ -19,11 +19,14 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+import dogpile.cache
 from functools import cached_property
 
 from gql import Client, gql
 from gql.dsl import DSLQuery, DSLSchema, dsl_gql
 from gql.transport.requests import RequestsHTTPTransport
+
+from freshmaker import conf
 
 PYXIS_PAGE_SIZE = 250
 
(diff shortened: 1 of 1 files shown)


qixiang:
Github is having some problem of scheduling and running the tests, but the tests passed in my local environment.

FernandesMF:
The implementation of the cache is very elegant, awesome work!

I was a bit hesitant about the dogpile library at first, because it seemed like a smaller project. But then I researched it a bit and found that it is well maintained, and that its predecessor project Beaker was quite popular and well coded. I guess that won't be a problem at all!

qixiang:
@FernandesMF could you give a +1 to this PR?
I guess the CI didn't start is because we're still using the old ubuntu runner and it's probably not available now. It has been fixed in `main` branch (also the docs build failure), I can rebase this `pyxis` branch on top of `main` branch` later, that will requires a force push to the `pyxis` branch to fix the conflicts, so I'd prefer to do that before we merge all commits in `pyxis` branch to `main`.

FernandesMF (approved):

FernandesMF:
Sorry about forgetting the approval!

PR #216, opened 2023-05-09
Keep olm.substitutesFor from original image

qixiang:
When original image has `olm.substitutesFor` annotation, in most cases it was added by Freshmaker with the `operator_csv_modifications_url` in build task parameters, it's not in any file from dist-git repo, when Freshmaker rebuild this image and don't add any new `olm.substitutesFor` annotation, the `olm.substitutesFor` annotation in original image will be lost, this is wrong.

JIRA: CWFHEALTH-2003


FernandesMF (approved):
Looks good to me!

PR #218, opened 2023-05-10
Asynchronous Pyxis GQL module

FernandesMF:
This commit introduces the asynchronous Pyxis GQL module. It will hopefully allow for a better performance when interfacing with the Pyxis GQL endpoint.

The changes include:
- the `pyxis_gql_async.py` module,
- unit tests of the module,
- a documentation page with usage examples,
- updates to the dependencies of the project.

Most of the code is reused from the synchronous Pyxis GQL module.

#### Questions
I have a few questions that I'd like to ask you regarding the async module.
- Neither `AIOHTTPTransport`, nor `aiohttp.ClientSession` have a 'retries' parameter, which we did have in the trasnport used in the synchronous module; do you think this is an important feature? Should I make a manual implementation of that?
- In line 166-169, there is an `if` block that I think is behaving weridly: if the parameter `published=False`, we will still append `True` to the query filter dictionary; is that the desired behaviour?
- many functions make queries sequentially, because of the pagination logic; can you think of a way we could avoid that and make it more suitable for async?

#### Documentation
I zipped the html files generated with sphinx, in case you don't want to clone the repo just to compile the documentation. Here it is:
[docs.zip](https://github.com/redhat-exd-rebuilds/freshmaker/files/11445472/docs.zip)

I know this is a pretty big PR, so thanks a lot for your time! 😄 

JIRA: [CWFHEALTH-1913]


qixiang:
Hi @FernandesMF LGTM, well done.

This has been done on top of the pyxis_gql module from `main` branch, we'll merge the updated modules from `pyxis` branch to `main` branch later, so there will be some work to update this module once the `pyxis` branch is merged, I'm fine to do that later.

Regarding the retry mechanism, yes, we'll need to add that if the async transport doesn't support retries, we can do that later as well, freshmaker has `retry` decorator in `utils.py`, but I'd prefer to adopt [backoff](https://github.com/litl/backoff) module, the long term plan is replacing all `retry` decorators with the backoff decorators.

qixiang (approved):

qixiang:
@FernandesMF I forgot to mention that there are some tailing spaces in `docs/pyxis_gql_async.rst`, could you remove that before merging this?

freshmaker/pyxis_gql_async.py
```diff
@@ -0,0 +1,472 @@
...
+        """
+        query_filter: dict = {}
+        query_filter["and"] = []
+        # Query Red Hat repositories only
+        query_filter["and"].append({"vendor_label": {"eq": "redhat"}})
+
+        if published is not None and isinstance(published, bool):
+            query_filter["and"].append(
+                {"published": {"eq": True}}
+            )  # TODO what if published=False? should True be still appended?
```
qixiang, line 169: This has been fixed in `pyxis` branch, some modules in `main` branch are outdated.

qixiang (commented):
> In line 166-169, there is an if block that I think is behaving weridly: if the parameter published=False, we will still append True to the query filter dictionary; is that the desired behaviour?

Commented in code line.

> many functions make queries sequentially, because of the pagination logic; can you think of a way we could avoid that and make it more suitable for async?

For a single query API, it's acceptable, the most time consuming call that is impacted by the pagination is the `find_repositories`, but it will only be called once in an event. You can get the total number of items in the first call and then calculate the pages and fetch the result in parallel, however I think the current implementation of looping the pages is much easier to maintain.

freshmaker/pyxis_gql_async.py
FernandesMF: I see. Yeah, I definitely missed out quite a bit by not getting the code from the `pyxis` branch 😅

FernandesMF:
I looked into the code from `pyxis` branch and there are indeed some important differences wrt the code in `main`, which I based this commit on.
I made another branch in my fork that tracks `pyxis`, and I'll reimplement this module in a branch stemming from that.

PR #220, opened 2023-05-29
Limit number of result while querying for images match name and version

qixiang:
Due to ISV-3642, in some cases, Pyxis is not able to return result when a query in API find_images_by_name_version is submitted, we need to limit the number of images in such queries to workaround the issue.

JIRA: CWFHEALTH-2047


FernandesMF (approved):
Looks good to me!

PR #222, opened 2023-06-01
Certificate-based authentication for Pulp

FernandesMF:
Pulp is deprecating user/password authentication, and switching to certificate-based auth instead.

This commit:
- implements the change in the authentication method in the pulp module
- updates unit tests for teh pulp module
- fixes a call in `freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py`
- substitutes the credential templates in `freshmaker/config.py`

JIRA: CWFHEALTH-2004

#### Testing remarks
I have been able to test the access of our certificates to the Pulp API, and also the updated authentication method in the Pulp requests:
- I ran the curl command the Pulp team suggested in their documentation page and checked that the certificates (both dev and prod) are working fine
- I made a script and called the Pulp class and its post method (using the dev certificates), it also worked fine.

However, I have not been able to check if the new credentials are being served correctly.

### DIFF
--- a/freshmaker/config.py
+++ b/freshmaker/config.py
@@ -311,14 +311,14 @@ class Config(object):
             'type': str,
             'default': '',
             'desc': 'Server URL of Pulp.'},
-        'pulp_username': {
+        'pulp_crt_path': {
             'type': str,
             'default': '',
-            'desc': 'Username to login Pulp.'},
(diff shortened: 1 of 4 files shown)


qixiang (approved):
LGTM

PR #223, opened 2023-06-09
Make page_size in Pyxis requests configurable

qixiang:
Introduced two config variables:

1. pyxis_default_page_size
2. pyxis_small_page_size

which can be specified as page_size in Pyxis requests.

Some Pyxis queries are resource consuming and Pyxis may not be able to
handle them in a time manner, use a low number of page_size can help in
such cases.

The resource consuming queries usually have complex search criteria and
Pyxis has to check a lot of entities in backend, or the search hit a
large number of entities in database which is unpredictable at client
side, in such cases, use the pyxis_small_pyxis_size can probably help
Pyxis to handle the request.

The find_images_by_installed_rpms API is one of such queries, so we use
the configured small page_size in this API.

### DIFF
--- a/freshmaker/config.py
+++ b/freshmaker/config.py
@@ -398,6 +398,16 @@ class Config(object):
             'default': [],
             'desc': 'Query Pyxis for index images only with these organizations'
         },
+        'pyxis_default_page_size': {
+            'type': int,
+            'default': 200,
+            'desc': 'Default page size to be used in Pyxis requests'
+        },
+        'pyxis_small_page_size': {
+            'type': int,
+            'default': 50,
+            'desc': 'Small page size to be used in Pyxis requests'
+        },
         'product_pages_api_url': {
             'type': str,
             'default': '',
--- a/freshmaker/pyxis_gql.py
+++ b/freshmaker/pyxis_gql.py
@@ -29,8 +29,6 @@ from gql.transport.requests import RequestsHTTPTransport
 
 from freshmaker import conf
 
-PYXIS_PAGE_SIZE = 200
-
 
 class PyxisGQLRequestError(RuntimeError):
     pass
@@ -160,7 +158,7 @@ class PyxisGQL:
         while True:
             query_dsl = ds.Query.find_repositories(
                 page=page_num,
-                page_size=PYXIS_PAGE_SIZE,
+                page_size=conf.pyxis_default_page_size,
                 filter=query_filter,
             ).select(
                 ds.ContainerRepositoryPaginatedResponse.error.select(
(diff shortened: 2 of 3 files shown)


freshmaker/config.py
```diff
@@ -398,6 +398,16 @@ class Config(object):
             'default': [],
             'desc': 'Query Pyxis for index images only with these organizations'
         },
+        'pyxis_default_page_size': {
```
hluk, line 401: It is not clear to me from the description when I would use the `default` one or the `safe` one.

hluk (commented):
There is a LOT of duplicate code that handles the pagination.

freshmaker/config.py
qixiang: Actually `safe` is not 100% safe :(

When we submit a request to Pyxis, Pyxis will check repositories, images at backend, if a query has complex filters and Pyxis has to check a lot of entities in backend, it may not be able to return result to us (sometimes it even returns different errors), but it's hard to tell when we should use a small page size, and which number is small enough to be safe.

There are several factors can impact on this:
1. The query criteria. For example if we want to find images that have any of the specified content sets enabled, then the number of the content sets makes differences, 4 and 20 content sets require different time to complete the filter at Pyxis backend.
2. The number of images hit the search criteria, e.g. there is problematic collection, the e2e images have thousands of images that have unexpected tags, if our search hit these images, Pyxis will not be able to handle it.

The `safe` page size here is a number I found that can work in most of cases during my testing (at least I haven't hit any error with this number so far), with this change it will be easier for us to deploy Freshmaker with new page size when we run into some issues.

And welcome for any better name :)

Regarding the duplicate code for pagination, the queries actually not duplicate, they have different filters and data selection and hard to simply, but the code of handling the returned paginated result do have spaces to improve. Currently some of the implementation are far away from the ideal solution, we tried to catch up with plan of decommissioning LightBlue with some short-term implementation, some modules are really making confusions (e.g: pyxis.py, pyxis_gql.py, pyxis_gql_async.py), there're a lot of stuff need to be refactored.

FernandesMF (approved):
I agree with the points raised in the discussion, we can certainly improve the code with refactoring, though it might not be that simple.

For now, the new pyxis page parameter looks good to me!

freshmaker/config.py
hluk: Thanks for the detailed description.

Would it be possible to guess the page size from query parameters? Though this would make the configuration and/or the logic somewhat complex and difficult to maintain.

Or make it at most 50 for all Pyxis queries?

Or rename `pyxis_safe_page_size` to `pyxis_small_page_size` and mention in description that it is for complex/large queries?

hluk (approved):

PR #225, opened 2023-06-12
NVR explicit replacing in manual bundle rebuilds

FernandesMF:
- added a new attribute `manual_image_remap` to the `ManualBundleRebuildEvent` class
- implemented logic to make the remapping of nvrs, following that new attribute, in `HandleBotasAdvisory`
- implemented logic to conditionally include the remapping information to the event in `FreshmakerManualRebuildParser`

### DIFF
--- a/freshmaker/handlers/botas/botas_shipped_advisory.py
+++ b/freshmaker/handlers/botas/botas_shipped_advisory.py
@@ -143,12 +143,24 @@ class HandleBotasAdvisory(ContainerBuildHandler):
         log.debug("Getting NVR mapping of original images to rebuilt images")
         nvrs_mapping = self._create_original_to_rebuilt_nvrs_map()
 
+        # Apply manual NVR remapping, if the user requested it. All the requested overrides are
+        # kept, and nvrs_mapping is incremented with new mappings if they are requested but were
+        # not present previously. E.g., if nvr_mapping={or_nvr1:new_nvr1, or_nvr2:new_nvr2} and
+        # manual_remapping = {or_nvr2:man_nvr2, or_nvr3:man_nvr3}, then after the manual remapping
+        # nvr_mapping={or_nvr1:new_nvr1, or_nvr2:man_nvr2, or_nvr3:man_nvr3}.
+        if (self.event.get("requester_metadata_json", False) and
+                self.event.requester_metadata_json.get("bundle_related_image_overrides", False)):
+            log.info("Performing manual NVR remapping")
+            manual_remapping = self.event.requester_metadata_json["bundle_related_image_overrides"]
+            for x in manual_remapping:
+                nvrs_mapping[x] = manual_remapping[x]
+
         original_nvrs = nvrs_mapping.keys()
         if not original_nvrs:
             return None, "Can't find any published original builds for images in advisory."
 
         log.info(
-            "Orignial NVRs of build in advisory %s are: %s",
+            "Original NVRs of build in advisory %s are: %s",
             self.event.advisory.errata_id,
             " ".join(original_nvrs),
         )
--- a/freshmaker/views.py
+++ b/freshmaker/views.py
@@ -398,6 +398,19 @@ def _validate_rebuild_request(request):
     if not isinstance(data.get('force', False), bool):
         return json_error(400, 'Bad Request', '"force" must be a boolean.')
 
+    if data.get('bundle_related_image_overrides', False) and not container_images:
+        return json_error(
+            400,
+            'Bad Request',
+            'Manual image overriding allowed only when "container_images" is set explicitly'
+        )
+
+    if (data.get('bundle_related_image_overrides', False) and
+            not isinstance(data.get('bundle_related_image_overrides'), dict)):
+        return json_error(
+            400, 'Bad Request', '"bundle_related_image_overrides" must be a dictionary'
+        )
+
     return None
 
 
--- a/tests/handlers/botas/test_botas_shipped_advisory.py
+++ b/tests/handlers/botas/test_botas_shipped_advisory.py
@@ -597,6 +597,225 @@ class TestBotasShippedAdvisory(helpers.ModelsTestCase):
         assert bundles_to_rebuild[0] in self.handler._prepare_builds.call_args.args[0]
         assert bundles_to_rebuild[1] in self.handler._prepare_builds.call_args.args[0]
 
+    @patch.object(conf, "handler_build_allowlist", new={
+        "HandleBotasAdvisory": {"image": {"advisory_name": "RHBA-2020"}}
+    })
+    @patch('freshmaker.models.Event.get_artifact_build_from_event_dependencies')
+    def test_handle_bundle_rebuild_manual_nvr_override(self, get_dependent_event_build):
+
+        """ Test handling of MANUALLY triggered bundle rebuild with manual nvr override.
+
+        Bundle 1 images are left alone, bundle 2 images are manually overriden, and there
+        is an override that is introduced manually in the request (#FIXME specify)
+        """
+        # operators mapping
+        nvr_to_digest = {
+            "original_1": "original_1_digest",
+            "some_name-1-12345": "some_name-1-12345_digest",
+            "original_2": "original_2_digest",
+            "some_name_2-2-2": "some_name_2-2-2_digest",
+            "manual_name_2-2-2": "manual_name_2-2-2_digest",
+            "original_3": "original_3_digest",
+            "manual_name_3-3-3": "manual_name_3-3-3_digest",
+        }
+        # related image digest -> bundle
+        bundles_with_related_images = {
+            "original_1_digest": [
+                {
+                    "bundle_path_digest": "bundle_with_related_images_1_digest",
+                    "csv_name": "image.1.2.3",
+                    "version_original": "1.2.3",
+                },
+            ],
+            "original_2_digest": [
+                {
+                    "bundle_path_digest": "bundle_with_related_images_2_digest",
+                    "csv_name": "image.1.2.4",
+                    "version_original": "1.2.4",
+                },
+            ],
+            "original_3_digest": [
+                {
+                    "bundle_path_digest": "bundle_with_related_images_3_digest",
+                    "csv_name": "image.1.2.5",
+                    "version_original": "1.2.5",
+                },
+            ],
+        }
+        image_by_digest = {
+            "bundle_with_related_images_1_digest": {"brew": {"build": "bundle1_nvr-1-1"}},
+            "bundle_with_related_images_2_digest": {"brew": {"build": "bundle2_nvr-1-1"}},
+            "bundle_with_related_images_3_digest": {"brew": {"build": "bundle3_nvr-1-1"}},
+        }
+        image_by_nvr = {
+            "bundle1_nvr-1-1": {"brew": {"build": "bundle1_nvr-1-1"}},
+            "bundle2_nvr-1-1": {"brew": {"build": "bundle2_nvr-1-1"}},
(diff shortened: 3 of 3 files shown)


freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -92,6 +92,10 @@ def handle(self, event):
             self.db_event.transition(EventState.SKIPPED, msg)
             return []
 
+        if isinstance(self.event, ManualBundleRebuildEvent) and self.event.manual_image_remap:
+            log.info("Performing manual NVR remapping")
+            self._manual_bundle_remap(bundles_to_rebuild)  #inplace substitutions  to bundles_to_rebuild
```
qixiang, line 97: We are not overriding the bundles to be rebuilt, we want to override the related images in bundle while rebuilding the bundles.

When a rebuilds advisory (filed by Botas) is shipped, it has a list of images which were built by Freshmker, for each image, it has corresponding original image, so we can query Errata and Freshmaker DB to get some data like this:
```
original_nvr1 -> rebuilt_nvr1
original_nvr2 -> rebuilt_nvr2
original_nvr3 -> rebuilt_nvr3
```
the `rebuilt_nvr*` are images in the advisory.

Then in this handler, Freshmaker will try to find the bundles that are referencing the following images:
```
original_nvr1
original_nvr2
original_nvr3
```
If any bundle is referencing any of the listed original images, then Freshmaker will rebuild that bundle by:
1. replace `original_nvr1` with `rebuilt_nvr1`
2. replace `original_nvr2` with `rebuilt_nvr2`
3. replace `original_nvr3` with `rebuilt_nvr3`

Our purpose here is to support overriding the mapping, for example, if we have something like:
```
'bundle_related_image_overrides': {'original_nvr3': 'some_nvr3', 'original_nvr4': 'some_nvr4'}
```
then we should be able to override the mapping to:
```
original_nvr1 -> rebuilt_nvr1
original_nvr2 -> rebuilt_nvr2
original_nvr3 -> some_nvr3
original_nvr4 -> some_nvr4
```
we can override the mappings after https://github.com/redhat-exd-rebuilds/freshmaker/blob/9788db751ee6c94c6b59c01464622c53c4bdc494/freshmaker/handlers/botas/botas_shipped_advisory.py#L144

But to prevent unexpected rebuilds, I'd suggest only allow the overrides when bundle images (`container_images` in request) are specified explicitly in the request, we can add the check in https://github.com/redhat-exd-rebuilds/freshmaker/blob/9788db751ee6c94c6b59c01464622c53c4bdc494/freshmaker/views.py#LL361C7-L361C7

freshmaker/parsers/internal/manual_rebuild.py
```diff
@@ -61,10 +61,12 @@ def parse_post_data(self, data):
...
 
+        init_args = [msg_id, advisory, data.get("container_images", []),]
+        if data.get("manual_image_remap", False):
+            init_args.append(data.get("manual_image_remap"))
+
         return klass(
-            msg_id,
-            advisory,
-            data.get("container_images", []),
+            *init_args,
```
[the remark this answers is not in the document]
FernandesMF: I see! I'll use the `metadata` field then

freshmaker/handlers/botas/botas_shipped_advisory.py
FernandesMF: I see! Thanks for the detailed explanation!

FernandesMF:
@qixiang thanks a lot for your input! I have reimplemented the nvr mapping (and reverted the previous changes that will not be needed). Could you please have another look?

Btw, I have one question. In the `_manual_nvr_remap`, I made it so that if we specify a manual remapping of an nvr that is not being rebuilt, the method will just ignore that and carry on. For example, if the nvrs that will be rebuilt are
```
original_nvr1 -> rebuilt_nvr1
original_nvr2 -> rebuilt_nvr2
original_nvr3 -> rebuilt_nvr3
```
and we make a manual override request 
```
'bundle_related_image_overrides': {'original_nvr3': 'some_nvr3', 'original_nvr4': 'some_nvr4'}
```
then after the remap is applied, the nvrs will be
```
original_nvr1 -> rebuilt_nvr1
original_nvr2 -> rebuilt_nvr2
original_nvr3 -> some_nvr3
```
and the override for `original_nvr4` will be ignored.

Is this logic okay, or should it work differently?

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -143,6 +143,11 @@ def _get_bundles_to_rebuild(self):
         log.debug("Getting NVR mapping of original images to rebuilt images")
         nvrs_mapping = self._create_original_to_rebuilt_nvrs_map()
 
+        if (isinstance(self.event, ManualBundleRebuildEvent) and 
+            self.event.requester_metadata_json.get("bundle_related_image_overrides", False)):
```
qixiang, line 147: Just check `self.event.requester_metadata_json.get("bundle_related_image_overrides", False)` is good enough.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -143,6 +143,11 @@ def _get_bundles_to_rebuild(self):
...
+            log.info("Performing manual NVR remapping")
+            self._manual_nvr_remap(nvrs_mapping)  #inplace substitutions to nvrs_mapping
```
qixiang, line 149: I think it's unnecessary to have a method to keep several lines with simple logic, but please add a comment to explain what exact the replacement is doing.

freshmaker/views.py
```diff
@@ -398,6 +398,19 @@ def _validate_rebuild_request(request):
...
+            'Bad Request',
+            'Manual image overriding allowed only when "container_images" is set explicitly'
+        )
+
+    if not isinstance(data.get('bundle_related_image_overrides', False), dict):
+        return json_error(
+            400, 'Bad Request', '"bundle_related_image_overrides" must be a dictionary'
+        )
+    
+
```
qixiang, line 413:
```suggestion
    
```

qixiang (commented):
> I made it so that if we specify a manual remapping of an nvr that is not being rebuilt, the method will just ignore that and carry on. For example, if the nvrs that will be rebuilt are

we should keep the mappings that doesn't exist in the `nvrs_mapping`, as mentioned before there is an important reason we want to add this feature: when a bundle was not built successfully in previous events, then it's possible the `original_nvr*` will not be present in the latest bundle, and we need to override that by `original_nvr* -> rebuilt_nvr*`.

freshmaker/handlers/botas/botas_shipped_advisory.py
FernandesMF: Okay! I'll "de-extract" that method and make an explanatory comment

freshmaker/views.py
FernandesMF: Github is not showing any change in your suggestion. Is there anything you wanted to discuss about the checks in `view.,py`?

qixiang: It was just a suggestion of removing the redundant blank line, but it seems github didn't render the suggestion correctly.

qixiang (commented):
The change looks good to me, please fix the tests and add new cases to cover the NVR replacement cases, and squash the commits before mark it as ready.

FernandesMF:
Hi @qixiang, I fixed the old tests and wrote a new unit test to check the manual remapping. Currently this new test is not passing. I think I found the reason, but I'll need some help to figure the solution.

In the test scenario, I'm using three fictional bundles; bundles 1 and 2 are passed in the `container_images` attribute of the event, and bundle 3 is passed (only) in the manual requests: https://github.com/redhat-exd-rebuilds/freshmaker/blob/08e62e8d3172663a19c53f420cb8ccc61b2af992/tests/handlers/botas/test_botas_shipped_advisory.py#L714-L722
The test is breaking because the information on bundle 3 is not present in the return by the end. The error message is:
``` python
FAILED tests/handlers/botas/test_botas_shipped_advisory.py::TestBotasShippedAdvisory::test_handle_bundle_rebuild_manual_nvr_override
AssertionError: 'get_build' does not contain all of (call('bundle3_nvr-1-1'),) in its call list, found [] instead
```

I traced the problem to this block of code, which builds the variable `bundle_nvr`: https://github.com/redhat-exd-rebuilds/freshmaker/blob/08e62e8d3172663a19c53f420cb8ccc61b2af992/freshmaker/handlers/botas/botas_shipped_advisory.py#L211-L228
  - in the `if` clause (which starts in line 213), the variable `bundle_nvr` only holds information about bundle 1 and 2:
    ``` python
    (Pdb) bundle_nvrs
    ['bundle1_nvr-1-1', 'bundle2_nvr-1-1']
    ```
  - however in the `else` clause (line 228), `bundle_nvrs` holds information about all the three bundles:
    ``` python
    (Pdb) self._get_impacted_bundles(original_digests_by_nvr.values())
    ['bundle3_nvr-1-1', 'bundle2_nvr-1-1', 'bundle1_nvr-1-1']
    ```
    and if I use this value, the new test passes.

Do you have some idea about how we could make `bundle_nvrs` become "aware" of the third bundle, which is not included in `container_images`, in the `if` clause?
Or should we make a new block after this one, to include the missing information?
I feel I don't know the context of this code, could you also explain a bit what cases this block was designed to handle?

Thanks a lot!

qixiang:
> The test is breaking because the information on bundle 3 is not present in the return by the end. The error message is:
> ```python
> FAILED tests/handlers/botas/test_botas_shipped_advisory.py::TestBotasShippedAdvisory::test_handle_bundle_rebuild_manual_nvr_override
> AssertionError: 'get_build' does not contain all of (call('bundle3_nvr-1-1'),) in its call list, found [] instead
> ```

the bundle nvrs to be rebuilt has been specified in the request explicitly by the `container_images`, so Freshmaker will only rebuild the specified `container_images` which is `['bundle1_nvr-1-1', 'bundle2_nvr-1-1']`, so it's expected the `bundle3_nvr-1-1` will not be rebuilt.

> * in the `if` clause (which starts in line 213), the variable `bundle_nvr` only holds information about bundle 1 and 2:
>   ```python
>   (Pdb) bundle_nvrs
>   ['bundle1_nvr-1-1', 'bundle2_nvr-1-1']
>   ```
> * however in the `else` clause (line 228), `bundle_nvrs` holds information about all the three bundles:
>   ```python
>   (Pdb) self._get_impacted_bundles(original_digests_by_nvr.values())
>   ['bundle3_nvr-1-1', 'bundle2_nvr-1-1', 'bundle1_nvr-1-1']
>   ```

when nvrs_mapping is something like this:
```
{
"nvr-1-123": "nvr-1-123.1687259260",
"nvr-2-234": "nvr-2-234.1687259286"
}
```
Then in the `if` clause, Freshmaker will only rebuild the specified bundles in request, which is `bundle1_nvr-1-1` and `bundle2_nvr-1-1`, the related images in these two bundles will be updated by:
```
nvr-1-123 -> nvr-1-123.1687259260,
nvr-2-234 -> nvr-2-234.1687259286
``` 
of course only if `nvr-1-123` and `nvr-2-234` are present in the two bundles' related images list.

In the `else` clause, Freshmaker will use `nvr-1-123` and `nvr-2-234` to find out which bundles are having these two images present in the related images list.

So the `if-else` is covering two different cases, and in your test you only need to test the first case (`if` clause).

qixiang:
@FernandesMF is this ready for the next round of review? If yes, could you squash the commits and rebase this PR?

qixiang:
> @qixiang I believe it is not quite ready yet. Since I might find a problem during the testing in the dev environment, it could be that I might to introduce additional changes. Do you agree?

No problem, just want to check if you want some reviews/approvals at this moment.

FernandesMF:
Alright, thank you for your concern! 😄

qixiang (approved):
LGTM

PR #226, opened 2023-06-26
Log pyxis traceid

FernandesMF:
I sketched how we could log the trace_id field in pyxis responses, in case they are present. I am not sure if the way I am accessing them is correct, especially in `pyxis_gql.py`. Do you think that would work?


freshmaker/pyxis_gql.py
```diff
@@ -174,7 +176,11 @@ def find_repositories(self, published=None, release_categories=None, auto_rebuil
             result = self.query(query_dsl)
             error = result["find_repositories"]["error"]
             if error is not None:
-                raise PyxisGQLRequestError(str(error))
+                trace_msg = ""
```
qixiang, line 179: `trace_id` is not in Graphql returned data, it's in response header:
https://gql.readthedocs.io/en/v3.3.0/usage/headers.html

Instead of logging this in all methods, you can do it in `query`: https://github.com/redhat-exd-rebuilds/freshmaker/blob/9788db751ee6c94c6b59c01464622c53c4bdc494/freshmaker/pyxis_gql.py#L65 .

FernandesMF: I see! I'll correct that. I also liked the idea of logging in the `query` method, instead of in all methods, thanks a lot!

FernandesMF:
Hi @qixiang! I have reimplemented the logging in `pyxis_gql.py`, following your suggestions.

Now it gets the trace id from the headers of the response, and does the logging in the `query` method. I also refactored the error checks along the code to the query method as well; that block of code was repeating several times and was running immediately after the call to `query`, so I thought it could be incorporated by the method.

I also replicated the logic the `pyxis_gql_async.py`.

Could you have another look, please? Thanks!

FernandesMF:
After the correction in our main branch history, there were many conflicts to be fixed between this branch and the new main branch.
I made a new branch in my fork and applied the changes there. I opened a new PR with that branch (#227), and will drop this one.

PR #227, opened 2023-07-04
feat: Log trace_id in case of Pyxis request error

FernandesMF:
In case of an error with a request to Pyxis, it is useful to log the `trace_id` property for later debugging. This property is stored in the response headers from pyxis. This commit fetches its value and stores it in the logging apparatus.

This commit includes the following changes:
- fetching the `trace_id` relative to the Pyxis query in case of an error, in the three Pyxis modules (`pyxis.py`, `pyxis_gql.py` and `pyxis_gql_async.py)
- new error classes with more structure in the gql modules (`pyxis_gql.py` and `pyxis_gql_async.py)
- refactoring of the error checks in the gql modules, which are now centralized to the query methods; this should avoid unnecessary duplication
- unit tests for the fetching and storing of trace id in the three Pyxis modules
- a few linting and formatting adjustments to the Pyxis modules and its tests

JIRA: CWFHEALTH-2023

This PR is a continuation of #226.

### DIFF
--- a/freshmaker/pyxis.py
+++ b/freshmaker/pyxis.py
@@ -1,10 +1,11 @@
-import dogpile.cache
-import requests
 import urllib
 from datetime import datetime
-from requests_kerberos import HTTPKerberosAuth, OPTIONAL
 
-from freshmaker import log, conf
+import dogpile.cache
+import requests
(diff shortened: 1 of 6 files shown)


qixiang (approved):
LGTM

PR #228, opened 2023-07-07
Improve logging

FernandesMF:
# Improve logging

There are a few issues with the current logging, let's enhance it!

### Issue: lack of debug-level logs
The cause was that in the `logger.py` module, the log level was being set from an environment variable called "DEBUG" that was not being set in the deployment.

Since there is a setting for the config level in the configuration file, I changed it to get the setting from there.

After doing that, I made a quick deployment to dev and I saw some debug-level lines in the logs.

### Issue: too many info-level logs from ~~pyxis~~ gql, including its data
The solution I found was to include change the gql module logging level in the configs. That change is in the internal repo.

### Issue: duplicated log lines in console
The cause behind this issue was that there were two handlers to the console: one in the root logger, which was being configured with a stream handler to `stderr` by the `basicConfig()` commands, and the second in the child "freshmaker" logger, pointing to `stdout`.

The solution was to  explicitly set the console handler in the root logger and not set it in the freshmaker logger. That way we have only one logger, and also  prevent calls from basicConfig and logging.debug/warning/... to create a standard handler in the root logger.

Also, I removed the 'journal' and 'file' logging backends, which were not really used in practice. We decided to remove
those options and the related code to simplify the project.

JIRA: CWFHEALTH-2103

### DIFF
--- a/conf/config.py
+++ b/conf/config.py
@@ -33,12 +33,6 @@ class BaseConfiguration(object):
 
     SYSTEM = 'koji'
 
-    # Available backends are: console, file, journal.
-    LOG_BACKEND = 'journal'
-
-    # Path to log file when LOG_BACKEND is set to "file".
-    LOG_FILE = 'freshmaker.log'
-
     # Available log levels are: debug, info, warn, error.
     LOG_LEVEL = 'info'
 
@@ -210,7 +204,6 @@ class BaseConfiguration(object):
 
 class DevConfiguration(BaseConfiguration):
     DEBUG = True
-    LOG_BACKEND = 'console'
     LOG_LEVEL = 'debug'
 
     MESSAGING_TOPIC_PREFIX = ['org.fedoraproject.dev', 'org.fedoraproject.stg']
@@ -236,7 +229,6 @@ class DevConfiguration(BaseConfiguration):
 
 
 class TestConfiguration(BaseConfiguration):
-    LOG_BACKEND = 'console'
     LOG_LEVEL = 'debug'
     DEBUG = True
 
--- a/freshmaker/logger.py
+++ b/freshmaker/logger.py
@@ -56,16 +56,15 @@ levels = {
 
 DEV_ENV = os.environ.get("FRESHMAKER_TESTING_ENV") == "1"
 LOG_DIR = "." if DEV_ENV else "/var/log/freshmaker"
-LOG_LEVEL = "DEBUG" if os.environ.get("DEBUG") else "INFO"
 
 
-def setup_logger():
+def setup_logger(conf):
     """
     Set up and configure 'freshmaker' logger.
     """
     logger = logging.getLogger("freshmaker")
     logger.propagate = False
-    logger.setLevel(LOG_LEVEL)
+    logger.setLevel(conf.log_level)
     source = "freshmaker"
     custom_attr = {
         "environment": lambda: os.environ.get("FRESHMAKER_ENV"),
@@ -102,20 +101,9 @@ def setup_logger():
         backupCount=1,
         mode="a",
     )
-    file_handler.setLevel(LOG_LEVEL)
+    file_handler.setLevel(conf.log_level)
     file_handler.setFormatter(file_formatter)
     logger.addHandler(file_handler)
-    if os.environ.get("CONSOLE_LOGS", "").lower() != "false":
-        console_handler = logging.StreamHandler(stream=sys.stdout)
-        console_handler.setLevel(LOG_LEVEL)
-        console_handler.setFormatter(
-            logging.Formatter(
-                "[%(asctime)s.%(msecs)d] [%(processName)s - %(threadName)s] "
-                "[%(levelname)s] %(message)s",
-                "%Y-%m-%dT%H:%M:%S",
-            )
-        )
-        logger.addHandler(console_handler)
 
 
 def str_to_log_level(level):
@@ -131,32 +119,21 @@ def str_to_log_level(level):
     return levels[level]
 
 
-def supported_log_backends():
-    return ("console", "journal", "file")
-
-
 def init_logging(conf):
     """
     Initializes logging according to configuration file.
     """
-    log_format = '%(levelname)s - %(message)s'
-    log_backend = conf.log_backend
+    log = logging.getLogger()
 
-    if not log_backend or len(log_backend) == 0 or log_backend == "console":
-        logging.basicConfig(level=conf.log_level, format=log_format)
-        log = logging.getLogger()
-        log.setLevel(conf.log_level)
-    elif log_backend == "journal":
-        logging.basicConfig(level=conf.log_level, format=log_format)
-        try:
-            from systemd import journal
-        except ImportError:
-            raise ValueError("systemd.journal module is not installed")
-
-        log = logging.getLogger()
-        log.propagate = False
-        log.addHandler(journal.JournalHandler())
-    else:
-        logging.basicConfig(filename=conf.log_file, level=conf.log_level,
-                            format=log_format)
-        log = logging.getLogger()
+    formatter = logging.Formatter(
+        "[%(asctime)s.%(msecs)d] [%(processName)s - %(threadName)s] [%(levelname)s] %(message)s",
+        "%Y-%m-%dT%H:%M:%S",
+    )
+    console_handler = logging.StreamHandler(stream=sys.stdout)
+    console_handler.setLevel(conf.log_level)
+    console_handler.setFormatter(formatter)
+    # explicitly add a handler to root logger to avoid duplicated log lines:
+    # if the root logger does not have a handler, a call to logging.basicConfig() or
+    # logging.debug/warning/error/... will create and attach one, and if we set another handler
+    # to the console, we will have duplicated log lines
+    log.addHandler(console_handler)
--- a/freshmaker/__init__.py
+++ b/freshmaker/__init__.py
@@ -49,7 +49,7 @@ conf = init_config(app)
 db = SQLAlchemy(app)  # type: Any
 
 init_logging(conf)
-setup_logger()
+setup_logger(conf)
 log = getLogger(__name__)
 
 login_manager = LoginManager()
--- a/freshmaker/config.py
+++ b/freshmaker/config.py
@@ -139,14 +139,6 @@ class Config(object):
             'type': bool,
             'default': False,
             'desc': 'Debug mode'},
-        'log_backend': {
-            'type': str,
-            'default': None,
-            'desc': 'Log backend'},
-        'log_file': {
-            'type': str,
-            'default': '',
-            'desc': 'Path to log file'},
         'log_level': {
             'type': str,
             'default': 0,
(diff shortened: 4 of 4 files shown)


freshmaker/pyxis.py
```diff
@@ -49,6 +52,39 @@ def __init__(self, server_url):
...
+        """ Creates the logging structure of the Pyxis module.
+        
+        We might want a logging level in this module that is different (and probaly higher) than the
+        main-level. Therefore we create a logger that has a custom log level and; otherwise, it will
+        just forward the logs to the main-level logger to handle.
+
+        Also, because of the large amount of data that can be fetched from Pyxis, we will log it
+        into a separate file.
+        """
+        self.logger = logging.getLogger("freshmaker.pyxis")
```
qixiang, line 67: The pretty large amount of logs come from `gql` module, not `freshmaker.pyxis`, and these logs are in level `info`, not `debug`, we can change the log level of `gql` module to warning: `internal-repo-url/freshmaker-playbooks/-/blob/master/roles/freshmaker-openshift/templates/fedmsg_logging.py#L43`

We don't have much logs from `freshmaker.pyxis`, with `gql` module logging level changed to `warning`, we probably need to add some logging in `freshmaker.pyxis` for troubleshooting purpose, but we can do that later.

qixiang (commented):
> Since there is a setting for the config level in the configuration file, I changed it to get the setting from there.

Does it result in circular import?

freshmaker/pyxis.py
FernandesMF: Oh, I see... I really misunderstood this issue 😅 
I'll correct it and revert these changes in the pyxis modules.

And just for the curiosity, the circular import didn't happen. That's because I passed the configuration dict as a parameter in the calls to the logger module that are made in `__init__.py`.

In an earlier attempt, I tried to import the configuration dict directly in the logger module, and that resulted indeed a circular error!

freshmaker/logger.py
```diff
@@ -102,12 +101,12 @@ def setup_logger():
         backupCount=1,
         mode="a",
     )
-    file_handler.setLevel(LOG_LEVEL)
+    file_handler.setLevel(conf.log_level)
     file_handler.setFormatter(file_formatter)
     logger.addHandler(file_handler)
     if os.environ.get("CONSOLE_LOGS", "").lower() != "false":
```
qixiang, line 107: This adds another console stream handler:
```
logger.addHandler(console_handler)
```
that's why we see duplicate log lines (with different formats).

When the root logger is configured by:
```
logging.basicConfig(level=conf.log_level, format=log_format)
```
in `init_logging`, a console stream handler is enabled implicitly, so we have two console stream handlers enabled at the end.

qixiang (approved):
This change looks good to me, I also commented why we see duplicated log lines, but I'm fine with fixing that in another PR, so I'll just approve this PR and leave it to you to decide whether fix the duplicate logs in this PR or a different PR.

freshmaker/logger.py
FernandesMF: I see! Thanks for clarifying. I think I'm almost getting this solved, so I guess I'll try a bit more on this PR!

Indeed, the functions `init_logging` and `setup_logger` were instantiating two different loggers with their own handlers:
``` python
(init_logging)
<RootLogger root (INFO)>
[<StreamHandler <stderr> (NOTSET)>, <JournalHandler (NOTSET)>]

(setup_logger)
<Logger freshmaker (INFO)>
[<RotatingFileHandler /var/log/freshmaker/freshmaker.splunk.log (INFO)>, <StreamHandler <stdout> (INFO)>]
```
The `getLogger()` commands in `init_logging` were not referencing to the module/app name, and were therefore referencing to the root logger. I changed them to `logging.getLogger("freshmaker")`:
https://github.com/redhat-exd-rebuilds/freshmaker/blob/344423d7d76356c4e69ca16ab224f87188ba5266/freshmaker/logger.py#L146-L161
and now we only have one logger:
``` python
(init_logging)
<Logger freshmaker (INFO)>
[<JournalHandler (NOTSET)>]

(setup_logger)
<Logger freshmaker (INFO)>
[<JournalHandler (NOTSET)>, <RotatingFileHandler /var/log/freshmaker/freshmaker.splunk.log (INFO)>, <StreamHandler <stdout> (INFO)>]
```

The handler to stderr, which was previously attached to the root logger, is gone. But I think there wasn't a second console handler to stdout. (At least not for now, as the current deployments are using `log_backend=journal` if I'm not mistaken). Since the duplicated lines were not just for errors, I thought there must be something else to this issue.

I'm under the impression that as long as we have two handlers, one for the console and another for the log file, we will have those duplicated lines when we use `oc logs` (although they have different formatting). The solution that I could think for that was to set console logging conditionally to some configuration value, and in fact the logic for this was already in the code! 

I made an adjustment to the condition for setting console logging, and I introduced a new variable to the configuration files to control it. I set its default value to False, that way we can avoid logging to console:
https://github.com/redhat-exd-rebuilds/freshmaker/blob/344423d7d76356c4e69ca16ab224f87188ba5266/freshmaker/logger.py#L107-L108

(I believe the previous condition was not working as intended; if the "CONSOLE_LOGS" variable was not set, it would still make the console logging, since `("" != "false") = True` )

What do you think about these changes?

Thanks a lot again, @qixiang! You are always very helpful!

qixiang: The logging module is designed in some way against Python's zen, so it easily causes some confusions. A very important thing to understand this issue is these lines from https://docs.python.org/3/library/logging.html#logging.basicConfig
```
logging.basicConfig(**kwargs)
Does basic configuration for the logging system by creating a StreamHandler with
a default Formatter and adding it to the root logger. The functions debug(), info(),
warning(), error() and critical() will call basicConfig() automatically if no
handlers are defined for the root logger.

This function does nothing if the root logger already has handlers configured,
unless the keyword argument force is set to True.
```
So whenever you call `logging.basicConfig` (or call the `logging.debug` functions) it will **create a StreamHandler and add it to the root logger** unless the root handler is already has some handlers configured.

In our code:
```
logging.basicConfig(level=conf.log_level, format=log_format)
```
this configured the root logger with a stream handler attached, then if we have some code like:
```
import logging
logging.basicConfig(level=logging.ERROR, format="%(levelname)s: %(message)s")
freshmaker_log = logging.getLogger("freshmaker")
freshmaker_log.setLevel(logging.INFO)
freshmaker_log.info("test line")
```
the output is:
```
INFO: test line
```
The reason is `freshmaker_log` doesn't have any handlers, but the root logger has a stream handler (configured by `logging.basicConfig`, so the line is printed by the root logger.

In above example, another thing which can also be confusing is the root logger has log level set to `ERROR`, we called `freshmaker_log.info()`, but root logger can still print the line, this is because the log level is only checked by the `freshmaker_log` which the message was given to, and all other loggers doesn't check log level at all.

conf/config.py
```diff
@@ -42,6 +42,8 @@ class BaseConfiguration(object):
     # Available log levels are: debug, info, warn, error.
     LOG_LEVEL = 'info'
 
+    CONSOLE_LOG = False
```
qixiang, line 45: This is the config file used in testing or developer mode (https://github.com/redhat-exd-rebuilds/freshmaker/blob/807bf9e3b1bbcd79ce00be615e3a6e9c742e1e17/freshmaker/config.py#L67)

You have to update this in https://github.com/redhat-exd-rebuilds/freshmaker/blob/807bf9e3b1bbcd79ce00be615e3a6e9c742e1e17/freshmaker/config.py#L135

freshmaker/logger.py
FernandesMF: Ah, I see! I was misunderstanding the issue again... but now I got it! Thanks for you patience, @qixiang!

I reverted this last attempt I made for not logging to the console, and instead implemented a function to remove the console handler of the root logger. I guess that would do the trick, right?

conf/config.py
FernandesMF: After your last explanation, these changes became unnecessary and I just reverted them.

freshmaker/logger.py
```diff
@@ -102,22 +120,22 @@ def setup_logger():
...
-        console_handler.setLevel(LOG_LEVEL)
+        console_handler.setLevel(conf.log_level)
         console_handler.setFormatter(
             logging.Formatter(
                 "[%(asctime)s.%(msecs)d] [%(processName)s - %(threadName)s] "
                 "[%(levelname)s] %(message)s",
                 "%Y-%m-%dT%H:%M:%S",
             )
         )
+        remove_root_default_handler()  # avoid duplicated log lines due to two handlers
```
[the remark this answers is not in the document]
FernandesMF: Good point! I moved the console handler to the root logger:

https://github.com/redhat-exd-rebuilds/freshmaker/blob/b3096339a89c1a607230bd7fbe4950db02a6e073/freshmaker/logger.py#L126-L161

I left the console handler part out of the if clauses. I tried to keep the same functionality as we had before, but it does look a bit weird to me. In case log_backend=journal, in the older version we would still have the stream handler  in the root logger because of a `basicConfig` call, so I kept that for now. Do you think that is okay? Or should I move it to the first if block?

Also, I made the last case more explicit and required `log_backend=file`, and raised an exception for unrecognized clauses. In the older version, if for example someone mistyped `log=console="cosnole"`, they would end up with a file backend and no error, which i think is a bit weird...

freshmaker/logger.py
```diff
@@ -139,24 +127,35 @@ def init_logging(conf):
...
             raise ValueError("systemd.journal module is not installed")
-
-        log = logging.getLogger()
-        log.propagate = False
         log.addHandler(journal.JournalHandler())
+
+    elif log_backend == "file":
+        file_handler = logging.FileHandler(conf.log_file)
+        file_handler.setFormatter(formatter)
+        log.removeHandler(console_handler)
```
qixiang, line 157: why not moving `log.addHandler(console_handler)` to the "console" and "journal" branches?

I don't see any requirements about writing the logs to journal log, probably we can remove it in future improvements.

qixiang (approved):
LGTM, just one minor comment.

freshmaker/logger.py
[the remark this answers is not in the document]
qixiang: > I left the console handler part out of the if clauses. I tried to keep the same functionality as we had before, but it does look a bit weird to me. In case log_backend=journal, in the older version we would still have the stream handler in the root logger because of a basicConfig call, so I kept that for now. Do you think that is okay? Or should I move it to the first if block?

ok, I missed this comment when approve it and posted my comment, I agree it looks a little weird and +1 to move it to the "console" branch. We also need it in "journal" branch, because freshmaker is currently deployed with log backend "journal", and we do want to have logs in console, a more ideal solution would be changing the deployment to use "console" log backend and remove the "journal" support, I have no idea why it was there and use it as default in our deployment.

freshmaker/logger.py
```diff
@@ -139,24 +127,35 @@ def init_logging(conf):
     """
     Initializes logging according to configuration file.
     """
-    log_format = '%(levelname)s - %(message)s'
+    log = logging.getLogger()
+    log.setLevel(conf.log_level)
```
qixiang, line 131: I just realized we should not change the default log level of root logger, instead of that, the `conf.log_level` should be applied to `freshmaker` logger.

FernandesMF: Done!

freshmaker/logger.py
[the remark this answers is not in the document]
FernandesMF: I removed the 'journal' and 'file' log backends and their related configurations, that will make the code simpler! I also updated the commit message to reflect that.

qixiang:
Good job, LGTM.

PR #229, opened 2023-07-11
Logging invalid image NVR for troubleshooting

qixiang:
Pyxis may contains some invalid image metadata, for example, an image missing rpm_manifest, when such image data is returned by Pyxis, Freshmaker rebuild event will raise an exception, but it's hard to figure out which image caused the exception, logging the image NVR can help us on finding out the problematic image and fix it in metadata.

JIRA: CWFHEALTH-2104

### DIFF
--- a/freshmaker/image.py
+++ b/freshmaker/image.py
@@ -798,8 +798,17 @@ class PyxisAPI(object):
             ]
 
         def _image_has_rpm(image, rpms):
-            # rpms is a list of rpm names
-            rpm_names = [x["name"] for x in image["rpm_manifest"][0]["rpms"]]
+            try:
+                # rpms is a list of rpm names
+                rpm_names = [x["name"] for x in image["rpm_manifest"][0]["rpms"]]
+            except Exception:
+                # When this is called, the image should have rpm_manifest, but Pyxis
+                # may have some invalid image data which is missing rpm_manifest; when
+                # such images are returned by Pyxis, it will trigger an exception; we need
+                # to log the image NVR for troubleshooting purposes.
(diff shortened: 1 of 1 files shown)


freshmaker/image.py
```diff
@@ -798,8 +798,17 @@ def get_images_by_nvrs(self, nvrs, published=True, content_sets=None,
...
-            # rpms is a list of rpm names
-            rpm_names = [x["name"] for x in image["rpm_manifest"][0]["rpms"]]
+            try:
+                # rpms is a list of rpm names
+                rpm_names = [x["name"] for x in image["rpm_manifest"][0]["rpms"]]
+            except Exception:
+                # When this is called, the image should always has rpm_manifest, but Pyxis
+                # may has some invalid image data which is missing rpm_manifest, when
+                # such images are returned by Pyxis, it will trigger an exception, we need
+                # to logging the image NVR for troubleshooting purpose.
```
FernandesMF, line 808: I just have a silly linguistic suggestion for the explanatory comment:
```
# When this is called, the image should have rpm_manifest, but Pyxis
# may have some invalid image data which is missing rpm_manifest; when
# such images are returned by Pyxis, it will trigger an exception; we need
# to log the image NVR for troubleshooting purposes.
```
but feel free to ignore it if I didn't understand what you meant!

FernandesMF (approved):
Looks good to me!
I just had a minor linguistic suggestion, but feel free to ignore it in case it is not accurate.

freshmaker/image.py
qixiang: It's much better, I've applied the suggestion, thank you!

PR #230, opened 2023-07-20
Reorder the Pyxis query filters

qixiang:
Reordering the filters from most specific to least specific can improve the performance of queries in Pyxis.


FernandesMF (approved):
Makes all the sense, and looks good to me!

PR #231, opened 2023-07-20
Update event state when the last build failed due to ODCS compose error

qixiang:
When the last build in event failed due to ODCS compose error, the build state will be changed to FAILED, but Freshmaker doesn't update the event to COMPLETE state.

JIRA: CWFHEALTH-2162

### DIFF
--- a/freshmaker/handlers/__init__.py
+++ b/freshmaker/handlers/__init__.py
@@ -434,6 +434,35 @@ class BaseHandler(object):
                        criteria, artifact_type.name.lower())
         return False
 
+    def _mark_event_complete_when_all_builds_done(self, db_event):
+        """Mark Event COMPLETE
+
+        When all builds in an event have been done (either in FAILED or DONE
+        state), the event should be marked as COMPLETE accordingly. If not all
+        builds are finished, then the state is not changed
+
+        :param Event db_event: instance of Event
+        """
+        num_failed = 0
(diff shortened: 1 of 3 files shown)


freshmaker/handlers/__init__.py
```diff
@@ -434,6 +434,35 @@ def allow_build(self, artifact_type, **criteria):
                        criteria, artifact_type.name.lower())
         return False
 
+    def _mark_event_complete_when_all_builds_done(self, db_event):
+        """Mark Event COMPLETE
+
+        When all builds in an event have been done (either in FAILED or DONE
+        state), the event should be marked as COMPLETE accordingly. If not all
+        finish, nothing change to the state.
```
FernandesMF, line 442: I just have a minor suggestion to rephrase that last sentence in the docstring:
"If not all builds are finished, then the state is not changed"?

FernandesMF (approved):
I just have a minor linguistic suggestion, looks good to me otherwise!

freshmaker/handlers/__init__.py
qixiang: Addressed, thanks.

qixiang:
The doc build job error is fixed in https://github.com/redhat-exd-rebuilds/freshmaker/pull/233 .

The comment is addressed, I'll merge this.

PR #232, opened 2023-07-20
Automatic rebuild of old operators/operands images

FernandesMF:
Sometimes, it can happen that a bundle loses track of the latest image of an operator/operand, in case they are both rebuilt but the bundle rebuild fails to ship. In this case, new rebuilds of those operator/operand will no longer trigger a rebuild of the bundle.

For example, imagine we have a bundle 'bundle-2-234' and a related image 'foo-1-123', and the following happens:
1. the image gets rebuilt to 'foo-1-123.234' and the bundle also happens a rebuild triggered, but it fails to ship
2. then the image gets rebuild again to 'foo-123-456'
3. this new rebuild of 'foo-1-123' will not trigger a rebuild of the        bundle, as it is not related to 'foo-123-456'

The manual overriding feature can be used to circumvent this issue, but it is better if we can make an automatic solution.

This commit enables FM to automatically search, during bundle rebuilds, for previous rebuilds of those related images and replace them with the latest images.

JIRA: CWFHEALTH-2141

### DIFF
--- a/freshmaker/config.py
+++ b/freshmaker/config.py
@@ -362,6 +362,13 @@ class Config(object):
             'default': '',
             'desc': 'A suffix to add to the rebuilt_nvr release in addition to the timestamp.',
         },
+        'bundle_include_previous_rebuilds': {
+            'type': bool,
+            'default': True,
+            'desc': 'When True, enables an automatic search in bundle rebuilds for previous'
+                    'Freshmaker builds of the current operator/operand images, and replace them'
+                    'in the bundle.'
+        },
         'container_release_categories': {
             'type': tuple,
             'default': ("Generally Available", "Tech Preview", "Beta",),
--- a/freshmaker/handlers/botas/botas_shipped_advisory.py
+++ b/freshmaker/handlers/botas/botas_shipped_advisory.py
@@ -607,7 +607,6 @@ class HandleBotasAdvisory(ContainerBuildHandler):
             for build in product_info["builds"]:
                 # Each build is a one key/value pair, and key is the build NVR
                 build_nvr = next(iter(build))
-
                 log.debug("Getting published original image of %s", build_nvr)
                 # Search for the first build that triggered the chain of rebuilds
                 # for every shipped NVR to get original NVR from it
@@ -626,7 +625,46 @@ class HandleBotasAdvisory(ContainerBuildHandler):
                         and block_build_nvr["version"] == parsed_build_nvr["version"]  # noqa: W503
                     ):
                         nvrs_mapping[block_build] = build_nvr
+
+        if conf.bundle_include_previous_rebuilds:
+            nvrs_mapping = self._completion_with_previous_rebuilds(nvrs_mapping)
+            log.info(f"NVRs mapping udpated with previous rebuilds: {nvrs_mapping}")
+
         return nvrs_mapping
+
+    def _completion_with_previous_rebuilds(self, nvrs_mapping: dict[str, str]) -> dict[str, str]:
+        """Generates an udpated NVR mapping with the chain of previous freshmaker rebuilds
+
+        This function searches the previous rebuilds of the NVRs in the current NVR mapping
+        and adds them to the mapping, in case they are previous freshmaker rebuilds.
+
+        This is used to trigger rebuilds of bundles that lost the "link" to a latest version
+        of an image (due to problems with a former rebuild process, for example) in an automatic
+        fashion, without requiring manual overriding.
+
+        :param nvrs_maping: the current NVR mapping
+        :type nvrs_mapping: dict
+        :return: udpated NVR mapping including the previous rebuilds
+        :rtype: dict
+        """
+        nvr_mappings_to_add = {}
+        for original_nvr, latest_nvr in nvrs_mapping.items():
+            # The nvrs_mapping is construced by checking database, so the latest_rebuild always
+            # exists and has only one record.
+            latest_rebuild = ArtifactBuild.query.filter_by(rebuilt_nvr=latest_nvr).first()
+
+            # Check whether the original NVR is also a freshmaker rebuild
+            former_build = ArtifactBuild.query.filter_by(
+                rebuilt_nvr=latest_rebuild.original_nvr
+            ).first()
+            while former_build:
+                nvr_mappings_to_add[former_build.rebuilt_nvr] = latest_nvr
+                log.info(f"Found former rebuild '{former_build.rebuilt_nvr}' of '{latest_nvr}'")
+                former_build = ArtifactBuild.query.filter_by(
+                    rebuilt_nvr=former_build.original_nvr
+                ).first()
+
+        return {**nvrs_mapping, **nvr_mappings_to_add}
 
     def _prepare_builds(self, to_rebuild_bundles):
         """
--- a/tests/handlers/botas/test_botas_shipped_advisory.py
+++ b/tests/handlers/botas/test_botas_shipped_advisory.py
@@ -29,7 +29,9 @@ from freshmaker import db, conf
 from freshmaker.events import (
     BotasErrataShippedEvent,
     ManualRebuildWithAdvisoryEvent,
-    TestingEvent, ManualBundleRebuildEvent)
+    TestingEvent,
+    ManualBundleRebuildEvent,
+)
 from freshmaker.handlers.botas import HandleBotasAdvisory
 from freshmaker.errata import ErrataAdvisory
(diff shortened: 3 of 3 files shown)


freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -626,6 +626,37 @@ def _create_original_to_rebuilt_nvrs_map(self):
                         and block_build_nvr["version"] == parsed_build_nvr["version"]  # noqa: W503
                     ):
                         nvrs_mapping[block_build] = build_nvr
+
+        if self.event.get("autocomplete_field", False): #TODO implement "autocomplete" field of event
```
qixiang, line 630: If this is per event option, we will not be able to enable/disable this for the events triggered by advisories, and can only use this feature while triggering the rebuild by manual. I'd prefer to have a configuration option to enable/disable this (enabled by default), which can be toggle on/off globally.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -626,6 +626,37 @@ def _create_original_to_rebuilt_nvrs_map(self):
...
+                parent_nvr = None
+                stmt = select(ArtifactBuild).filter_by(rebuilt_nvr=child_nvr)
+                query_result = db.session.execute(stmt).first()
+                if query_result:
+                    # query_result is a tuple with the artifact build object on its first position
+                    parent_nvr = query_result[0].original_nvr
+                    nvr_mappings_to_add.update({parent_nvr: nvrs_mapping[nvr]})
+                    child_nvr = parent_nvr
+        
+        nvrs_mapping.update(nvr_mappings_to_add)
```
qixiang, line 659: Please also add logging (about which NVRs have been added) to help with troubleshooting.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -626,6 +626,37 @@ def _create_original_to_rebuilt_nvrs_map(self):
                         and block_build_nvr["version"] == parsed_build_nvr["version"]  # noqa: W503
                     ):
                         nvrs_mapping[block_build] = build_nvr
+
+        if self.event.get("autocomplete_field", False): #TODO implement "autocomplete" field of event
+            self._add_parent_nvrs(nvrs_mapping)
```
qixiang, line 631: `_add_parent_nvrs` returns a dict, but you didn't use it, the `nvrs_mapping` is actually updated implicitly which is not good.

freshmaker/handlers/botas/botas_shipped_advisory.py
FernandesMF: I see! I'll create a parameter in the configurations and fetch it in the `conf` dictionary then.

freshmaker/handlers/botas/botas_shipped_advisory.py
FernandesMF: Great point! I'll return a dictionary with the updated items and store it explicitly in `nvrs_mapping`, in order to avoid the implicit udpate.

freshmaker/handlers/botas/botas_shipped_advisory.py
FernandesMF: I created two log calls, one in the `while` loop in `_add_previous_rebuilds` and another in the `if` clause in `_create_original_to_rebuilt_nvrs_map`. Please let me know if you think that is excessive, if there is information missing, or anything that could improve the logging!

conf/config.py
```diff
@@ -207,6 +207,9 @@ class BaseConfiguration(object):
     # repositories that should be searched for unpublished images, specifically because of EUS base images
     UNPUBLISHED_EXCEPTIONS: list[dict[str, str]] = []
 
+    # option for enabling the automatic trigger of previous versions of operators
+    PREVIOUS_REBUILDS_AUTOTRIGGER = True
```
qixiang, line 211: How about rename this to `BUNDLE_INCLUDE_PREVIOUS_REBUILDS`?

freshmaker/config.py
```diff
@@ -370,6 +370,12 @@ class Config(object):
             'default': '',
             'desc': 'A suffix to add to the rebuilt_nvr release in addition to the timestamp.',
         },
+        'previous_rebuilds_autotrigger': {
+            'type': bool,
+            'default': False,
```
qixiang, line 375: I'd prefer to enable this by default, and we will not need to update the configurations (`conf/config.py` for locally, or playbooks for our deployments).

freshmaker/config.py
```diff
@@ -370,6 +370,12 @@ class Config(object):
...
+            'desc': 'When True, enables an automatic search for previous Freshmaker rebuilds'
+                    'of the current operators and triggers a new rebuild for each of them.'
```
qixiang, line 377: Better to mention this is something for bundle rebuild.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -649,13 +655,13 @@ def _add_parent_nvrs(nvrs_mapping: dict[str, str]) -> dict[str, str]:
...
-        
-        nvrs_mapping.update(nvr_mappings_to_add)
-        return nvrs_mapping
+            former_build = ArtifactBuild.query.filter_by(rebuilt_nvr=latest_rebuild.original_nvr).first()
+            while former_build:
+                nvr_mappings_to_add[former_build.rebuilt_nvr] = latest_nvr
+                log.info(f"Found former rebuild '{former_build.rebuilt_nvr}' of '{latest_nvr}'")
+                former_build = ArtifactBuild.query.filter_by(rebuilt_nvr=former_build.original_nvr).first()
+
+        return{**nvrs_mapping, **nvr_mappings_to_add}
```
qixiang, line 664: I'd suggest:
1. If you want to update the `nvr_mapping` directly, then just update this function to `_add_previous_rebuilds(nvrs_mapping: dict[str, str]) -> None`, and have:
```
while former_build:
    nvr_mapping[former_build.rebuilt_nvr] = latest_nvr
```
2. Or do it in more safe way (though seems not so necessary in this simple function):
```
nvrs_mapping = copy.deepcopy(nvrs_mapping)
...
while former_build:
    nvr_mapping[former_build.rebuilt_nvr] = latest_nvr
    ...
return nvrs_mapping
```

qixiang (commented):
It looks good to me, just minor suggestions.

conf/config.py
FernandesMF: That's a much better name! I'll update it throughout the files.

freshmaker/handlers/botas/botas_shipped_advisory.py
FernandesMF: I guess I feel more comfortable with explicitly setting nvrs_mapping to a new value, instead of changing it implicitly.
I changed the name of the new function to `_completion_with_previous_rebuilds`, as I thought the latter name was not making that much sense with that way of using. Now we have:
```python
nvrs_mapping = self._completion_with_previous_rebuilds(nvrs_mapping)
```

What do you think?

freshmaker/config.py
```diff
@@ -370,6 +370,13 @@ class Config(object):
             'default': '',
             'desc': 'A suffix to add to the rebuilt_nvr release in addition to the timestamp.',
         },
+        'bundle_include_previous_rebuilds': {
+            'type': bool,
+            'default': True,
+            'desc': 'When True, enables an automatic search in bundle rebuilds for previous'
+                    'Freshmaker builds of the current operators, and triggers a new rebuild'
+                    'for each of them.'
```
qixiang, line 378:
```suggestion
                    'Freshmaker builds of the current operator/operand images, and replace'
                    'them in bundle.'
```

freshmaker/handlers/botas/botas_shipped_advisory.py
qixiang: Yes, it looks good to me.

freshmaker/handlers/botas/botas_shipped_advisory.py
```diff
@@ -26,6 +26,7 @@
 import koji
 from kobo.rpmlib import parse_nvr
 import semver
+from sqlalchemy import select
```
qixiang, line 29: We don't need it anymore.

qixiang (commented):
Just minor comments.

traytorous:
Hello

qixiang (approved):
LGTM

ElenaKarolinaSemanova (approved):
Nothing to add, LGTM

PR #233, opened 2023-07-21
Fix doc build job

qixiang:
The "tox-github-action" has been changed to use `main` branch as default branch and `master` branch is not available now.


FernandesMF (approved):
Looks good to me!

PR #238, opened 2023-08-25
Simplify rpm advisory shipped event

FernandesMF:
Simplify handling of 'errata RPM advisory shipped' events

Freshmaker handled 'errata RPM advisory shipped' events with a workflow that is overly complicated. This commit simplifies that workflow.

Originally, the steps for handling 'errata RPM advisory shipped' events were:

1. when an rpm was signed, `generate_advisory_signed_on_rpm_sig.py` would generate an `ErrataAdvisoryRPMSignedEvent` event
2. when an advisory state was changed, `update_db_on_advisory_change.py` would generate an `ErrataAdvisoryRPMSignedEvent`
3. the `ErrataAdvisoryRPMSignedEvent` events would then be processed in `rebuild_imaged_on_rpm_advisory_change.py` and be turned into other events.

However, steps 1 and 2 are not needed at all. These steps were originally designed because Freshmaker was originally designed to handle both module and image rebuilds.
However, it does not have to handle module rebuilds anymore, and hence we can simplify the workflow.

The changes in this commit are the following:
- removal of `freshmaker/handlers/internal/generate_advisory_signed_event_on_rpm_sign.py` and `freshmaker/handlers/internal/update_db_on_advisory_change.py` modules, moer specifically of `GenerateAdvisorySignedEventOnRPMSign` and `UpdateDBOnAdvisoryChange` classes
- removal of `freshmaker/handlers/brew/sign_rpm.py` and `BrewSignRpmParser`
- removal of related configurations and tests
- renaming of `ErrataAdvisoryRPMsSignedEvent` to `ErrataRPMAdvisoryShippedEvent`
- adjustment of `ErrataAdvisoryStateChangedParser` to return `ErrataRPMAdvisoryShippedEvent` events directly

Now the parser `ErrataAdvisoryStateChangedParser` will generate `ErrataRPMAdvisoryShippedEvent` events directly, and we can shortcircuit what the handlers in steps 1 and 2 did previously.

This will make for a simpler workflow, which is easier to understand and to maintain.

JIRA: CWFHEALTH-2259

### Testing remarks

I made a test deployment to the dev environment and used the init_messages mount. Freshmaker processed the test messages without errors.


qixiang:
Good job, :+1:

qixiang (approved):

PR #239, opened 2023-09-04
Fix async build

qixiang:
1. Include `brew.package` in image metadata returned by Pyxis, it's used by async build feature.

2. In async build handler, PyxisAPI instance is used with its APIs, but while resolving an image, the resolve methods accepts a PyxisGQL instance, so async build fails with `find_images_by_nvr` method not found, just wrap the API in PyxisAPI as a workaround.


FernandesMF (approved):
Looks good to me! 👍🏼

PR #240, opened 2023-09-13
Fix pyxis timeout retry

qixiang:
1. Fix retry for Pyxis request timeout
    
In 31cc4ca, we added retry mechanism when Pyxis request timeout, however
we detect the Pyxis request timeout by:

    if "Pyxis API was unable to fetch data from MongoDB in time" in error

`error` is actually a dict, not a string, so this doesn't work.

2. Disable flake8 W503

W503 warning in Flake8 goes against PEP 8 recommendation, disable it to
make it compatible with black.

    W503 line break before binary operator

https://peps.python.org/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

### DIFF
--- a/freshmaker/pyxis_gql.py
+++ b/freshmaker/pyxis_gql.py
@@ -98,7 +98,10 @@ class PyxisGQL:
         if error is not None:
             trace_id = self._client.transport.response_headers.get("trace_id", False)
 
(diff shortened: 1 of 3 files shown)


FernandesMF (approved):
Looks great to me!

PR #241, opened 2023-09-13
Remove caching for some find images APIs

qixiang:
Caching should not be enabled for some find images APIs, because the result can change at any time, if we have caching for such APIs, images just be released may not be returned by these APIs which can cause wrong result of images to be rebuilt.

These APIs are from `pyxis_gql` module:

    1. find_images_by_installed_rpms
    2. find_images_by_names
    3. find_images_by_name_version
    4. find_images_by_repository

Caching for other find images APIs is fine:

    1. find_images_by_nvr
    2. find_images_by_nvrs

### DIFF
--- a/freshmaker/pyxis_gql.py
+++ b/freshmaker/pyxis_gql.py
@@ -438,7 +438,6 @@ class PyxisGQL:
 
         return images
 
-    @region.cache_on_arguments()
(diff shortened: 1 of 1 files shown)


FernandesMF (approved):
Makes all the sense, and looks good to me!

PR #243, opened 2023-09-18
Exclude `total` field from Pyxis request's projection

qixiang:
The requests in find_images_by_installed_rpms API are very inefficient, including the `total` field can make the query performance much worse.


FernandesMF (approved):
Looks good to me! I hope this solves the timeout problem!

PR #245, opened 2023-09-20
Append advisory name to message id

qixiang:
Before b76bd0a, freshmaker appends the advisory name to message id when event is triggered by rpm advisory shipped message, it's still useful for us to know whether an event was triggered by E2E testing advisories (which type is RHBA) without opening Errata tool, so adding it back.


FernandesMF (approved):
Looks good to me!

PR #247, opened 2023-09-27
Refactor the Container class

qixiang:
Changes to the Container class:

1. Change Container class to normal class

The class originally implemented using dataclass didn't have many of the advantages of dataclass, but rather made creating instances more difficult to understand. The reason is a container can have multi-arch images, most attributions doesn't have one-to-one mapping with a single image data.

2. Use class method `create_from_images` to create Container instances

It was implemented with instantiating it first then load per-arch data with `load` which is confusing.

3. Remove `resolve_published`

Just check if a container is published by checking that in all deliver repositories.

JIRA: CWFHEALTH-2313

### DIFF
--- a/conf/config.py
+++ b/conf/config.py
@@ -2,7 +2,7 @@
 
 import os
 import tempfile
-from typing import Optional, Tuple, Union  # noqa
+from typing import Any, Optional, Tuple, Union  # noqa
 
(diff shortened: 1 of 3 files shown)


FernandesMF (approved):
Looks good to me, I don't have anything to add!

PR #249, opened 2023-09-27
only generate ErrataRPMAdvisoryShippedEvent for 'shipped live' events

FernandesMF:
Check event state before issuing ErrataRPMAdvisoryShippedEvent

In a previous refactoring, we simplified how Freshmaker generates ErrataRPMAdvisoryShippedEvent's.
However, we forgot to check the state of the incoming errata events, and FM is currently issuing
ErrataRPMAdvisoryShippedEvent for all Errata events, even those which are not shipping events.

This commits places a check to ensure that ErrataRPMAdvisoryShippedEvent will only be issued
for the correct kind of Errata events.

### Testing remarks

I made a deployment to the dev environment, using the initial messages. I didn't see any errors for the initial messages nor the other messages after two days.

JIRA: CWFHEALTH-2351

### DIFF
--- a/freshmaker/parsers/errata/state_change.py
+++ b/freshmaker/parsers/errata/state_change.py
@@ -63,12 +63,14 @@ class ErrataAdvisoryStateChangedParser(BaseParser):
         # type of advisory triggered the event without opening Errata tool.
         msg_id = f"{msg_id}.{str(advisory.name)}"
 
-        # If advisory created by BOTAS and it's shipped,
-        # then return BotasErrataShippedEvent event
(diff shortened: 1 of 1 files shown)


freshmaker/parsers/errata/state_change.py
```diff
@@ -65,10 +65,10 @@ def parse(self, topic, msg):
 
         # If advisory created by BOTAS and it's shipped,
         # then return BotasErrataShippedEvent event
```
qixiang, line 67: Looks good to me, but please move this comment accordingly.

qixiang (approved):
Minor comment.

PR #250, opened 2023-10-10
Black-reformat step in tox

FernandesMF:
Freshmaker is currently using a black-check step in tox. That means that black will check if it would reformat some file, and exits an error in case it would.

However, it would also be nice to have black actually reformat files during development, using the same parameters as will the check step.

This commit introduces the `black-reformat` in tox, which can be called manually during development to make the actual reformatting. It also introduces quick instructions on how to call it the readme file.

JIRA: CWFHEALTH-2357

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,23 @@ python -m pip install pip-tools
 more info available at: https://github.com/jazzband/pip-tools/
 
 
+# Formatting code with black
+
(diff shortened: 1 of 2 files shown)


tonisbones:
LGTM. Thanks Mario!

tonisbones (approved):
LGTM

ElenaKarolinaSemanova (approved):
LGTM :+1:

PR #252, opened 2023-10-25
Replace the parent image only with the latest one

gnaponie:
Due to regressing content, it might happen, in rare conditions, that the parent image of the image we are rebuilding after a CVE fix doesn't include all the latest fixes.

Example: we have 3 images:
	NVR | CVE fixes
	foo-1-125: fix-a, fix-b
	foo-1-124: fix-a, fix-c
	bar-2-220: fix-a, fix-b
foo is parent of bar. Before this change, Freshmaker would ask Pyxis if there's already a parent image fixing fix-c, and it would return foo-1-124, so that Freshmaker can use that as parent of bar. But that image is missing fix-b!
To prevent this issue, from now on, Freshmaker ask to Pyxis only the latest image.

Tests are not updated, since the current tests should cover the case.

### DIFF
--- a/freshmaker/pyxis_gql.py
+++ b/freshmaker/pyxis_gql.py
@@ -24,6 +24,7 @@ from typing import Optional
 
 import backoff
 import dogpile.cache
+import kobo.rpmlib
 from gql import Client, gql
 from gql.dsl import DSLQuery, DSLSchema, dsl_gql
 from gql.transport.requests import RequestsHTTPTransport
@@ -624,22 +625,17 @@ class PyxisGQL:
 
         return images
 
-    def find_images_by_name_version(
-        self, name, version, published=None, content_sets=None, limit=20
-    ):
+    def find_latest_images_by_name_version(self, name, version, published=None, content_sets=None):
         """
-        Find all the images that match the specified name, version, and are filtered by the given content sets.
+        Find the latest images that match the specified name, version, and are filtered by the given content sets.
 
         :param name str: image name
         :param version str: image version
         :param published bool: image is published or not
         :param content_sets list: list of content set names
-        :param limit int: the maximum number of images to be returned
         :return: list of container images matching the name and version
         :rtype: list of image data
         """
-        images = []
-
         query_filter = {"and": []}
         query_filter["and"].append({"brew": {"package": {"eq": name}}})
         query_filter["and"].append({"brew": {"build": {"regex": f"{name}-{version}-.*"}}})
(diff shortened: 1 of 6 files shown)


gnaponie:
@qixiang @FernandesMF PTAL
Please notice that I didn't add any new tests, since the existing one is testing the function already.
I thought about one that would test this corner case, but since we are mocking the Pyxis data nothing came to my mind.
In case you have any suggestions I'll be happy to add some more tests. Thank you.

FernandesMF (approved):
Looks good to me, great job! 😁 

I agree that the unit tests are fine as they are.

About the black error, I think the 'black-format' step that was recently merged can take care of it (https://github.com/redhat-exd-rebuilds/freshmaker/commit/91156689a7d595b43a6bd1dd5ee9001b06225de8).

Or, if you want to be extra careful (and if tox is running fine on your local machine), you can temporarily add a `--diff` flag on the black step, just to check what it would reformat.

freshmaker/pyxis_gql.py
```diff
@@ -652,46 +649,26 @@ def find_images_by_name_version(
...
+            ds.ContainerImagePaginatedResponse.total,
+            ds.ContainerImagePaginatedResponse.data.select(*self._get_image_projection()),
+        )
 
-        # Note: this can truncate the list with keeping some arches of an image build but
-        # removing other arches, but we don't care about it in our use cases, we just want
-        # the most higher NVR builds, if you're calling this with a much smaller limit
-        # number (e.g.: 1, 2, 3) or for different purpose, you should pay attention to this.
-        return images[:limit]
+        return self.query(query_dsl)["find_images"]["data"]
```
qixiang, line 674: This works when image supports 4 arches, but when it supports just 1, 2 or 3 arches, this will return images for more than one NVR. For example, when it supports 3 arches, this will return 2 NVRs like:
```
foo-125
foo-125
foo-125
foo-125
```

And I checked with Pyxis, seems sorting by `brew.build` doesn't work for us, because Pyxis considers `foo-99` is higher than `foo-125`. 

I also tried sorting by `repositories.published_date`, in some cases it doesn't work either, for example, the query for `s2i-base-container-1-*` images returns this:
```
s2i-base-container-1-465.1698060557.amd64
s2i-base-container-1-465.1697652944.amd64
s2i-base-container-1-465.1698060557.s390x
s2i-base-container-1-465.1698060557.ppc64le
s2i-base-container-1-465.1697652944.ppc64le
s2i-base-container-1-465.1698060557.arm64
s2i-base-container-1-465.1697652944.arm64
s2i-base-container-1-465.1697652944.s390x
...
```
so probably we should still query with more page_size (e.g. 20), but sort by `repositories.published_date`, then compare the NVRs in result in freshmaker and only return the images for the highest NVR.
```
latest_nvr = images[0]["brew"]["build"]

for img in images[1:]:
    parsed_latest_nvr = kobo.rpmlib.parse_nvr(latest_nvr)

    current_nvr = img["brew"]["build"]
    parsed_current_nvr = kobo.rpmlib.parse_nvr(current_nvr)
    if kobo.rpmlib.compare_nvr(parsed_current_nvr, parsed_latest_nvr) == 1:
        latest_nvr = current_nvr

return [img for img in images if img["brew"]["build"] == latest_nvr]
```

gnaponie: Gotcha. Thanks for the catch. I'll make the change and I will try query pyxis too, perhaps adding some test with a real query result data.

gnaponie:
@FernandesMF ok, thanks for the heads-up. I think I'll rebase on main to include your changes, and if it keeps failing I'll add a commit to fix the black check.

gnaponie:
@qixiang @FernandesMF could you please check again? I've fixed black, but I'm not really sure why it still fails. I've applied all the suggestions it prompted.

FernandesMF:
Hi @gnaponie! I pulled the code from your branch and took a look at black. It turns out black is acting weird indeed.

When I ran black, it complained about line 61 in `test_pyxis_gql_async.py`,
```
        fake_query = ds.Query.find_repositories(page=0, page_size=2, filter={},).select(
```
and wanted to rewrite it the way you had it before with respect to the arguments of `find_repositories`.

Seems like we found a bug, and black is not idempotent in this case.

I found a quick solution of disabling it for this line. You can put a `  # fmt: skip` by the end of the `select` block:
```
        fake_query = ds.Query.find_repositories(page=0, page_size=2, filter={},).select(
           ...
        )  # fmt: skip
```
at least, that did the trick in my test. Notice the two spaces before the `#`!

FernandesMF:
Oh, and I think we need to manually squash all our commits together in Freshmaker. So don't forget to squash all your commits when you are ready to merge!

qixiang:
> Hi @gnaponie! I pulled the code from your branch and took a look at black. It turns out black is acting weird indeed.

If you're running the same version black (as it is in the github tests) locally, it should show you consistent result. This is caused by a change in black version 23 (https://github.com/psf/black/blob/ddfecf06c13dd86205c851e340124e325ed82c5c/CHANGES.md?plain=1#L323).

It's unnecessary to have the magic trailing comma in:
```
fake_query = ds.Query.find_repositories(page=0, page_size=2, filter={},).select(
```
so we can just remove it:
```
fake_query = ds.Query.find_repositories(page=0, page_size=2, filter={}).select(
```

The rest of the code looks good to me.

freshmaker/pyxis_gql.py
```diff
@@ -652,46 +648,38 @@ def find_images_by_name_version(
...
-                ),
-                ds.ContainerImagePaginatedResponse.page,
-                ds.ContainerImagePaginatedResponse.page_size,
-                ds.ContainerImagePaginatedResponse.total,
-                ds.ContainerImagePaginatedResponse.data.select(*self._get_image_projection()),
-            )
+        query_dsl = ds.Query.find_images(
+            page=0,
+            # "page_size" is equal to 4, because the maximum arches supported
+            # by RH containers is 4 (amd64, ppc64le, aarch64 and s390x)
```
qixiang, line 658: This comment is not valid anymore, we can remove it to avoid confusions.

qixiang (approved):

PR #255, opened 2023-11-23
Enable use of external repos in lower-severity rebuilds

FernandesMF:
When rebuilding for lower-severity advisories, we do not want to check the images in  all repositories in Pyxis. Instead, we want to restrain our search to a (dynamic) short list of predefined repos, stored in a remote yaml file, as this is more efficient.

This commit implements this functionality through the following changes:
- new logic in `RebuildImagesOnRPMAdvisoryChange._find_images_to_rebuild` to fetch external repos from the remote yaml file, in case advisory priority is `moderate`
- new functions `RebuildImagesOnRPMAdvisoryChange._lookup_external_repos` and `utils.py/load_remote_yaml` to fetch the content of the remote yaml file
- new parameters `repositories` and `repositories` in the methods `PyxisAPI.find_images_with_packages_from_content_set`, `PyxisAPI.find_images_to_rebuild` and `PyxisAPI.find_repositories`, but mostly as a means of forwardng to `PyxisGQL/PyxisGQLAsync.find_repositories`
- new parameter `names` in  `PyxisGQL/PyxisGQLAsync.find_repositories` to limit the search to the received repo names
- (chore) removal of unnecessary `is not None` clauses in `PyxisGQL/PyxisGQLAsync`
- (chore) reorganization of imports in `utils.py` to conform with PEP8
- unit tests for the new functions and cases

JIRA: CWFHEALTH-2493

### DIFF
--- a/freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
+++ b/freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
@@ -34,6 +34,7 @@ from freshmaker.pulp import Pulp
 from freshmaker.errata import Errata
 from freshmaker.types import ArtifactType, ArtifactBuildState, EventState, RebuildReason
 from freshmaker.models import Event, Compose, ArtifactBuild
+from freshmaker.utils import load_remote_yaml
 
 
 class RebuildImagesOnRPMAdvisoryChange(ContainerBuildHandler):
@@ -101,7 +102,7 @@ class RebuildImagesOnRPMAdvisoryChange(ContainerBuildHandler):
         batches = self._find_images_to_rebuild(db_event.search_key, skip_nvrs=skip_nvrs)
         builds = self._record_batches(batches, event)
 
-        if not builds:
+        if not builds and db_event.state != EventState.FAILED:
             msg = "No container images to rebuild for advisory %r" % event.advisory.name
             self.log_info(msg)
             db_event.transition(EventState.SKIPPED, msg)
@@ -451,6 +452,31 @@ class RebuildImagesOnRPMAdvisoryChange(ContainerBuildHandler):
             "Going to find all the container images to rebuild as " "result of %r update.",
             affected_nvrs,
         )
+
+        adv_is_compliance_priority = getattr(self.event.advisory, "is_compliance_priority", False)
+        # We should only use a restriction for external repos for advisories that are not critical
+        # or important; we should not apply the restriction for critical/important advisories when
+        # they have the 'compliance-priority' label
+        compliance_priority_repos = None
+        if adv_is_compliance_priority and self.event.advisory.security_impact not in [
+            "critical",
+            "important",
+        ]:
+            try:
+                compliance_priority_repos = self._lookup_external_repos()
+            except Exception as e:
+                msg = f"Unable to fetch external repos: {str(e)}"
+                self.log_error(msg)
+                db_event = Event.get_or_create_from_event(db.session, self.event)
+                db_event.transition(EventState.FAILED, msg)
+                db.session.commit()
+                return []
+
+            if not compliance_priority_repos:
+                msg = "No external repositories are specified in the remote yaml, skipping this event."
+                self.log_info(msg)
+                return []
+
         batches = pyxis.find_images_to_rebuild(
             affected_nvrs,
             content_sets,
@@ -459,5 +485,16 @@ class RebuildImagesOnRPMAdvisoryChange(ContainerBuildHandler):
             release_categories=release_categories,
             leaf_container_images=leaf_container_images,
             skip_nvrs=skip_nvrs,
+            repositories=compliance_priority_repos,
         )
         return batches
+
+    def _lookup_external_repos(self) -> list[str]:
+        """Fetches the external repositories to be used for lower criticality CVEs
+
+        :return: Names of the external repositories
+        :rtype: list of str
+        """
+
+        url = conf.compliance_priority_repositories_remote_file
+        return load_remote_yaml(url).get("repositories", [])
--- a/freshmaker/image.py
+++ b/freshmaker/image.py
@@ -513,6 +513,7 @@ class PyxisAPI(object):
         published=True,
         release_categories=conf.container_release_categories,
         auto_rebuild=True,
+        names: list[str] | None = None,
     ):
         """
         Returns dict with repository name as key and ContainerRepository as
@@ -524,12 +525,13 @@ class PyxisAPI(object):
             release categories (options: Deprecated, Generally Available, Beta,
             Tech Preview)
         :type release_categories: tuple[str] or list[str]
+        :param names: names of the repos to limit the search to (optional)
         :rtype: dict
         :return: Dict with repository name as key and ContainerRepository as
             value.
         """
         repositories = self.pyxis.find_repositories(
-            published=published, release_categories=release_categories
+            published=published, release_categories=release_categories, names=names
         )
 
         # If the query is for published images, add configurable repos for
@@ -1019,6 +1021,7 @@ class PyxisAPI(object):
         published=True,
         release_categories=conf.container_release_categories,
         leaf_container_images=None,
+        repositories: list[str] | None = None,
     ):
         """Query Pyxis and find containers which contain given
         package from one of content sets
@@ -1040,12 +1043,15 @@ class PyxisAPI(object):
         :param list leaf_container_images: List of NVRs of leaf images to
             consider for the rebuild. If not set, all images found in
             Pyxis will be considered for rebuild.
+        :param repositories: Repos to restrict the search to (Optional)
+        :type repositories: list of str
 
         :return: a list of dictionaries which represents container images
         :rtype: list
         """
 
-        repos = self.find_repositories(published, release_categories)
+        repos = self.find_repositories(published, release_categories, names=repositories)
+
         if not repos:
             return []
         if not leaf_container_images:
@@ -1335,6 +1341,7 @@ class PyxisAPI(object):
         filter_fnc=None,
         leaf_container_images=None,
         skip_nvrs=None,
+        repositories: list[str] | None = None,
     ):
         """
         Find images to rebuild through image build layers
@@ -1363,6 +1370,8 @@ class PyxisAPI(object):
             Pyxis will be considered for rebuild. Note that `published`
             is not respected when `leaf_container_images` are used.
         :param list skip_nvrs: List of NVRs of images to be skipped.
+        :param repositories: Repos to restrict the search to (Optional)
+        :type repositories: list of str
         """
         images = self.find_images_with_packages_from_content_set(
             rpm_nvrs,
@@ -1371,6 +1380,7 @@ class PyxisAPI(object):
             published,
             release_categories,
             leaf_container_images=leaf_container_images,
+            repositories=repositories,
         )
 
         # Remove any hotfix images from list of images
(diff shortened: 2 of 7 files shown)


gnaponie:
Hey Mario, could you check the failures in the CI? It looks like there is some failure with black and flake8.

freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
```diff
@@ -451,6 +452,29 @@ def _find_images_to_rebuild(self, errata_id, skip_nvrs=None):
             "Going to find all the container images to rebuild as " "result of %r update.",
             affected_nvrs,
         )
+
+        adv_is_compliance_priority = getattr(self.event.advisory, "is_compliance_priority", False)  
+        # We should only use a restriction for external repos for advisories that are not critical
+        # or important, even if they have the 'compliance-priority' label
+        restrict_to_ext_repos = None
+        if adv_is_compliance_priority and self.event.advisory.security_impact not in ["critical", "important"]:
```
qixiang, line 460: The current requirement is only rebuild for "moderate" advisories, "low" security impact advisories are not in current scope.

gnaponie:
> * I included the transitioning of events to 'skipped' in 'RebuildImagesOnRPMAdvisoryChange._find_images_to_rebuild' [(link)](https://github.com/redhat-exd-rebuilds/freshmaker/commit/dbbc11a0088a60f7829b13b750252bae8edfe384#diff-cf23b67460e3e2483625c426152414515c3375daf541ae537277fc1126b2bc93R467-R475) in case there is an HTTP error or there are no external repositories in the remote yaml file. However, I thought that `RebuildImagesOnRPMAdvisoryChange.handle` would already transition the events to 'skipped' if we just returned an empty list from `_find_images_to_rebuild` [(link)](https://github.com/redhat-exd-rebuilds/freshmaker/blob/dbbc11a0088a60f7829b13b750252bae8edfe384/freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py#L102-L110) , is that not so?


Looks like you're right! Sorry I've forgot about it. I believe you can remove the SKIPPED transition since it will be handled by the `handle` method when the builds are empty.

freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
```diff
@@ -451,6 +452,29 @@ def _find_images_to_rebuild(self, errata_id, skip_nvrs=None):
             "Going to find all the container images to rebuild as " "result of %r update.",
             affected_nvrs,
         )
+
+        adv_is_compliance_priority = getattr(self.event.advisory, "is_compliance_priority", False)  
+        # We should only use a restriction for external repos for advisories that are not critical
+        # or important, even if they have the 'compliance-priority' label
+        restrict_to_ext_repos = None
```
qixiang, line 459: I feel "external repos" is a little confusing, I'd suggest just call it "compliance priority repositories".

freshmaker/image.py
```diff
@@ -1371,6 +1380,7 @@ def find_images_to_rebuild(
             published,
             release_categories,
             leaf_container_images=leaf_container_images,
+            restrict_to_ext_repos=restrict_to_ext_repos,
```
qixiang, line 1383: As above, "external repos" is misleading, I'd suggest just use `repositories` here in parameters, it should not be specific to the "compliance priority" case, it can be used in other cases as well, just name `repositories` makes more sense.

freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
```diff
@@ -451,6 +452,29 @@ def _find_images_to_rebuild(self, errata_id, skip_nvrs=None):
...
+        if adv_is_compliance_priority and self.event.advisory.security_impact not in ["critical", "important"]:
+            db_event = Event.get_or_create_from_event(db.session, self.event)
+            try:
+                restrict_to_ext_repos = self._lookup_external_repos()
+            except Exception as e:
+                msg = f"Unable to fetch external repos: {str(e)}"
+                self.log_error(msg)
+                db_event.transition(EventState.SKIPPED, msg)
```
qixiang, line 467: When error happens, we should update the event to `FAILED` state with the error message in `state_reason`.

gnaponie:
> * In `test_moderate_cve_external_repo` [(link)](https://github.com/redhat-exd-rebuilds/freshmaker/commit/dbbc11a0088a60f7829b13b750252bae8edfe384#diff-c52676f2db73e5e8e2ac8b161dfc3fd237e1d643973d7e41295948a2963a8ac1R692) , is it okay to use `release_categories = None`? I'm not sure about the logic behind `release_categories`, so I just to double check if that is okay 😅

`release_categories` can be equal to Deprecated, Generally Available, Beta, Tech Preview. I don't think it's really important for the sake of the test.

freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
```diff
@@ -459,5 +483,19 @@ def _find_images_to_rebuild(self, errata_id, skip_nvrs=None):
...
+    def _lookup_external_repos(self) -> list[str]:
+        """Fetches the external repositories to be used for lower criticality CVEs
+
+        :return: Names of the external repositories
+        :rtype: list of str
+        """
+        
+        url = conf.compliance_priority_repositories_remote_file
+        parsed_content = load_remote_yaml(url)
+        ext_repos = parsed_content["repositories"]
```
gnaponie, line 499: `ext_repos` is redundant here. You can just return
`return parsed_content["repositories"]`

Even shorter:
`return load_remote_yaml(url).get("repositories")`
and you can delete `parsed_content` too.

freshmaker/utils.py
```diff
@@ -30,6 +30,9 @@
 import koji
 import kobo.rpmlib
 
+import backoff
```
gnaponie, line 33: Optional: being very annoying that's the best practices about order when importing:
``Imports should be grouped in the following order:
Standard library imports.
Related third party imports.
Local application/library specific imports.``

So perhaps I'd put these first.

freshmaker/utils.py
```diff
@@ -275,3 +278,25 @@ def is_valid_semver(version_string):
...
+    :type url: str or bytes
+    :return: The parsed yaml content
+    :rtype: Any
+    """
+
+    response = requests.get(url=url, timeout=conf.requests_timeout)
+    response.raise_for_status()
+    parsed_content = yaml.safe_load(response.content)
+
+    return parsed_content
```
gnaponie, line 302: also here you could just:
`return yaml.safe_load(response.content)`

tests/handlers/koji/test_rebuild_images_on_rpm_advisory_change.py
```diff
@@ -664,8 +672,57 @@ def test_affected_packages_with_modules(self, exists, affected_rpm_nvrs):
...
+
+        db_event = Event.get_or_create_from_event(db.session, self.event)
+        self.assertEqual(db_event.state, 4)
+        self.assertEqual(result, [])
+
+    @patch(
+        "freshmaker.handlers.koji.RebuildImagesOnRPMAdvisoryChange._lookup_external_repos",
+        return_value = []
+    )
+    def test_moderate_cve_external_repo_empty(self, mock_lookup):
```
gnaponie, line 716: This and the previous test are pretty similar. You could try to use `parametrize`. It allows you to change the variable values, but keep the code as it is. You can see an example in `tests/test_utils.py`. It's not always possible to use it (depends on how the test is written), but I think it should work in this case. At least, give it a try.

freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
FernandesMF: Right! I've changed that if clause to allow only for 'moderate' advisories. Thanks for spotting that! 😁

freshmaker/image.py
FernandesMF: I liked your suggestion a lot! These parameter names are indeed much more clear.

freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
FernandesMF: I see! I changed it to transition the event to 'failed' in case we get an HTTP error.

I had a look in `handle`, and I saw that there was a block that transitioned the state of the event to 'skipped' in case we returned an empty list from `_find_images_to_rebuild` [(link)](https://github.com/FernandesMF/freshmaker/blob/rebuild-from-predefined-repos/freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py#L102-L110).

Thus I added a new condition to that 'if' clause to not do that, in case the event state is 'failed'. That way, `handle` will not transition the event from 'failed' to skipped after an error happens. Is that okay?

freshmaker/handlers/koji/rebuild_images_on_rpm_advisory_change.py
FernandesMF: That's much better! Thanks a lot 😁

freshmaker/utils.py
FernandesMF: That's not being annoying, don't worry! 😂 
I like to follow good practices 😁 

I ran isort in `utils` and reorganized all the imports. I saw that there were a few others in the wrong place. 
I also ran `isort --check` for the whole project, and seems like pretty much every file could use an import reorg 😂 
We could do that later in a separate commit and put it in the 'git-blame-ignore-revs' file.

freshmaker/utils.py
FernandesMF: Much better as well! Thanks a lot 😁

tests/handlers/koji/test_rebuild_images_on_rpm_advisory_change.py
FernandesMF: I didn't know about this trick in pytest, that is so cool!

I gave it a try, and I kept getting a "missing argument" error. It seems that parametrize is not compatible with `unittest.TestCase` subclasses according to [this](https://docs.pytest.org/en/stable/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses).

However, I saw that there is a llib called [parameterized](https://github.com/wolever/parameterized) that solves just that! I would be happy to try it out, if you think it is a nice addition to the project. What do you think, @gnaponie and @qixiang?

freshmaker/utils.py
gnaponie: Oh no :'( great job for checking that! Let's totally fix that. I'll file a card.

tests/handlers/koji/test_rebuild_images_on_rpm_advisory_change.py
gnaponie: I don't have a strong preference about `parametrize` or `parametrized`. I've also incurred some issues with `parametrize` from pytest in the past (I think it might be incompatible with some other decorator or something). So you can try parametrized if it is easier.

qixiang (approved):
LGTM

tests/handlers/koji/test_rebuild_images_on_rpm_advisory_change.py
gnaponie: Alright. Let's skip it then. I believe the change looks good at this point :+1: 
good job

gnaponie (approved):

PR #258, opened 2024-01-18
Fix unexpected argument 'flags' to _fake_odcs_new_compose

qixiang:
In `0a95329`, `flags` was also added to _fake_odcs_new_compose call, this function doesn't accept `flags` argument and we don't need to support it in the fake function.


gnaponie (approved):

gnaponie:
+1

PR #259, opened 2024-01-18
Fix find_latest_images_by_name_version

qixiang:
It doesn't handle the case when Pyxis returns empty images.


gnaponie (approved):

gnaponie:
looks good

PR #260, opened 2024-02-14
Fix for ET's builds-by-cve API new result format

qixiang:
Errata has changed its builds-by-cve API to return rpms in string list to list of dicts, old format example:

    "x86_64": [
        "krb5-libs-1.8.2-3.el6_0.7.i686.rpm",
        "krb5-libs-1.8.2-3.el6_0.7.x86_64.rpm"
    ]

new format:

    "x86_64": [
        {
            "filename": "krb5-libs-1.8.2-3.el6_0.7.i686.rpm",
            "is_signed": true
        },
        {
            "filename": "krb5-libs-1.8.2-3.el6_0.7.x86_64.rpm",
            "is_signed": true
        }
    ]

This commit updates Freshmaker to accommodate the changes in the result format introduced by the latest ET API.

JIRA: CLOUDWF-10046

### DIFF
--- a/freshmaker/errata.py
+++ b/freshmaker/errata.py
@@ -472,7 +472,7 @@ class Errata(object):
                                 # Remove '.arch.....' part from rpm's name
                                 # and make a list from them
                                 if arch != "SRPMS":
-                                    just_nvrs = [rpm.rsplit(".", 2)[0] for rpm in rpms]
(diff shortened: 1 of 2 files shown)


qixiang:
The unit test failures are caused by the updated version of `black` and other test libraries, which will be fixed in a separated PR.

cit1zen:
+1 to the hotfix.

PR #269, opened 2024-03-25
Fix Github action and test failures

qixiang:
1. Fix doc build error in Github action
2. Fix unit test failures


FernandesMF (approved):
Looks good to me, I can't think of any suggestion!

PR #280, opened 2024-07-29
Replace outdated base images with the latest ones

qixiang:
When a base image included in the rebuild list as a dependency is not the latest published release, replace it with the latest one. This ensures that outdated base images, which might reference non-existent resources, do not cause build failures. By using the latest base images, we can ensure successful image builds.

JIRA: CWFHEALTH-3258

### DIFF
--- a/freshmaker/image.py
+++ b/freshmaker/image.py
@@ -111,6 +111,10 @@ class ContainerImage(dict):
     def nvr(self):
         return self["brew"]["build"]
 
+    @property
+    def is_base_image(self):
+        return self["filesystem_koji_task_id"] is not None
+
     def log_error(self, err):
         """
         Logs the error associated with this image and sets self["error"].
@@ -153,6 +157,7 @@ class ContainerImage(dict):
             "arches": None,
             "odcs_compose_ids": None,
             "parent_image_builds": None,
+            "filesystem_koji_task_id": None,
         }
 
     @classmethod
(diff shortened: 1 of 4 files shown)


tests/test_image.py
```diff
@@ -3335,3 +3335,57 @@ def test_get_fixed_published_image_not_found_by_nvr(
...
     assert image is None
+
+
+@patch("freshmaker.pyxis_gql.PyxisGQL.find_latest_images_by_name_version")
+@patch("freshmaker.pyxis_gql.PyxisGQL.find_images_by_nvr")
+@patch("freshmaker.image.ContainerImage.resolve")
+def test_replace_base_images(
+    mocked_resolve, mocked_find_images_by_nvr, mocked_find_images_by_name_version
+):
+    vulerable_bash_rpm_manifest = {
```
gnaponie, line 3346: typo: s/vulerable/vulnerable/g

freshmaker/image.py
```diff
@@ -1554,6 +1564,66 @@ def _filter_out_already_fixed_published_images(
...
+            if image.nvr in replacements:
+                new_image = replacements[image.nvr]
+                if not new_image:
+                    continue
+                image_group[-1] = new_image
+                image_group[-2]["parent"] = new_image
+                continue
+
+            # Skip non-base images
+            if not image.is_base_image:
```
gnaponie, line 1592: Aren't you sure already that this is a base image? Because it is in `image_group[-1]`.

If not, wouldn't it be better to move this if in the beginning of the for loop?

gnaponie:
Looks good +1
only minor comments, not really important.

gnaponie (approved):

freshmaker/image.py
qixiang: Base image is always the last image in `image_group`, but it doesn't mean the last image is always base image, when the vulnerable rpms are not from base image, the last image is a "layered" image.

tests/test_image.py
qixiang: Fixed

freshmaker/image.py
qixiang: Moved to the beginning of loop.

gnaponie:
Looks good

PR #284, opened 2024-10-25
Ignore images with missing brew nvr data

FernandesMF:
Images built and released by Konflux lack Brew metadata. Freshmaker will encounter problems, because it currently expects this data.

This commit makes Freshmaker ignore those images. This is done in `images.PyxisAPI.find_images_with_included_rpms`, right after querying pyxis, by checking the brew nvr field.

### DIFF
--- a/freshmaker/image.py
+++ b/freshmaker/image.py
@@ -726,6 +726,10 @@ class PyxisAPI(object):
         if not images:
             return []
 
+        # Skip images without Brew metadata. Images built and released by Konflux
+        # lack Brew metadata. We don't support rebuilding such images.
+        images = [x for x in images if x["brew"]]
+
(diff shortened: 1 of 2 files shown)


freshmaker/image.py
```diff
@@ -726,6 +726,10 @@ def find_images_with_included_rpms(
         if not images:
             return []
 
+        # Images whose parents are built by Konflux will not have NVR
+        # information in the 'brew' field, and should be skipped
```
qixiang, line 730:
```suggestion
        # Skip images without Brew metadata. Images built and released by Konflux
        # are lack Brew metadata. We don't support rebuilding such images.
```

gnaponie (approved):

PR #290, opened 2025-01-20
Skip retries for Jira errors

qixiang:
Due to the implementation of the allow list and block list, freshmaker is actually checking all advisories with all state changes. When freshmaker lacks permissions to access certain Jira issues, it continuously retries these checks, causing significant performance degradation.

This change removes retry attempts for Jira permission errors, as these retries would never succeed. This workaround prevents unnecessary delays in processing advisory state change events.


FernandesMF (approved):
Good catch! I have no suggestions to the code.

PR #291, opened 2025-02-07
Update product page query params

qixiang:
Freshmaker queries product page to determine whether an openshift version is released or not, however the old query params doesn't work now because the task name of "GA" is not always used consistently, using the "ga" or "ga-main" flag works for us.


FernandesMF (approved):
Looks good to me!

PR #293, opened 2025-02-12
Update GitHub runner image

hluk:
The Ubuntu 20.04 runner image will be unsupported by April 1.

https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


qixiang:
There is another one in `.github/workflows/docs.yml`, I opened #292 to update these images.

hluk:
> There is another one in `.github/workflows/docs.yml`, I opened #292 to update these images.

Oh, cool. I have not seen that. :)

PR #306, opened 2026-06-29
CWFHEALTH-4892: Upgrade to fedora 43

lslebodn:
* Migrate from pkg_resources to mportlib.metadata
* Allow to sue different python interpreter for the manager script

# Upgrade to fedora43
    
We still use python3.12 due to pinned dependencies in requirements.txt and smallest chance of break.
That is reason why python-rhmsg is installed from tarball(extracted from src.rpm) for python3.12
Because  rpm from copr on fedora 43 is built for python 3.14

 * Remove unneded dependenciesfor systemd
    
    Logging to journald was removed in 740f13377b96be7ec32c8d4a6a36ac0aa2b2b279
   
*     Install missign service-identity
```   
    There are warnigns:
    bash-5.1(freshmaker)$ oc logs pod/freshmaker-prod-consumer-177-bns57 -c consumers | grep WARN
    {"asctime": "2026-07-23 07:14:37", "name": "moksha.hub", "levelname": "WARNING", "message": "Cannot find qpid python module. Make sure you have python-qpid installed.", "taskName": null}
    {"asctime": "2026-07-23 07:14:37", "name": "moksha.hub", "levelname": "WARNING", "message": "Connecting without SNI support due to absence of service_identity module.", "taskName": null}
    
    sh-5.2$ rpm -q python3-service-identity
    python3-service-identity-23.1.0-3.fc39.noarch
    sh-5.2$ rpm -e python3-service-identity
    error: Failed dependencies:
            python3.12dist(service-identity) >= 18.1 is needed by (installed) python3-twisted+tls-23.10.0-1.fc39.noarch
    
    pip-compile -P 'twisted==23.10.0' -P 'service-identity==23.1.0' requirements.in --output-file requirements.txt
```
* Install mod_wsgi for python3.12
```    
    [root@c9160fed813b src]# rpm -qa | grep wsgi
    python3-mod_wsgi-5.0.2-6.fc43.x86_64
    
    Using the same version as in rpm
    
     pip-compile -P 'mod_wsgi==5.0.2' requirements.in --output-file requirements.txt
```
*     Install missing stomper
```    
      File "/usr/local/lib/python3.12/site-packages/moksha/hub/stomp/stomp.py", line 108, in buildProtocol
        self.proto = StompProtocol(self, self.username, self.password)
      File "/usr/local/lib/python3.12/site-packages/moksha/hub/stomp/protocol.py", line 52, in __init__
        stomper.Engine.__init__(self)
    builtins.NameError: name 'stomper' is not defined
    
    pip-compile -P 'stomper==0.4.3' requirements.in --output-file requirements.txt
```
* requirements: Add missing psycopg2
```    
    pip-compile -P 'psycopg2==2.9.6' requirements.in --output-file requirements.txt
```
*  tox: Fix execution on fedora43
````    
    Some pinned dependencies in requirements.txt cannot be installed with python3.14.
    Therefore some linting tests failed with default python3 on fedora43.
    
    Most of them needn't install dependencies but some of them need's them.
    So we use default python3.12 for them.
```
* actions: Upgrade fedora-python/tox-github-action 

Most of version were pinned based on previous image based on f39
@FernandesMF PTAL


manager
```diff
@@ -33,7 +33,7 @@ case $1 in
         ;;
 
     runfreshmaker)
-        "${project_root}/start_backend_from_here" &
+        ${PYTHON:-python} "${project_root}/start_backend_from_here" &
```
FernandesMF, line 36: `start_backend_from_here` is still using `pkg_resources`, I guess you will need to update it there too:
https://github.com/lslebodn/freshmaker/blob/fedora_43/start_backend_from_here#L4

Dockerfile
```diff
@@ -1,28 +1,36 @@
...
 WORKDIR /src
 
 COPY . .
 
 RUN \
-    pip3 install -r requirements.txt && \
-    pip3 install .
-
-ENV REQUESTS_CA_BUNDLE='/etc/pki/tls/certs/ca-bundle.crt'
+    pip3 --python /usr/bin/python3.12 install "setuptools<82" && \
```
FernandesMF, line 23: I think it would be nice to have a comment explaining why setuptools>=82 would cause problems, what do you think?

Dockerfile
```diff
@@ -1,28 +1,36 @@
...
+    pip3 --python /usr/bin/python3.12 install -r requirements.txt && \
+    mkdir /tmp/python-rhmsg && \
+    pushd /tmp/python-rhmsg && \
+    dnf download --srpm python3-rhmsg && \
+    rpm2archive python-rhmsg-0.12*.*.src.rpm | tar -xvzf - && \
+    pip --python /usr/bin/python3.12 install ./rhmsg-0.12.0.tar.gz && \
```
FernandesMF: Since the previous line uses the glob expansion `rpm2archive python-rhmsg-0.12*.*.src.rpm`, shouldn't this line use a similar approach for the tarball?

This might be nitpicking though, since we are using qwan's copr for this package, so there shouldn't be an update unless we make it ourselves, right?

FernandesMF (changes requested):

manager
lslebodn: initially, I was too ambitious and wanted get rid of pkg_resources which were removed in setuptools 82.0
* https://setuptools.pypa.io/en/stable/history.html#v82-0-0

I kept just few simple commits and decided to pin down setuptools 82.0 for this PR

Dockerfile
lslebodn: One never know; which reminds me to fork the repo in copr :-)
just to have a backup

Dockerfile
lslebodn: Comment added

FernandesMF (approved):
Looks good and makes sense to me!

PR #307, opened 2026-06-29
CWFHEALTH-5089: Fix IndexError in _get_impacted_bundles when bundle has no brew image

lslebodn:
get_images_by_digest can return an empty list when a bundle path
digest has no brew-built data in pyxis (e.g. Konflux-built bundles).
    
We need to skip these instead of crashing with an IndexError on bundle_images[0].
We cannot rebuilt operator bundles which were not built in brew.
```    
    Traceback:
      File "freshmaker/handlers/botas/botas_shipped_advisory.py", line 372,
        in _get_impacted_bundles
        bundle_nvr = bundle_images[0]["brew"]["build"]
    IndexError: list index out of range
```    


FernandesMF (dismissed):
Makes sense and looks good to me!

FernandesMF (approved):

PR #308, opened 2026-07-24
CWFHEALTH-5089: Bump graphql-core and gql

lslebodn:
It is needed for new version of pyxis which is already deployed in production
```
(venv) sh$ python -V
Python 3.12.13

(venv) sh$ pip-compile -P 'graphql-core>=3.2.11' -P 'gql>=4.0' requirements.in --output-file requirements.txt
```
Fixes: CWFHEALTH-5089

@FernandesMF @ElenaKarolinaSemanova PTAL

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -8,7 +8,7 @@ Flask
 Flask-Login
 Flask-Migrate
 Flask-SQLAlchemy
-gql[requests, aiohttp]
+gql[requests]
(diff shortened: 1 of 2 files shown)


FernandesMF (dismissed):
Looks good and makes sense to me

FernandesMF (approved):

PR #310, opened 2026-08-12
CWFHEALTH-5125: Use GSSAPI authenticated bind for LDAP queries

lslebodn:
Switch from anonymous LDAP binding to Kerberos (GSSAPI) authenticated
binding, and add the cyrus-sasl-gssapi system dependency required for
the SASL mechanism.

And couple of more tiny changes found by AI agents 

*  Sanitize uid in LDAP filter to prevent LDAP injection
    
    Escape special LDAP filter characters in the uid parameter using
    ldap.filter.escape_filter_chars() before interpolating into the
    search filter string.
    
* Ensure LDAP connection cleanup with explicit unbind_s
    
    Relying on __del__ for LDAP cleanup is not best practice — if an
    exception occurs mid-function, or on non-CPython runtimes (PyPy,
    etc.), cleanup timing isn't guaranteed. Wrap the LDAP search in
    try/finally with client.unbind_s() to ensure the SASL/GSSAPI context
    and connection are properly released.
    
### DIFF
--- a/freshmaker/auth.py
+++ b/freshmaker/auth.py
@@ -133,6 +133,7 @@ def query_ldap_groups(uid):
     :rtype: set
     """
     client = ldap.initialize(conf.auth_ldap_server)
+    client.sasl_gssapi_bind_s()
     users = client.search_s(
         conf.auth_ldap_user_base,
         ldap.SCOPE_ONELEVEL,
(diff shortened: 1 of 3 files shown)


FernandesMF (approved):
Looks good to me

PR #311, opened 2026-08-12
Apply blocklist filter to parent images in rebuild chain

lslebodn:
Previously, HANDLER_BUILD_BLOCKLIST only filtered directly-affected images. Parent images discovered via find_parent_images_with_package() were never checked against the blocklist. This meant that if a parent image couldn't be rebuilt (e.g. using obsolete base images), the failure cascaded to all children with "Cannot build artifact, because its dependency cannot be built."

Now filter_fnc is also applied to parent images in _get_images_to_rebuild. When a parent is blocked, the chain is truncated at that point and the child builds independently against its existing (non-rebuilt) parent.

This allows configuring e.g.:

    {"image_name": ["...-container"], "image_version": "7\\.4\\.16"}

to block specific parent versions while still rebuilding child images that pick up CVE fixes via yum update in their Dockerfiles.

### DIFF
--- a/freshmaker/image.py
+++ b/freshmaker/image.py
@@ -1430,6 +1430,22 @@ class PyxisAPI(object):
                 rebuild_list[rpm_name] = self.find_parent_images_with_package(
                     image, rpm_name, images=[], pyxis_api_instance=pyxis_api_instance
                 )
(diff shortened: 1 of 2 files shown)


FernandesMF (approved):
Makes sense and looks good to me

